Issue #6, opened 2017-02-02
Automate the Kubernetes Cluster Setup using Vagrant


yudaykiran:
Currently the Kubernetes Cluster Setup is a manual process. This process has to optimized by automating this. using vagrant for the creation of the nodes and forming a cluster out of the created nodes.

kmova:
Thanks Uday for the pull request.

PR #9, opened 2017-02-09
Issue #6   k8s cluster setup automation

yudaykiran:
Updated the logic to generate a token instead of hardcoding it in the Vagrantfile. Host nodes request Master Node for token and join the cluster on receiving the token.


prateekpandey14:
LGFM

Issue #70, opened 2017-03-15
Modify the plugin to interact with mayaserver to create vsm


akashshah26595:
#### Steps to be followed:
- create an openebs-pvc yaml using the parameters passed to the plugin
- issue the curl command with above openebs-pvc yaml
- wait (poll) using another curl for the volume to be created
- on success, login to the iscsi volume
- on successful login, format it with ext4
- mount the volume to the provided mount-path

PR #72, opened 2017-03-18
Modify script to interact with mayaserver to create vsm.

akashshah26595:
**Following methods introduced in the script to create/delete a volume, to create/delete a spec file and to also check whether a volume already exists.**
- createvolumespec
- clearvolumespec
- checkvolumeexists
- createvolume
- deletevolume

First step to fix issue #70.

**Pending**: 
-  To use iscsiadm.
-  Format the mounted disk.
-  Mount the volume.

### DIFF
--- a/k8s/demo/scripts/setup_omm.sh
+++ b/k8s/demo/scripts/setup_omm.sh
@@ -3,7 +3,7 @@
 #Variables:
 machineip="0.0.0.0"
 hostname=`hostname`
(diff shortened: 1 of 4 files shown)


akashshah26595:
Flex volume provides data only in flex volume block of yaml file along with filesystem type, and one another field( probably the type of volume)  , so volume-name won't be received by script.

PR #86, opened 2017-04-04
Change format to store curl responses.

akashshah26595:
Using **<command_name>-<volume_name>.json** format  to store curl response.

**Example:** 
curl response to create a volume *demo-vsm-1* is stored in *createvolume-demo-vsm-1.json* file.


kmova:
Please raise a new PR, that will only show the new commits.

Issue #104, opened 2017-04-30
Create an Ansible project for automating Continuous Integration.


yudaykiran:
This issue will track the feature implementation of automating CI using Ansible.

ksatchit:
The basic ansible based e2e framework for now available at :  https://github.com/openebs/openebs/tree/master/e2e

A brief description and contributing guidelines are here: https://github.com/openebs/openebs/blob/master/contribute/CONTRIBUTING-TO-CI-AND-E2E.md

PR #107, opened 2017-05-16
 PR-#104 Improvements to Ansible based CI framework

ksatchit:
- Created a new role for IOmeter. This role helps in creating & instantiating a container that
  consumes iSCSI storage and runs the dynamo.

- Placed the Dockerfile for building the iometer image & dynamo binary (which will be copied
  into the container) in ./files

- Created a new playbook, test-iometer.yml to run the iometer test on the lines of test-fio.yml

- Updated run-tests.yml to include test-iometer testcase. Also created a commenting convention
  to capture test details

- Updated fio role's defaults file (defaults/main.yml) to include fio_run_duration as an argument to
  the fio container command. Also, changed the variable name for the tests-fio container reflect
  "image" instead of "container"

- Updated the fio role's tasks (tasks/main.yml) to pass fio_run_duration while specifying the
  container command. Changed fio container variable name as mentioned above

- Updated the cleanup tasks (tasks/main.yml) to remove the stale node entry for the previous
  iSCSI session using iscsiadm-m node -T <iqn> -o delete


kmova:
Accepting this pull request. As a followup to this, some re-factoring is required w.r.t the contents of the ansible/files directory, which is becoming an umbrella folder for misc files.

PR #122, opened 2017-06-02
PR #104 Improvements to Ansible based CI framework

ksatchit:
Code Changes:
----------------

- Updated inventory generation file (generate-inventory.py) to include localhost details
  and added 'ansible_become_pass' arg for all hosts. This is needed for tasks that run
  with privilege escalation in case of non-root users

- Changed "is_vagrant_vm" variable to true in global vars file all.yml to facilitate updates
  to .profile of ansible_user

- Updated hosts role to check for status of maya-master before joining cluster

- Updated the vagrant role to use {{ansible_env.HOME}} ansible variable instead od constructing
  the ansible user's home directory path

- Updated the following roles to use the privilege escalation directive only where absolutely
  necessary, while removing them for other steps (if present). This was done to facilitate
  successful playbook execution on non-root users

   - master, hosts, common, prerequisites, cleanup

- Updated the test-k8s-mysql-pod testcase playbook to use $HOME of k8s-master machine instead of
  ansible user's home

- Added a new testcase playbook along with associated vars file to deploy percona-mysql pod on k8s
  cluster

- Updated the run_tests.yml playbook to include testcase block for test-k8s-percona-mysql-pod

- Updated the machines.in template with entry for localhost. Also updated the notes with regard
  to the localhost entry

Tested On:
-----------
Bare-metal/Non-vagrant VMs with
 - Non root users on all boxes - test harness/localhost, maya/k8s master, maya/k8s hosts

### DIFF
--- a/e2e/ansible/files/generate-inventory.py
+++ b/e2e/ansible/files/generate-inventory.py
@@ -71,6 +71,7 @@ def generateInventory(config, boxlist, path, codes):
             varsgroupname = '%s:vars' %(hostgroupname)
             config[varsgroupname] = {'ansible_ssh_user': user}
             config[varsgroupname]['ansible_ssh_pass'] = passwd
+            config[varsgroupname]['ansible_become_pass'] = passwd
             config[varsgroupname]['ansible_ssh_extra_args'] = "'-o StrictHostKeyChecking=no'"
             f = open(path, 'w')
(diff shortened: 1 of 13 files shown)


yudaykiran:
The changes look good to me. Can be merged.

Issue #198, opened 2017-08-23
oebs-cloud.sh script not allowiing me to create new bucket on AWS.


rafihnl:
I tried to create new bucket in AWS account using  **oebs-cloud.sh** script but it throughing an error. 

root@rafi-virtual-machine:~/openebs# ./oebs-cloud.sh --create-cluster-config
Creating Group on AWS for OpenEBS Users...
{
    "Group": {
        "Path": "/",
        "CreateDate": "2017-08-23T06:58:07.986Z",
        "GroupId": "AGPAJ4KQIS4QW6SRDJ4UW",
        "Arn": "arn:aws:iam::536403140708:group/openebsusers",
        "GroupName": "openebsusers"
    }
}
Creating S3 bucket to store cluster state...

**An error occurred (OperationAborted) when calling the CreateBucket operation: A conflicting conditional operation is currently in progress against this resource. Please try again.**

In AWS account BUCKET name should be unique please modify the script to create a unique name for the new buckets.

rafihnl:
After changing the script file  i'm able to create a new bucket with unique name.

But not able to create cluster configuration error as follow.

Creating Group on AWS for OpenEBS Users...
{
    "Group": {
        "Path": "/",
        "CreateDate": "2017-08-23T09:17:28.755Z",
        "GroupId": "AGPAIG5RPGSA27JZWF6H2",
        "Arn": "arn:aws:iam::536403140708:group/openebsusers",
        "GroupName": "openebsusers"
    }
}
Creating S3 bucket to store cluster state...
{
    "Location": "/openebs-k8s-hgtmrrk3dn-local-state-store"
}
Generating Public Key for Passwordless SSH
Generating public/private rsa key pair.
Your identification has been saved in /root/.ssh/id_rsa.
Your public key has been saved in /root/.ssh/id_rsa.pub.
The key fingerprint is:
SHA256:cszWEZ7TqmdEogKgvsLbhzygR40qxb06Wf66qumv9SI root@rafi-virtual-machine
The key's randomart image is:
+---[RSA 2048]----+
|.         .      |
|..       . +     |
|. .     . * .    |
|.  .   + + +     |
| o +. o S +      |
|..* +. + o       |
|o*o=..  . o      |
|=EB=+.   o       |
|=****=.          |
+----[SHA256]-----+
Generating terraform file(.tf)

[...]

PR #199, opened 2017-08-23
Fix for #198: Contains the code changes to generate a random unique name for AWS S3 bucket.

yudaykiran:
Generate unique S3 bucket names on AWS.

Code Changes:
-------------------

- Contains the fix to generate a random name for S3 bucket.
- Fix adheres to standard S3 bucket naming conventions.

Tested On:
-------------

ESX VM - (Ubuntu 16.04)

Details of the fix:
--------------------

The oebs-cloud.sh script used a standard name 'openebs-k8s-local-state-store' for creating a S3 bucket on AWS. But the S3 bucket names, seem to be universal and AWS does not allow the creation of the bucket with the same name for any other AWS user. So the name needs to be randomized so as  to avoid the duplicate name conflict.

### DIFF
--- a/e2e/terraform/oebs-cloud.sh
+++ b/e2e/terraform/oebs-cloud.sh
@@ -3,6 +3,7 @@
 set -e
 
 master_dns_name=
+s3_bucket_name=
 
 function show_help() {
     cat << EOF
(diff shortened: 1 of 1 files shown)


ksatchit:
Nice !! LGTM

Issue #645, opened 2017-10-17
Add "replica" property as configurable in openebs-storageclasses


yudaykiran:
## Is this a BUG REPORT or FEATURE REQUEST?
 BUG REPORT

**What happened**:
openebs-provisioner always creates two replicas, ignoring the value given for "replica" property inside the openebs-storageclasses yaml.

**What you expected to happen**:
openebs-provisioner must read the replica count assigned to the "replica" property in the openebs-storage classes and bring up the replicas as per the count mentioned in the storage classes.
 
**How to reproduce it (as minimally and precisely as possible)**:
Steps to reproduce the issue:
1. In the openebs-storageclasses.yaml change the "replica" value from "2" to "1".
2. Create a StatefulSet utilizing this storage class.
3. openebs-provisioner ends up creating "2" replicas rather than "1".

**Anything else we need to know?**:

kmova:
@yudaykiran  -- can you change the storage class properties to define parameters using the new policy format as explained in https://github.com/openebs/openebs/issues/863

For example, the replica policy should be:
```
opebebs.io/jiva-replica-count: "1"
```

Issue #638, opened 2017-10-16
Reduce the number of OpenEBS StorageClasses to Two. One for applications requiring HA and another for applications not requiring replication.


yudaykiran:
## Is this a BUG REPORT or FEATURE REQUEST?
 FEATURE REQUEST

**What happened**:
The openebs-storageclasses.yaml is getting cluttered as more and more applications are being added to the demo applications. It is better to categorize the storage classes based on what type of storage the application needs. 

**What you expected to happen**:
The openebs-storageclasses.yaml must contain only 2 storage classes, named:
- openebs-no-replication (Examples: redis, ETCD...)
- openebs-ha (Example: Percona, MongoDB, Cassandra, Postgres, Jupyter...)

Also:
- For the storage class of type openebs-no-replication the replica count must be 1. 
- All the existing demo applications also have to be updated to reflect these changes in storage classes.
 
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

ksatchit:
Merged with #652 

The new storage class ```openebs-standard``` has been tested by CI

PR #648, opened 2017-10-17
Fix for Issue #638

yudaykiran:
- Refactors the storage classes for OpenEBS and reduces the number of classes to 2.
- Updates the storage class changes to reflect in all the demo applications.

Issue #638

This PR is also dependent on the fix for Issue #645 , to pickup the properties from the storage classes. 

### DIFF
--- a/k8s/demo/cassandra/cassandra-statefulset.yaml
+++ b/k8s/demo/cassandra/cassandra-statefulset.yaml
@@ -80,7 +80,7 @@ spec:
   - metadata:
       name: cassandra-data
       annotations:
-        volume.beta.kubernetes.io/storage-class: openebs-cassandra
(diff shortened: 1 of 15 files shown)


kmova:
@yudaykiran 

Just looking at the changes, this is touching quite a lot of files. And can cause merge conflicts with some of the open PRs. Can we split the change as follows:

- Only replace the openebs-basic to openebs-standard (change only the demo yaml files that were using openebs-basic)
- Add a new class called openebs-standalone (for the no-replication case). Don't use this anywhere yet. 

I am going to close this PR, can you resend a new PR with the above changes.

yudaykiran:
Sure, will raise a new PR with the suggested changes.

Issue #877, opened 2017-11-08
terraform plan shows error in AWS


nsathyaseelan:
BUG REPORT 

**What happened**:
while trying to run ``` terraform plan``` in aws it will show the following error
``` Error asking for user input: 1 error(s) occurred:
* provider.aws: dial unix /tmp/plugin367202256|netrpc: connect: no such file or directory ```

yudaykiran:
`terraform init` downloads the latest version of the aws provider plugin. In this particular case the plugin that got downloaded was not backward compatible with the terraform version that was installed.

An issue for this is already logged: https://github.com/terraform-providers/terraform-provider-aws/issues/2121

There are a couple of workarounds:
1. Upgrade terraform to the latest version.
2. In the kubernetes.tf file search for `provider aws` and append `version = "1.1" ` to that section.

I will be raising a PR for the script where it will always download the latest version of terraform.

kmova:
Resolved by https://github.com/openebs/openebs/pull/882

PR #882, opened 2017-11-09
PR to fix Issue #877: terraform plan shows error in AWS

yudaykiran:
- Updated code to download the latest version of terraform.

Fix for Issue #877.

### DIFF
--- a/e2e/terraform/oebs-cloud.sh
+++ b/e2e/terraform/oebs-cloud.sh
@@ -100,7 +100,7 @@ function setup_local_env() {
         sleep 2
     else 
         echo "Missing terraform; Downloading and installing..."
-        terraform_url=$(curl https://releases.hashicorp.com/index.json | jq '{terraform}' | egrep "linux.*amd64" | sort --version-sort -r | egrep -v 'beta|rc' | head -1 | awk -F[\"] '{print $4}')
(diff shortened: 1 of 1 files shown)


e2e/terraform/oebs-cloud.sh
```diff
@@ -99,8 +100,9 @@ function setup_local_env() {
         sleep 2
     else 
         echo "Missing terraform; Downloading and installing..."
-        wget "https://releases.hashicorp.com/terraform/0.10.0/terraform_0.10.0_linux_amd64.zip"
-        unzip terraform_0.10.0_linux_amd64.zip
+        terraform_url=$(curl https://releases.hashicorp.com/index.json | jq '{terraform}' | egrep "linux.*amd64" | sort --version-sort -r | egrep -v 'beta|rc' | head -1 | awk -F[\"] '{print $4}')
```
ksatchit: Good one ! Any thoughts about egrep v/s grep -e , most linux flavours do support it though.

yudaykiran: egrep is equivalent to using "grep -E". Do you suggest I use "grep -E" instead?

ksatchit: Right. grep -e would be safer, I suppose.

yudaykiran: Done, changed "egrep" to "grep -E".

Issue #1035, opened 2017-12-19
OpenEBS Helm installation fails if namespace is used


muratkars:
While installing OpenEBS if you specify a namespace (other than default), openebs-provisioner fails to start and shows error message "Back-off restarting failed container, Error syncing pod"
In deployments, I noticed that maya-apiserver and openebs-provisioner are still in the default namespace and everything else are in the namespace I specified.
As I understand, we have "namespace: default" hardcoded in few files (deployment-maya-apiserver.yaml, deployment-maya-provisioner.yaml)
provisioner is deployed in default namespace and looking for maya-apiserver-service in the default namespace.

openebs-provisioner Pod status shows "Waiting: CrashLoopBackOff" and logs below:
```
$ kubectl logs openebs-provisioner-699b94bd7b-fhbnn
I1219 19:55:41.489689       1 maya_api.go:63] OpenEBS volume provisioner namespace default
E1219 19:55:41.491357       1 maya_api.go:67] Error getting maya-apiserver IP Address: services "maya-apiserver-service" not found
E1219 19:55:41.491386       1 openebs-provisioner.go:66] Error getting maya-apiserver IP Address: services "maya-apiserver-service" not found
```

List of available services:
```
$ kubectl get services --all-namespaces
NAMESPACE     NAME                         TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)             AGE
default       kubernetes                   ClusterIP   10.3.0.1     <none>        443/TCP             2h
kube-system   etcd-service                 ClusterIP   10.3.0.15    <none>        2379/TCP            2h
kube-system   heapster                     ClusterIP   10.3.0.226   <none>        80/TCP              2h
kube-system   kube-dns                     ClusterIP   10.3.0.10    <none>        53/UDP,53/TCP       2h
kube-system   kube-etcd                    ClusterIP   None         <none>        2379/TCP,2380/TCP   2h
kube-system   kube-etcd-client             ClusterIP   10.3.0.237   <none>        2379/TCP            2h
kube-system   kubernetes-dashboard         ClusterIP   10.3.0.44    <none>        80/TCP              2h
kube-system   tiller-deploy                ClusterIP   10.3.0.254   <none>        44134/TCP           2h
openebs       maya-apiserver-service       ClusterIP   10.3.0.227   <none>        5656/TCP            29m
openebs       openebs-grafana              NodePort    10.3.0.191   <none>        3000:32515/TCP      29m
[...]

PR #1037, opened 2017-12-20
Removing hardcoded namespace from Helm chart

muratkars:
Hardcoded namespace in OpenEBS Helm chart is causing deployment to fail if namespace is specified with Helm installation.
This issue is open under #1035 "OpenEBS Helm installation fails if namespace is used"

Solution is validated on StackPointCloud.


muratkars:
Previous Readme changes were meant for a partner's private repo, removed them and resent.

PR #1116, opened 2018-01-12
change in caption text colour to white in side navigation bar

ShovanMaity:
change the colour of heading inside the side navigation bar 

Previous screen

screenshot from 2018-01-12 18-38-23

New screen

new


inyee786:
@mahebbar  check out the colour of the text

PR #1119, opened 2018-01-15
Fix DEFAULT_REPLICA_COUNT environment variable issue

### LINKED ISSUE #1118: DEFAULT_REPLICA_COUNT environment variable doesn't work anymore
## Is this a BUG REPORT or FEATURE REQUEST?
 
 Choose one: BUG REPORT

**What happened**:
I see there are 2 replicas.

**What you expected to happen**:
I expected to see 1 replica when I set `DEFAULT_REPLICA_COUNT` to `1`.

**How to reproduce it (as minimally and precisely as possible)**:
Just update `openebs-operator.yaml` as docs says:
>Note By default, OpenEBS launches OpenEBS Volumes with two replicas. To set one replica, as is the case with single-node Kubernetes cluster, specify the environment variable DEFAULT_REPLICA_COUNT=1. This is supported in OpenEBS version 0.4 onwards.

**Anything else we need to know?**:
There is another environment variable called `OPENEBS_IO_JIVA_REPLICA_COUNT`. You should use it instead of `DEFAULT_REPLICA_COUNT`. I think we should update the docs.

mhemrg:
Updates related docs to deprecated `DEFAULT_REPLICA_COUNT` environment variable.
Fixes #1118

### DIFF
--- a/documentation/source/install/dev_solutions.rst
+++ b/documentation/source/install/dev_solutions.rst
@@ -153,9 +153,9 @@ Download the latest OpenEBS Operator files using the following commands.
    kubectl apply -f openebs-operator.yaml
 
 **Note** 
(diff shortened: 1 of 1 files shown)


mhemrg:
@AmitKumarDas I have updated it. Is it what you mean?

PR #1226, opened 2018-02-10
Update openebs-monitoring-pg.yaml to use Prometheus v2

metalmatze:

You're still running an outdated version of Prometheus. Let's bump it to v2.

I just saw that you also ship a Prometheus Operator folder. Should we update this one too?
It's fairly outdated as well. Instead of deploying Prometheus and Alertmanager as deployment the operator actually uses CRDs. Should we use these in the operator?
That would be another PR, I guess.

### DIFF
--- a/k8s/openebs-monitoring-pg.yaml
+++ b/k8s/openebs-monitoring-pg.yaml
@@ -75,15 +75,15 @@ spec:
       serviceAccountName: openebs-maya-operator
       containers:
         - name: prometheus
-          image: prom/prometheus:v1.7.2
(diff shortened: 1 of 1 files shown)


kmova:
Thank you @metalmatze. The changes look good. Yes, changing to prometheus operator can be in done in the next PR.  

Btw, could you please add the sign-off message to the PR commit message as mentioned here: https://github.com/probot/dco#how-it-works

metalmatze:
Done and force pushed the PR again.

kmova:
Related Issue: https://github.com/openebs/openebs/issues/1250

PR #1239, opened 2018-02-14
Update monitoring templates to support version > 2.0

utkarshmani1997:
* To update the the templates to support prometheus version > 2.0
* Latest version of grafana does not pick prometheus automatically due to different job name. After changing the job name from "openebs-prometheus-server" to "prometheus" it picks the prometheus default dashboard itself.
* When volume pod eviction takes place and same IP allocated to pod then multiple series error happens, so this commit fixes those errors.

None

Signed-off-by: utkarshmani1997 <utkarshmani1997@gmail.com>


kmova:
The prometheus and grafana have been updated. No changes were required to config. 

Refer: https://github.com/openebs/openebs/commit/5387cc52d6a904c6292179f9bfb8718e0cd70b53

PR #1259, opened 2018-02-21
Update operator to have permission to query namespaces

kmova:

Merge the changes to the cluster role as fixed in https://github.com/openebs/openebs/issues/1075 to master / helm charts.


prateekpandey14:
Can we also add CRD as well with `list , get, create` permissions ?

PR #1320, opened 2018-03-07
Openshift doc updates : More edits to restricted SCC

ksatchit:
As of today, the replica pod runs with a "root" user. By default the default (restricted) SCC disallows running pods as root. One way to get this running today is to use "RunAsAny" option in the restricted security context. Other approaches (creating a separate/dedicated security context for OpenEBS etc.,) needs to be thought about. 

Signed-off-by: ksatchit <karthik.s@cloudbyte.com>

### DIFF
--- a/k8s/openshift/byo/baremetal/README.md
+++ b/k8s/openshift/byo/baremetal/README.md
@@ -211,8 +211,9 @@ Alternatively, the following command may be used:
 oc adm policy add-scc-to-user anyuid -z default --as=system:admin 
 ```
 
-Note: While the above procedures may be sufficient to enable host access to the containers, it may also be needed to disable
-selinux (via ```setenforce 0```) to ensure the same. 
+Note: While the above procedures may be sufficient to enable host access to the containers, it may also be needed to :
+- Disable selinux (via ```setenforce 0```) to ensure the same. 
+- Edit the restricted scc to use ```runAsUser: runAsAny``` (the replica pod runs with root user) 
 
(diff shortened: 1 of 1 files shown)


k8s/openshift/byo/baremetal/README.md
```diff
@@ -211,8 +211,9 @@ Alternatively, the following command may be used:
 oc adm policy add-scc-to-user anyuid -z default --as=system:admin 
 ```
 
-Note: While the above procedures may be sufficient to enable host access to the containers, it may also be needed to disable
-selinux (via ```setenforce 0```) to ensure the same. 
+Note: While the above procedures may be sufficient to enable host access to the containers, it may also be needed to :
+- Disable selinux (via ```setenforce 0```) to ensure the same. 
+- Edit the restricted scc to use ```runAsUser: runAsAny``` (the replica pod runs with root user)
```
kmova: One of the users in slack mentioned that the above should be "RunAsAny"

kmova:
Let us put in a new issue for handling this via volume policy.

k8s/openshift/byo/baremetal/README.md
ksatchit: yes, will correct the typo !!

ksatchit: Done !

ksatchit:
Here is issue to track volume policy additions : https://github.com/openebs/openebs/issues/1325

Issue #1048, opened 2017-12-21
Explore available fault-injection tools for resiliency tests on kubernetes 


ksatchit:
## Is this a BUG REPORT or FEATURE REQUEST?
FEATURE REQUEST

**What happened**:

Here are some potential tools for fault-injection on kubernetes clusters. This issue has been raised to track the evaluation & comparison between said tools *for OpenEBS test purposes* 

- https://github.com/gaia-adm/pumba
- https://github.com/bloomberg/powerfulseal
- https://github.com/asobti/kube-monkey
- https://github.com/linki/chaoskube

**What you expected to happen**:

- Choose the appropriate tool for OpenEBS test requirements

ksatchit:
Here is the analysis after evaluation efforts of potential (opensource) fault-injection/resiliency tools for e2e.

## Powerfulseal

- Pros
  - Install via pip, runs as binary on master/node
  - Has both interactive/autonomous mode of operation (long weekend runs) 
  - Can kill pods by specifying matches for namespace, deployment, labels, with various filters (state, IP, regex, injection period etc., 
  - Can kill nodes as well

- Cons
   - Built for python 3, needed changes to run for python 2.7 (our env is based on this) 
   - Node kill (which is what differentiates it from other chaos tools for K8s) supported only for openstack (recent support for AWS). Needs separate inventory file/derives kubernetes inventory only for Openstack/AWS
   -  Demoed/Unveiled at KubeCon 2017, Austin, with not enough examples
   - Couldn't find any dedicated Slack channel/IRC
 
## Chaoskube

- Pros 
  - Can kill pods by specifying matches for namespace, deployment, labels with filters, regex
  - Simple to run - has a helm-based install & a deployment manifest
  - Fits into test approach where a "no-policy" deployment (w/ --dry-run) can be made upfront and test scripts can invoke chaoskube binary within pod with desired filters/regex.
  - Uses in-cluster config (kube-config) by default
  - Can setup autonomous mode with execution filters like day/time etc.., 
  - Written in GO - can extend it based on requirements. GoDocs reference available
  - Wider usage trends (~1.5 years old w/ active contributions) 
  
- Cons
   - Does "only" pod kills, no node kills OR network faults 
   - Pod kills are via kubectl (or mostly emulating kubectl delete pod) - So *relatively* graceful termination flow 
   - Uses RBAC as kill/delete is via kubernetes. Need the cluster setup to support/need right (admin) credentials to apply 
   - No slack channel/IRC

[...]

Issue #1383, opened 2018-03-27
Jiva volume is mounted as RO upon replica restart in a single replica configuration


ksatchit:
## Is this a BUG REPORT or FEATURE REQUEST?
BUG REPORT 

**What happened**:

A jiva volume replica restart caused by a chaos tool (chaoskube) resulted in the volume being remounted as RO at the client - on a single-replica setup, i.e., the PV consisted of a controller and one replica pod. 

The replica was seen to be restarted with 60s and was also scheduled back on the same node (the other node in the cluster was cordoned) 

- Application pod status: 

```
test@Master:~$ kubectl get pods
NAME                                                            READY     STATUS             RESTARTS   AGE
maya-apiserver-5d5944b47c-fbnjh                                 1/1       Running            0          5d
openebs-provisioner-6b8df9746c-gj4gk                            1/1       Running            1          5d
percona-85b8997987-r6bbz                                        0/1       CrashLoopBackOff   5          21m
pvc-b73b20fa-31ab-11e8-9309-000c298ff5fc-ctrl-f7d94646f-nz6r5   2/2       Running            0          21m
pvc-b73b20fa-31ab-11e8-9309-000c298ff5fc-rep-57d69d98b8-t5hgg   1/1       Running            0          7m
```

- Mount on the client: 

```
root@MayaHost02:/var/lib/kubelet/plugins/kubernetes.io/iscsi/iface-default/10.96.47.187:3260-iqn.2016-09.com.openebs.jiva:pvc-b73b20fa-31ab-11e8-9309-000c298ff5fc-lun-0# touch sample
touch: cannot touch 'sample': Read-only file system
```

- Controller logs: 

```
time="2018-03-27T10:43:51Z" level=warning msg="check condition"
time="2018-03-27T10:43:51Z" level=warning msg="check condition"
time="2018-03-27T10:57:16Z" level=error msg="Error reading from wire: EOF"
time="2018-03-27T10:57:17Z" level=error msg="Backend tcp://10.44.0.7:9502 monitoring failed, mark as ERR: EOF"
time="2018-03-27T10:57:17Z" level=info msg="Set replica tcp://10.44.0.7:9502 to mode ERR"
time="2018-03-27T10:57:17Z" level=info msg="Monitoring stopped tcp://10.44.0.7:9502"
[...]

PR #1385, opened 2018-03-27
Add the resiliency testcase for periodic single volume replica failures 

yudaykiran:
- Add the resiliency usecase for single replica.
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes # (Partial Fix for #1048)

### DIFF
--- a/e2e/ansible/playbooks/resiliency/test-single-replica-failure/chaoskube.yaml
+++ b/e2e/ansible/playbooks/resiliency/test-single-replica-failure/chaoskube.yaml
@@ -1,0 +1,55 @@
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: chaoskube
+  labels:
+    app: chaoskube
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: chaoskube
+    spec:
+      serviceAccountName: chaoskube
+      containers:
+      - name: chaoskube
+        image: quay.io/linki/chaoskube:v0.8.0
+        args:
+        # Provided below are possible flags to set in chaoskube.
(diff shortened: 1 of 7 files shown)


kmova:
@yudaykiran can you describe the approach followed here in the PR description?

ksatchit:
Note: Also refer #1383 for the bug seen while performing this test

yudaykiran:
When running OpenEBS with the number of volume replicas set to 1, the following behavior is expected of a pod when it is restarted:
- If the volume pod is restarted and is respawned on the same node it was running earlier, it should be able to start itself in RW mode rather than RO mode as quorum rules do not apply for this use case.

The ansible-playbook does the following:
1. Use Percona with load-gen to store data into the volume.
2. Identify the node in which the volume pod was created.
3. Cordon all other nodes, so that the volume pod gets scheduled on the same node if restarted.
4. Run chaos-kube for fault injection.(Pod Kills)
5. Check the status of the Percona Pod after the volume pod restart.
6. The Percona pod will be in running state if the volume pod started with the RW state and also the hostdir is in RW state, the test passes.
7. The Percona pod will be in CrashloopBackoff if the volume pod started with the RO or RW and the hostdir is in RO state, causing the volume to be inaccessible to write, the test failes.

kmova:
Shouldn't the replica always be restarted on the same node without cardoning?

kmova:
I got it. So this is to simulate the case where restart of the pod within the same node will not cause IO errors. The cases of ensuring that replica pods will not get rescheduled to a new node will be covered in a different test case.

Issue #1516, opened 2018-05-18
typing error in gokit.md file


sachinmukherjee:
BUG REPORT

**What happened**:
Some typing error in gokit.md file which resides under the directory `/documentation/techbits/gokit.md`. Current scenario `maya-apiserver` should be changed to `maya-api-server`.

**What you expected to happen**:
Add missing hyphens between `api` and `server`.

kmova:
The software component is called - maya-apiserver or m-apiserver. This is not a typo.

PR #1533, opened 2018-05-19
typing error in gokit.md file #1516

saumya1906:
Updated maya-api-server


saumya1906:
It was with reference to the following issue:
https://github.com/openebs/openebs/issues/1516

PR #1558, opened 2018-05-29
Use Bash's improved Conditional Expressions `[[` instead of `[`. Formatted some code.

shubhambhattar:

Changed code to use Bash's improved Conditional Expressions `[[` instead of `[`.

Since all the modified are supposed to run as Bash Scripts, its better to use Bash's new and improved conditional expressions. Makes the code simpler and more intuitive.

This is just to improve quality of code.


shubhambhattar:
@lxfontes Why is the DCO failing here? As per the discussion that happened [here](https://github.com/openebs/litmus/pull/55), I have put up the email as `shubhambhattar@users.noreply.github.com`.

I don't understand what exactly do I need to put as email in DCO. The above email was working the `litmus` repo but not working here.

shubhambhattar:
Check [#1564](https://github.com/openebs/openebs/pull/1564).

PR #1564, opened 2018-05-30
Use Bash's improved Conditional Expressions `[[` instead of `[`.

shubhambhattar:
Changed code to use Bash's improved Conditional Expressions `[[` instead of `[`.

Since all the modified are supposed to run as Bash Scripts, its better to use Bash's new and improved conditional expressions. Makes the code simpler and more intuitive.

 This is just to improve quality of code.


shubhambhattar:
@NisantaSahoo Can somebody review this PR? It's been open for some time.

PR #1658, opened 2018-07-12
[TA1787, TA1789] Rewrite PromQl to get the cumulative stats and recreate Dashboard

utkarshmani1997:
On branch US2355-dashboard
Changes to be committed:
	modified:   openebs-monitoring-pg.yaml
	modified:   openebs-pg-dashboard.json

1. Why is this change necessary?
-  PR openebs/maya#379 involves lot of change in logic w.r.t stats calculation
   so existing dashboard will not work.
-  To incorporate those changes in the dashboard.

2. How does it address the issue?
-  This commit make changes in the PromQL and dashboard JSON to get the
   cumulative stats.

3. What side effects does this change have?
-  Docker images with tags < 0.6.0 will not be compatible with this dashboard.

4. How to verify this change?
-  Make sure you are using the m-exporter's image tag > `0.6.0`
-  Deploy the openebs-monitoring-pg.yaml on kubernetes cluster.
-  Import the new dashboard and it should work.

5. Any specific message for reviewer ?
-  Added namespace field in the Prometheus's template.
-  Replaced the job name from `openebs-prometheus-server` to `prometheus` only.
   Prometheus's dashboard is breaking with the current specs, because grafana is autoconfigured with
   job name `prometheus`.

Signed-off-by: Utkarsh Mani Tripathi <utkarshmani1997@gmail.com>

### DIFF
--- a/k8s/openebs-monitoring-pg.yaml
+++ b/k8s/openebs-monitoring-pg.yaml
@@ -170,7 +170,7 @@ spec:
         app: openebs-grafana
     spec:
       containers:
-      - image: grafana/grafana:5.2.0
+      - image: grafana/grafana:5.1.3
         name: grafana
         ports:
         - containerPort: 3000
--- a/k8s/openebs-monitoring-pg.yaml
+++ b/k8s/openebs-monitoring-pg.yaml
@@ -170,7 +170,7 @@ spec:
         app: openebs-grafana
     spec:
       containers:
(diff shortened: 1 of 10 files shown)


utkarshmani1997:
Correct @avishnu but let's assume a case where a user using the both cstor and jiva then including the subsystem is making sense and making the dashboard does not require much effort.

PR #1719, opened 2018-08-02
Modified files: moved namespace deletion check to their respective main files. 

vibhor995:
Signed-off-by: vibhor995 <vvibhr995@gmail.com>

**Removed deletion check from cleanup files and added it to the main playbook in always block**:

1) It fixes the effort of manually deletion of namespaces if playbook failed when any of the task having 
     some errors.
2) Bring back the cluster to  its default state  even if any task is failing ,as namespaces is associated to the deployement gets deleted after complete run of playbook.

@ksatchit  @dargasudarshan @gprasath

### DIFF
--- a/e2e/ansible/playbooks/utils/delete_namespaces.yml
+++ b/e2e/ansible/playbooks/utils/delete_namespaces.yml
@@ -1,0 +1,17 @@
+---
+- name: Deletion of test specific  namespace.
+  shell: source ~/.profile; kubectl delete ns {{ namespace }}
+  args:
+    executable: /bin/bash
+  delegate_to: "{{groups['kubernetes-kubemasters'].0}}"
+
+- name: Check the namespace status.
+  shell: source ~/.profile; kubectl get ns 
+  args:
+    executable: /bin/bash
+  register: np
+  delegate_to: "{{groups['kubernetes-kubemasters'].0}}"
+  until: "'{{ namespace }}' not in np.stdout"
+  delay: 30
+  retries: 15
+  ignore_erros: true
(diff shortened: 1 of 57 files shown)


ksatchit:
@vibhor995 , how do we plan to handle namespaces stuck in "terminating" state post delete. I dont see confirmation checks for the namespace deletion.. this would be a good addition. And how about converting this into a namespace deletion util :)

vibhor995:
@ksatchit  yes we can do that ...i am working on it

vibhor995:
@ksatchit  review please

e2e/ansible/playbooks/utils/delete_namespaces.yml
```diff
@@ -0,0 +1,17 @@
...
+- name: Check the namespace status.
+  shell: source ~/.profile; kubectl get ns 
+  args:
+    executable: /bin/bash
+  register: np
+  delegate_to: "{{groups['kubernetes-kubemasters'].0}}"
+  until: "'{{ namespace }}' not in np.stdout"
+  delay: 30
+  retries: 15
+  ignore_erros: true
```
ksatchit: Typo at "ignore_erros" ! Also, why are we ignoring this ? The idea is to make an effort at all costs. 

Also, the namespace should ideally be passed as a param to util as a best practice (though it should work even w/o by virtue of adding vars file)

ksatchit:
Also, please resolve merge conflicts

gprasath:
@vibhor995 Can you execute all the tests with your changes and check if there is any impact.

PR #1734, opened 2018-08-07
Created utils file for namespace deletion/creation check and included it in  playbooks

vibhor995:
**Removed deletion check from cleanup files and added it to the main playbook in always block:**

1) It fixes the effort of manual deletion of namespaces if playbook failed when any of the tasks having
some errors.

2) Bring back the cluster to its default state even if any task is failing, as namespaces are associated with the deployment gets deleted after the complete run of a playbook.

@gprasath @ksatchit @dargasudarshan


e2e/ansible/playbooks/utils/namespace_task.yml
```diff
@@ -0,0 +1,44 @@
...
+       executable: /bin/bash
+     delegate_to: "{{groups['kubernetes-kubemasters'].0}}"
+
+   - name: Check the namespace status.
+     shell: source ~/.profile; kubectl get ns
+     args:
+      executable: /bin/bash
+     register: npdel
+     delegate_to: "{{groups['kubernetes-kubemasters'].0}}"
+     until: "'ns' not in npdel.stdout"
```
ksatchit: This will equate to a string called "ns" @vibhor995 !

vibhor995: @ksatchit  it was giving some warning message related to jinja templates so in order to keep the clean run of playbook I put it as ns, not {{ns}}. @dargasudarshan

ksatchit: @vibhor995 , there is no need to provide curly braces for the variables/jinja templates in conditional statements. Just *ns* should do. Adding single quotes will make it a strict string compare.

vibhor995:
**Status** : Complete run in proress with changes.

PR #1802, opened 2018-08-17
Decreased the length of the generated name for disk, due to gcloudsdk limit.

yudaykiran:
- Decreased the length of the generated name for disk, due to gcloudsdk limit.

Signed-off-by: Uday Kiran Y <uday.kiran@openebs.io>


ksatchit:
@yudaykiran , can you just add a comment linking to the relevant info on gcloudsdk disk name length limits (just to provide context for future code walkthroughs!)

yudaykiran:
@ksatchit , this issue is not there in recent versions of gcloudsdk, the ubuntu debian based package is lagging behind in versions. which is causing this problem, just to avoid the length issue, we reduced it by few characters.

PR #1837, opened 2018-08-23
Modified the cstor target label to get the target name in HA playbook

nsathyaseelan:
Signed-off-by: sathyaseelan <sathyaseelan.n@cloudbyte.com>

-This PR will change the label to get the target pod status.

### DIFF
--- a/e2e/ansible/playbooks/feature/ha/ha.yml
+++ b/e2e/ansible/playbooks/feature/ha/ha.yml
@@ -105,7 +105,7 @@
            - name: Get storage ctrl pod name
              shell: >
                source ~/.profile;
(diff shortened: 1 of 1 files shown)


e2e/ansible/playbooks/feature/ha/ha.yml
```diff
@@ -105,7 +105,7 @@
            - name: Get storage ctrl pod name
              shell: >
                source ~/.profile;
-               kubectl get pods -n {{ operator_ns }} -l openebs.io/controller=cstor-controller | grep {{ namespace }}
+               kubectl get pods -n openebs -l openebs.io/target=cstor-target | grep {{ namespace }}
```
kmova: We should continue to use operator_ns. It is possible that openebs components will be installed in a different namespace in future for cstor.

PR #1838, opened 2018-08-24
Refactor cStor Pool deployment template 

kmova:
Made the following changes:
- Provided a capability to set resource limits on cStor Pods. (Default is false). Can be overriden by specifying the limits in the SPC. 
- Made the Sparse Dir configurable
- Made the shared directory between containers in cstor pool pod, specific to that pool claim

Also, added an example percona deployment that uses sparse pool storage class.

### DIFF
--- a/k8s/openebs-pre-release-features.yaml
+++ b/k8s/openebs-pre-release-features.yaml
@@ -25,9 +25,15 @@ spec:
   # will be placed
   - name: RunNamespace
     value: openebs
-  # MemoryLimit specifies the max memory that can be used by the pool pods.
-  - name: MemoryLimit
-    value: 2Gi
+  # ResourceLimits allow you to set the limits on memory and cpu for pool pods
+  # The resource and limit value should be in the same format as expected by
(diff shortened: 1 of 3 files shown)


k8s/openebs-pre-release-features.yaml
```diff
@@ -18,10 +18,22 @@ spec:
   # nothing exists at the given path i.e. an empty directory will be created.
   - name: HostPathType
     value: DirectoryOrCreate
+  # SparseDir is a hostPath directory where to look for sparse files
+  - name: SparseDir
+    value: /var/openebs
```
avishnu: Should we make this default as /var/openebs/sparse ?
Reason is that in the default configuration, inside pool pod, /tmp is mounted from /var/openebs/<spc> and /var/openebs will be mounted from /var/openebs and potentially will contain the <spc> folder as well which is not needed.

kmova: good point! will make this change.

PR #1842, opened 2018-08-25
Fix naming conventions for run tasks

sonasingh46:
Put proper names for variables and task id in run tasks.
Signed-off-by: sonasingh46 <sonasingh46@gmail.com>

### DIFF
--- a/k8s/openebs-pre-release-features.yaml
+++ b/k8s/openebs-pre-release-features.yaml
@@ -39,7 +39,7 @@ spec:
     tasks:
     # Following are the list of run tasks executed in this order to
     # create a cstor storage pool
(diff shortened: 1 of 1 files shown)


k8s/openebs-pre-release-features.yaml
```diff
@@ -93,7 +93,7 @@ metadata:
   namespace: openebs
 spec:
   meta: |
-    id: cstorpoollistnode
+    id: listnode
```
kmova: what does this run task to? listnode or listdisksofnode?

sonasingh46: It will form a map of node to the disks.
The map will be formed only for the disk under investigation.

PR #1845, opened 2018-08-25
Update CAS templates to configure Pod Resource limits

kmova:
Updated the cStor Pool, cStor Target, jiva Target and Replica
    creation CAS templates with configuring CPU and Memory limits.
    The limits are disabled by default.
    
    For cStor Pool, the limits can be enabled via SPC
    For cStor Target, Jiva Target and Replica via the StorageClass
    
    SPC and StorageClass can also take AuxResourceLimits that will
    be set on the side-cars.
    
    The default jiva storage class, cstor sparse pool and storage class
    are updated with (commented) example on how to specify the CPU
    limits

### DIFF
--- a/k8s/openebs-pre-release-features.yaml
+++ b/k8s/openebs-pre-release-features.yaml
@@ -31,6 +31,8 @@ spec:
   #- name: PoolResourceLimits
   #  value: |-
   #      memory: 1Gi
+  - name: PoolResourceLimits
+    value: "false"
   # AuxResourceLimits allow you to set limits on side cars. Limits have to be specified
   # in the format expected by Kubernetes
   - name: AuxResourceLimits
--- a/k8s/openebs-pre-release-features.yaml
+++ b/k8s/openebs-pre-release-features.yaml
@@ -25,14 +25,19 @@ spec:
   # will be placed
   - name: RunNamespace
     value: openebs
-  # ResourceLimits allow you to set the limits on memory and cpu for pool pods
+  # PoolResourceLimits allow you to set the limits on memory and cpu for pool pods
   # The resource and limit value should be in the same format as expected by
   # Kubernetes. Example:
-  #- name: ResourceLimits
+  #- name: PoolResourceLimits
   #  value: |-
   #      memory: 1Gi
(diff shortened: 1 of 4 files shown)


k8s/openebs-pre-release-features.yaml
```diff
@@ -25,14 +25,15 @@ spec:
   # will be placed
   - name: RunNamespace
     value: openebs
-  # ResourceLimits allow you to set the limits on memory and cpu for pool pods
+  # PoolResourceLimits allow you to set the limits on memory and cpu for pool pods
```
sonasingh46: If `PoolResourceLimits` and `AuxResourceLimits` can be specified from SPC, what is intention to put it here?
Additionally if these are not found in SPC as well as CAS Template configs then it will be false by default.

k8s/openebs-pre-release-features.yaml
```diff
@@ -25,14 +25,15 @@ spec:
...
   # Kubernetes. Example:
-  #- name: ResourceLimits
+  #- name: PoolResourceLimits
   #  value: |-
   #      memory: 1Gi
-  # By default, the resource limits are disabled.
-  - name: ResourceLimits
+  # AuxResourceLimits allow you to set limits on side cars. Limits have to be specified
+  # in the format expected by Kubernetes
+  - name: AuxResourceLimits
```
sonasingh46: Similarly, this could be commented, if we do not want to use it.

k8s/openebs-pre-release-features.yaml
kmova: By default, there wont be any limits. Hence setting it to false inside the CAS templates. This feature is mostly for users to specify depending on their environment. 

There is no need to specify in SPC or StorageClasses and we can just document it. But I thought adding them there will serve as a reminder for users who might actually skip reading the documentation.

sonasingh46: Giving those in comments is fine,`PoolResourceLimits` is commented but not `AuxResourceLimits`

kmova: I see. That should be fixed!

PR #1848, opened 2018-08-26
Fix the cstor volume exporter mounts and add sample yamls

kmova:
- Updated the monitoring configuration to use the cstor / cas templates based labels. 
- Fixed the volume mount configuration for exporter side-car
- Added sample fio yamls with resource limit configuration settings.


k8s/openebs-pre-release-features.yaml
```diff
@@ -687,21 +687,21 @@ spec:
...
           labels:
             {{- if eq $isMonitor "true" }}
             monitoring: volume_exporter_prometheus
+            openebs_pv: {{ .Volume.owner }}
             {{- end}}
+            app: cstor-volume-manager
             openebs.io/target: cstor-target
             openebs.io/persistent-volume: {{ .Volume.owner }}
-            app: cstor-volume-manager
+            openebs.io/persistent-volume-claim: {{ .Volume.pvc }}
```
sonasingh46: Why same two labels with same value?

kmova: The below `pvc` is kept for tools that can't sustain a `.` in the label key - like the prometheus source label config. we might remove that `pvc` however in the future if no-one is reading it.

PR #1858, opened 2018-08-27
Env variable for default cstor sparsepool creation

sonasingh46:
Add a env variable for creation of cstor sparse pool when openebs is installed.
The env variable is : 
OPENEBS_IO_INSTALL_DEFAULT_CSTOR_SPARSE_POOL
If value of the env variable is true, installation of openebs will create a cstor sparse pool by default else it will not create.

Following PR should be merged to enable this feature.
https://github.com/openebs/maya/pull/483
Signed-off-by: sonasingh46 <sonasingh46@gmail.com>

### DIFF
--- a/k8s/openebs-operator.yaml
+++ b/k8s/openebs-operator.yaml
@@ -131,6 +131,11 @@ spec:
         # to maya install
         - name: OPENEBS_IO_INSTALL_CONFIG_NAME
           value: "maya-install-config-default-0.7.0"
+        # OPENEBS_IO_INSTALL_DEFAULT_CSTOR_POOL decides wheteher default cstor pool should be
+        # configured as a part op openebs installation.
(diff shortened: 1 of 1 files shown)


k8s/openebs-operator.yaml
```diff
@@ -131,6 +131,11 @@ spec:
         # to maya install
         - name: OPENEBS_IO_INSTALL_CONFIG_NAME
           value: "maya-install-config-default-0.7.0"
+        # OPENEBS_IO_INSTALL_DEFAULT_CSTOR_POOL decides wheteher default cstor pool should be
+        # configured as a part op openebs installation.
```
kmova: _n :  op -> of. 
-n : "cstor pool" -> cstor sparse pool"

sonasingh46: Yes

PR #1862, opened 2018-08-28
Add single sparse pool SPC and PVC yamls

prateekpandey14:
These yamls will be helpful to create single pool storagepoolclaim
 and cstor volumes in a single node kubernetes clusters like
 minikube etc.


k8s/demo/pvc-sparse-claim-cstor.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
kmova: See if these can be added into k8s/sample-yamls and also update the maya-apiserver ci to make use of the new location.

prateekpandey14: Done!

PR #1863, opened 2018-08-28
[WIP] Update mongo sts sample yaml with 0.7 annotations

kmova:


k8s/demo/mongodb/mongo-statefulset.yml
```diff
@@ -1,3 +1,24 @@
...
+# Since Mongo takes care of replication, one replica will suffice
+# Can be configured with Anti affinity topology key of hostname (default)
+#  or across zone. 
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: mongo-pv-az
+  annotations:
+    cas.openebs.io/config: |
```
yudaykiran: Are only the overrides required from now on for the CAS annotations? The other annotations that determine whether it is for jiva or cstor, Is it determined by the StoragePool and StoragePoolClaim?

kmova: Thats a good point @yudaykiran (cc: @princerachit ) -- However currently we default to jiva. In case cstor is required, we need to add an annotation saying "openebs.io/cas-type: cstor"

Issue #507, opened 2017-10-05
Enter symbol should be there after each command


NisantaSahoo:
**What happened**:
Enter symbol is missing after each command to execute it.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- kubectl get nodes
- kubectl get pods --all-namespaces
- kubectl get services
- kubectl get sc
- kubectl get pv
- kubectl get pvc
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

kmova:
@NisantaSahoo -- please update the issue with the links to the pages, where the modification is required.

rahulbajaj0509:
I would like to work on this bug, @NisantaSahoo can you please mention the files that need this update ?

kmova:
Please re-open this issue or raise a new issue with specific page that needs enhancements. The docs are now moved to https://github.com/openebs/openebs-docs

PR #1881, opened 2018-09-04
chore(monitoring): Removing unused relabel_config entries in openebs-monitoring-pg.yaml.

avishnu:
Removing the below unused relabel_config entry from openebs-monitoring-pg.yaml.
- Pod label: openebs_pv

Instead adding the following entry to make use of the latest label.
- Pod label: openebs.io/persistent-volume

Refer: [openebs/openebs#507](https://github.com/openebs/maya/pull/507)

### DIFF
--- a/k8s/openebs-monitoring-pg.yaml
+++ b/k8s/openebs-monitoring-pg.yaml
@@ -52,11 +52,9 @@ data:
       - source_labels: [__meta_kubernetes_pod_name]
         action: replace
         target_label: kubernetes_pod_name
-      - source_labels: [__meta_kubernetes_pod_label_vsm]
+      - source_labels: [__meta_kubernetes_pod_label_openebs_io_persistent_volume]
         action: replace
         target_label: openebs_pv
-      - source_labels: [__meta_kubernetes_pod_label_openebs_pv]
-        action: keep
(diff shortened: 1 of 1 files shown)


k8s/openebs-monitoring-pg.yaml
```diff
@@ -52,11 +52,9 @@ data:
       - source_labels: [__meta_kubernetes_pod_name]
         action: replace
         target_label: kubernetes_pod_name
-      - source_labels: [__meta_kubernetes_pod_label_vsm]
```
kmova: cool! didn't know about this special char replace with _. 

On the change itself:
- should we leave it as it is for backward compatibility, possibly for another release till all the 0.6 volumes get upgraded? 
- the below openebs_pv can be related with this openebs.io/persistent-volume?

avishnu: Yes @kmova , found that special char can be replaced with _ in the link:
https://github.com/prometheus/prometheus/blob/release-2.3/documentation/examples/prometheus-kubernetes.yml.

avishnu: We can leave the vsm tag for backward compatibility sake. I'll raise another commit for the same.

PR #1893, opened 2018-09-07
Create pvc-standard-cstor-disk.yaml

ranjithwingrider:
Signed-off-by: ranjith.raveendran@openebs.io

New sample yaml for consuming cstor pool created using external disk. We are proving sample SPC with storage class in the doc but we dont have pvc to consume this Pool.

So that all sample pvc yamls should be under same location


kmova:
@ranjithwingrider - Can you point me to the doc location where this file will be referred? wondering if the docs itself can contain this example.

ranjithwingrider:
@kmova It is the replacement of step 8 in https://docs.openebs.io/docs/next/deploycstor.html

ranjithwingrider:
I feel all samples will be under one location and  so that user can select the needed yaml as per the naming convention if he is not following doc also. Also existing one which is being used currently has been modified.

kmova:
cool. Can you put it with other sample yamls in `k8s/sample-pv-yamls`

ranjithwingrider:
It's Done.

PR #2058, opened 2018-10-07
Fix broken link for labels of issues in CONTRIBUTING.md

geekysrm:
Signed-off-by: Soumya <soumyacool2012@gmail.com>

**What this PR does / why we need it**:
This PR fixes the broken "labels of issues link" in CONTRIBUTING.md.

PR #2061, opened 2018-10-07
Added Github's recommended .gitignore template

### LINKED ISSUE #2042: Use Github's recommended .gitignore files
## Suggestion

Use [GitHub's `.gitignore` file templates](https://github.com/github/gitignore) instead of manually writing `.gitignore` files. They are more complete and serve as a good starting point for new projects.

### Golang's `.gitignore`
```
# Binaries for programs and plugins
*.exe
*.exe~
*.dll
*.so
*.dylib

# Test binary, build with `go test -c`
*.test

# Output of the go coverage tool, specifically when used with LiteIDE
*.out
```

geekysrm:
We should use GitHub's .gitignore file templates instead of manually writing .gitignore files. They are more complete and serve as a good starting point for new projects.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2042

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,16 @@ k8s/demo/*test*
 documentation/build/
 k8s/lib/vagrant/*.box
 k8s/lib/vagrant/*.log
+
+# Binaries for programs and plugins
+*.exe
+*.exe~
+*.dll
+*.so
+*.dylib
+
+# Test binary, build with `go test -c`
+*.test
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ There is always something more that is required, to make it easier to suit your 
 
 ## Contributing to Source Code and Bug Fixes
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -4,3 +4,16 @@ k8s/demo/*test*
 documentation/build/
```
kira0204: It would be great if we could club this under one of the headings (new heading preferred)

CONTRIBUTING.md
```diff
@@ -36,7 +36,7 @@ There is always something more that is required, to make it easier to suit your
 
 ## Contributing to Source Code and Bug Fixes
 
-Provide PRs with appropriate tags for bug fixes or enhancements to the source code. For a list of tags that could be used, see [this]((./contribute/labels-of-issues.md)).
+Provide PRs with appropriate tags for bug fixes or enhancements to the source code. For a list of tags that could be used, see [this](./contribute/labels-of-issues.md).
```
kira0204: Also, I think this change is already introduced in #2058. Let's solve the `.gitignore` issue only in this PR?

geekysrm: Yeah sure. It was added by mistake.
Closing this PR. Will open a PR with only the .gitignore part

PR #2076, opened 2018-10-12
chore(maya): Add snapshot cast as environment variables m-apiserver.

princerachit:
This is to support changes in maya as the snapshotting technique has changed and now uses cas templates.

Signed-off-by: princerachit <prince.rachit@mayadata.io>


princerachit:
closed in favour of https://github.com/openebs/maya/pull/646

PR #2252, opened 2018-11-02
refactor(operator-yaml): Set default value of sparse pool create to false

princerachit:


kmova:
Going ahead with default enabled for now. We will revisit this in the future.

PR #2259, opened 2018-11-08
Update es-data-svc.yaml

aslubsky:
Fix demo error `The Service "elasticsearch-data" is invalid: spec.clusterIP: Invalid value: "none": must be empty, 'None', or a valid IP address`

Signed-off-by: Alex Slubsky <aslubsky@gmail.com>


kmova:
Thank you @aslubsky Good Catch! could you update the PR or amend the commit with a email sign-off.

PR #2275, opened 2018-11-14
fix(sample): fix the sc beta annotation example yaml

prateekpandey14:


kmova:
Already pushed this change via this commit: https://github.com/openebs/openebs/commit/ccae860af9595b45be76e41740869e1c7f7059d2

PR #2291, opened 2018-11-22
Disk pattern /dev/md needs to be filtered out

umamukkara:
Following is the kubectl describe output for the unfiltered disk
kubectl describe disk disk-60f8ab1b1f6eedab231431c8de62183a
Name:         disk-60f8ab1b1f6eedab231431c8de62183a
Namespace:
Labels:       kubernetes.io/hostname=gke-gke-ssd-3zones-default-pool-da1e81a2-b0jn
              ndm.io/disk-type=disk
Annotations:  <none>
API Version:  openebs.io/v1alpha1
Kind:         Disk
Metadata:
  Creation Timestamp:  2018-11-22T14:27:36Z
  Generation:          1
  Resource Version:    26768
  Self Link:           /apis/openebs.io/v1alpha1/disks/disk-60f8ab1b1f6eedab231431c8de62183a
  UID:                 c25c4f7f-ee62-11e8-92ff-42010a80018c
Spec:
  Capacity:
    Logical Sector Size:  512
    Storage:              0
  Details:
    Compliance:
    Firmware Revision:
    Model:
    Serial:
    Vendor:
  Path:                 /dev/md0
Status:
  State:  Active
Events:   <none>

### DIFF
--- a/k8s/openebs-operator.yaml
+++ b/k8s/openebs-operator.yaml
@@ -332,7 +332,7 @@ data:
         name: path filter
         state: true
         include: ""
-        exclude: "loop,/dev/fd0,/dev/sr0,/dev/ram,/dev/dm-"
+        exclude: "loop,/dev/fd0,/dev/sr0,/dev/ram,/dev/dm-,/dev/md-"
 ---
 apiVersion: extensions/v1beta1
 kind: DaemonSet


k8s/openebs-operator.yaml
```diff
@@ -332,7 +332,7 @@ data:
         name: path filter
         state: true
         include: ""
-        exclude: "loop,/dev/fd0,/dev/sr0,/dev/ram,/dev/dm-"
+        exclude: "loop,/dev/fd0,/dev/sr0,/dev/ram,/dev/dm-,/dev/md-"
```
satbirchhikara: Do we need /dev/loop  instead of loop ?

kmova: it matches for substring. so loop should work. 

@umamukkara the suffix of "-" in the "/dev/md-" should be removed. 
(In case of device mapper, the paths will be /dev/dm-0 etc., But for md, looks like it is /dev/md0)

kmova:
The changes proposed in this PR have been pushed as part of https://github.com/openebs/openebs/pull/2314

PR #2383, opened 2019-02-11
fix (yaml) : changes for NDM configs in openebs operator

akhilerm:
`seachest-probe` is included in the list of probes.
Update strategy changed to `RollingUpdate` for NDM DaemonSet.
Entrypoint of ndm image is changed to script instead of directly
executing the binary.

### DIFF
--- a/k8s/charts/openebs/templates/cm-node-disk-manager.yaml
+++ b/k8s/charts/openebs/templates/cm-node-disk-manager.yaml
@@ -19,6 +19,9 @@ data:
       - key: udev-probe
         name: udev probe
         state: true
+      - key: seachest-probe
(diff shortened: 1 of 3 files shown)


kmova:
@akhilerm -- the changes look good. Just to keep the chart yamls in sync, can you make the same changes to the following files as well:
- https://github.com/openebs/openebs/blob/master/k8s/charts/openebs/templates/cm-node-disk-manager.yaml
- https://github.com/openebs/openebs/blob/master/k8s/charts/openebs/templates/daemonset-ndm.yaml

akhilerm:
@kmova The charts have been synced with latest changes.

PR #2402, opened 2019-02-18
(fix-bug) Killing the parent process when child process exit with status 1

mittachaitu:
Killing the parent process when the child process exited with status 1


mittachaitu:
Since it is executing as a child process it is exiting with exit code 0

PR #2526, opened 2019-04-16
fix(travis): Use cstorpool name instead of storagepool

prateekpandey14:


k8s/ci/test-script.sh
```diff
@@ -129,7 +129,7 @@ export MAPIPOD=$(kubectl get pods -o jsonpath='{.items[?(@.spec.containers[0].na
 export CSTORVOL=$(kubectl get pv -o jsonpath='{.items[?(@.spec.claimRef.name=="cstor-vol1-1r-claim")].metadata.name}')
 export CSTORVOLNS=$(kubectl get pv -o jsonpath='{.items[?(@.spec.claimRef.name=="openebs-pvc-in-custom-ns")].metadata.name}')
 export JIVAVOL=$(kubectl get pv -o jsonpath='{.items[?(@.metadata.annotations.openebs\.io/cas-type=="jiva")].metadata.name}')
-export POOLNAME=$(kubectl get csp -o jsonpath='{.items[?(@.metadata.labels.openebs\.io/cas-type=="cstor")].metadata.name}')
+export POOLNAME=$(kubectl get csp -o jsonpath='{.items[?(@.metadata.labels.openebs\.io/storage-pool-claim=="sparse-claim-auto")].metadata.name}')
```
mittachaitu: Good to get spc name via `kubectl get spc -o jsonpath='{.items[*].metadata.name}'` and use it to get csp

prateekpandey14: Its good idea, but above is only valid in case of singlepool , either way using direct labels will  be better in case of more then 1 pool. In coming future may be we can have different pools based on different test cases.

PR #2533, opened 2019-04-18
chore(upgrade): change upgrade version from 0.8.2-RC4 to 0.8.2

sonasingh46:
1. Change upgrade version from 0.8.0-RC4 to 0.8.2
2. Remove recreate strategy patch 

Signed-off-by: Ashutosh Kumar <ashutosh.kumar@openebs.io>


k8s/upgrades/0.8.1-0.8.2/jiva_volume_upgrade.sh
```diff
@@ -166,9 +151,6 @@ sed -u "s/@target_version@/$target_upgrade_version/g" jiva-target-svc-patch.tpl.
 if [[ "$replica_version" != "$target_upgrade_version" ]]; then
     echo "Upgrading Replica Deployment to $target_upgrade_version"
 
-    # Setting the update stratergy to recreate
-    setDeploymentRecreateStrategy $ns $r_dep
```
prateekpandey14: remove the same function call from the jiva target deployment.... line no 190

k8s/upgrades/0.8.1-0.8.2/cstor_volume_upgrade.sh
```diff
@@ -5,7 +5,7 @@
 #                                                              #
```
mittachaitu: Remove `setDeploymentRecreateStrategy` function. since the current upgrade script is supported only when user need to upgrade(cstor volumes) from 0.8.1 it's not required to check the strategy type for deployments.

Remove the `setDeploymentRecreateStrategy` function calls also(Line 151 and 152).

mittachaitu: @sonasingh46 Let's update README.md from (openebs-charts PR#### as reference) to point it to https://github.com/helm/charts/pull/12944

PR #2559, opened 2019-05-11
docs(examples): update demo yamls localpv hostpath

kmova:
This PR updates the examples of mongo and dbench
with option of using Local PV (hostpath)

Also updated the bench YAMLs with example on using
configurable Luworkers and QueueDepth parameters.

### DIFF
--- a/k8s/demo/dbench/sc-cstor-single-replica.yaml
+++ b/k8s/demo/dbench/sc-cstor-single-replica.yaml
@@ -1,0 +1,16 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: openebs-cstor-r1
+  annotations:
+    openebs.io/cas-type: cstor
+    cas.openebs.io/config: |
+      - name: StoragePoolClaim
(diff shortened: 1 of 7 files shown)


k8s/demo/dbench/sc-cstor-single-replica.yaml
```diff
@@ -0,0 +1,16 @@
...
+  annotations:
+    openebs.io/cas-type: cstor
+    cas.openebs.io/config: |
+      - name: StoragePoolClaim
+        value: "sparse-claim-auto"
+      - name: ReplicaCount
+        value: "1"
+      - name: TargetNodeSelector
+        value: |-
+            "kubernetes.io/hostname": "gke-kmova-helm-default-pool-a169cf59-kxfm"
```
prateekpandey14: May be we can add some comments here to change the hostname or put a info in README with details ?

kmova: I will incorporate this change in the next PR.. as this is already merged. Thank you!

PR #2601, opened 2019-05-30
fix(yaml,charts): add include-path in configmap for charts

akhilerm:
Added include-path in charts for path filter in config map. This will make the paths configurable from the charts.

### DIFF
--- a/k8s/charts/openebs/values.yaml
+++ b/k8s/charts/openebs/values.yaml
@@ -81,6 +81,7 @@ ndm:
     count: "1"
   filters:
     excludeVendors: "CLOUDBYT,OpenEBS"
+    includePaths: ""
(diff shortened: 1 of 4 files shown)


k8s/charts/openebs/values.yaml
```diff
@@ -81,6 +81,7 @@ ndm:
     count: "1"
   filters:
     excludeVendors: "CLOUDBYT,OpenEBS"
+    includePaths: ""
```
kmova: This option needs to be documented in the README.

akhilerm: @kmova  The readme for the charts right? Will do it .

akhilerm: Added in README.

PR #2606, opened 2019-05-31
chore(operator): add component label in admission-server pod

prateekpandey14:

### DIFF
--- a/k8s/openebs-operator.yaml
+++ b/k8s/openebs-operator.yaml
@@ -453,9 +453,6 @@ kind: Deployment
 metadata:
   name: openebs-admission-server
   namespace: openebs
-  labels:
-    app: admission-webhook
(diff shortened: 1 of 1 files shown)


k8s/openebs-operator.yaml
```diff
@@ -466,6 +464,7 @@ spec:
       labels:
         app: admission-webhook
```
kmova: app -> name : as per the convention followed for other deployments?

k8s/openebs-operator.yaml
```diff
@@ -454,8 +454,6 @@ metadata:
   name: openebs-admission-server
   namespace: openebs
   labels:
```
kmova: remove the labels - can cause error?

prateekpandey14: Yes..

PR #2613, opened 2019-06-03
fix(travis, blockdevice): add support to create cStor pools on block device

mittachaitu:

This PR is to fix travis in `openebs/maya` by adding support to create cStor pools on blockdevice

### DIFF
--- a/k8s/ci/test-script.sh
+++ b/k8s/ci/test-script.sh
@@ -69,19 +69,19 @@ sleep 10
 #echo "------------------ Deploy Pre-release features ---------------------------"
 #kubectl apply -f https://raw.githubusercontent.com/openebs/openebs/master/k8s/openebs-pre-release-features.yaml
 
-#echo "------------------------ Create block device claim --------------- "
-###TODO: Remove below snippet once manual BDC is supported to create cStor pools
-##### Get name of blockdevice to claim the blockdevice ####
-#block_device_name=$(kubectl get blockdevice -n ${NS} -o jsonpath='{.items[*].metadata.name}')
-#wget https://raw.githubusercontent.com/openebs/openebs/master/k8s/sample-pv-yamls/blockdeviceclaim.yaml -O temp_bdc.yaml
-#sed "s|sparse-1234|$block_device_name|g" temp_bdc.yaml > bdc.yaml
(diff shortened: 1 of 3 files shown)


k8s/ci/test-script.sh
```diff
@@ -69,17 +69,17 @@ sleep 10
...
+echo "------------------------ Create block device claim --------------- "
+##TODO: Remove below snippet once manual BDC is supported to create cStor pools
+#### Get name of blockdevice to claim the blockdevice ####
+block_device_name=$(kubectl get blockdevice -n ${NS} -o jsonpath='{.items[*].metadata.name}')
+wget https://raw.githubusercontent.com/openebs/openebs/master/k8s/sample-pv-yamls/blockdeviceclaim.yaml -O temp_bdc.yaml
+sed "s|sparse-1234|$block_device_name|g" temp_bdc.yaml > bdc.yaml
+kubectl apply -f bdc.yaml
+sleep 15
+block_device_status=$(kubectl get blockdevice -n ${NS} -o jsonpath='{.items[*].status.claimState}')
+echo "BlockDevice: ${block_device_name} Claimstate: ${block_device_status}"
```
vishnuitta: if status is not something what is expected, it can be errored out. This can be taken in next PR

k8s/ci/test-script.sh
```diff
@@ -69,19 +69,19 @@ sleep 10
...
-#sleep 15
-#block_device_status=$(kubectl get blockdevice -n ${NS} -o jsonpath='{.items[*].status.claimState}')
-#echo "BlockDevice: ${block_device_name} Claimstate: ${block_device_status}"
-#rm temp_bdc.yaml
-
-echo "------------------------ Create disk storagepoolclaim --------------- "
+echo "------------------------ Create block device claim --------------- "
+##TODO: Remove below snippet once manual BDC is supported to create cStor pools
+#### Get name of blockdevice to claim the blockdevice ####
+block_device_name=$(kubectl get blockdevice -n ${NS} -o jsonpath='{.items[*].metadata.name}')
```
vishnuitta: also, we need to make sure only one blockdevice exists (or) just pick the first one if there are many. This as well can be taken in next PR if we are creating only one sparse file as of now.

PR #2623, opened 2019-06-12
feat(autoprovisioning, blockdevice): add support to create cStor pools on sparse block device

mittachaitu:
Add support to create sparse based or disk based cstor pools.
User needs to set _`spec.type`_ type as **`sparse`** or **`disk`**.

### Auto Provisioning:
1. If cstor pools need to be created on _top of sparse disks_.
```yaml
kind: StoragePoolClaim
metadata:
  name: cstor-sparse
spec:
  name: cstor-sparse
  type: sparse
  maxPools: 3
  poolSpec:
    poolType: striped
    overProvisioning: false
```

2. If cstor pools need to be created _on top of disks attached to the node_.
```yaml
kind: StoragePoolClaim
metadata:
  name: cstor-disk
spec:
  name: cstor-disk
  type: disk
  maxPools: 3
  poolSpec:
    poolType: mirrored
    overProvisioning: false
```

### DIFF
--- a/k8s/sample-pv-yamls/spc-sparse-single.yaml
+++ b/k8s/sample-pv-yamls/spc-sparse-single.yaml
@@ -5,7 +5,7 @@ metadata:
   name: sparse-claim-auto
 spec:
   name: sparse-claim-auto
-  type: blockdevice
+  type: sparse
   maxPools: 1
   minPools: 1
   poolSpec:


ranjithwingrider:
I prefer to type as `block device` instead of `disk`

PR #2654, opened 2019-06-27
fix(upgrade) segregated verification steps from jiva volume upgrade script

shubham14bajpai:
Signed-off-by: shubham <shubham.bajpai@mayadata.io>

This PR separates the upgrade steps from verification steps so that verify can be re run.

### DIFF
--- a/k8s/upgrades/0.9.0-1.0.0/jiva/verify_volume_upgrade.sh
+++ b/k8s/upgrades/0.9.0-1.0.0/jiva/verify_volume_upgrade.sh
@@ -1,0 +1,146 @@
+#!/bin/bash
+
+target_upgrade_version=1.0.0
+is_upgrade_failed=0
+
+function usage() {
+    echo
+    echo "Usage:"
+    echo
+    echo "$0 <pv-name>"
+    echo
+    echo "  <pv-name> Get the PV name using: kubectl get pv"
+    exit 1
+}
+
+if [ "$#" -ne 1 ]; then
+    usage
+fi
+
+pv=$1
+
+echo "verifying jiva $pv volume upgrade"
+
+ns=$(kubectl get pv "$pv" -o jsonpath="{.spec.claimRef.namespace}")
+
+
+c_deploy_name=$(kubectl get deploy -n "$ns" \
+    -l openebs.io/persistent-volume="$pv",openebs.io/controller=jiva-controller \
+    -o jsonpath="{.items[*].metadata.name}" \
+)
+
+r_deploy_name=$(kubectl get deploy -n "$ns" \
+    -l openebs.io/persistent-volume="$pv",openebs.io/replica=jiva-replica \
+    -o jsonpath="{.items[*].metadata.name}" \
+)
+c_svc_name=$(kubectl get svc -n "$ns" \
+    -l openebs.io/persistent-volume="$pv" \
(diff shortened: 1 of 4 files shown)


k8s/upgrades/0.9.0-1.0.0/jiva/verify_volume_upgrade.sh
```diff
@@ -0,0 +1,135 @@
...
+    echo "  <pv-name> Get the PV name using: kubectl get pv"
+    exit 1
+}
+
+if [ "$#" -ne 1 ]; then
+    usage
+fi
+
+pv=$1
+
```
mittachaitu: We can print message saying `verifying jiva $pv volume upgrade`

shubham14bajpai: Done

k8s/upgrades/0.9.0-1.0.0/jiva/verify_volume_upgrade.sh
```diff
@@ -0,0 +1,140 @@
...
+        echo "Failed to upgrade Replica deployment $r_deploy_name images"
+        is_upgrade_failed=1
+    fi
+done
+
+#Verifying running status of controller and replica pods
+running_ctrl_pod_count=$(kubectl get pods -n "$ns" \
+-l openebs.io/controller=jiva-controller,openebs.io/persistent-volume="$pv" \
+--no-headers | wc -l)
+if [ "$running_ctrl_pod_count" != 1 ]; then
```
vishnuitta: do we need '"' here?
can you run shellcheck and see whether it is doing any suggestions

k8s/upgrades/0.9.0-1.0.0/jiva/verify_volume_upgrade.sh
```diff
@@ -0,0 +1,140 @@
...
+        -- bash -c "curl -s http://localhost:9501/v1/volumes" \
+        | grep -oP '("replicaCount":)[0-9]' | cut -d ':' -f 2
+    )
+    
+    retry=$(( retry+1 ))
+    sleep 5
+done
+
+if [ "$replica_count" != "$replication_factor" ]; then
+    echo "Failed to upgrade jiva replicas not registered"
```
vishnuitta: message can be like
'Failed validation for registered replica count.. expected $replication_factor but only $replica_count are registered`
check other error message whether it can be improved, instead of saying 'Failed to upgrade' as the start of error.. it can be like 'Failed validation..'

k8s/upgrades/0.9.0-1.0.0/jiva/verify_volume_upgrade.sh
shubham14bajpai: The `""` is suggested by shellcheck. 
```
Line 93:
if [ $running_ctrl_pod_count != 1 ]; then
     ^-- SC2086: Double quote to prevent globbing and word splitting.
```

k8s/upgrades/0.9.0-1.0.0/jiva/verify_volume_upgrade.sh
shubham14bajpai: Done

PR #2666, opened 2019-07-08
oep(ndm): proposal for unique disk identification

akhilerm:
Enhancement proposal for NDM. This proposal put forward a solution to uniquely identify a disk in a cluster.

### DIFF
--- a/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
+++ b/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
@@ -112,7 +112,7 @@ claim possibly leading to data loss
 
 1. We try to generate the UUID using the current implementation. 
 
-2. If the `ID_TYPE` is empty or `ID_MODEL` is `Virtual_disk` or `EphemeralDisk` 
+2. If the `ID_TYPE` is empty or `ID_MODEL` is `Virtual_disk` or `EphemeralDisk` and `ID_WWN` is empty 
 
     1. Check if the disk has a GPT label available on it.
     	- If a GPT label is available we will use only that label for UID generation
@@ -149,13 +149,13 @@ only partition is being wiped, it won't cause the labels to be removed.
 +-----------+
      |
      |
-+----v------------+        +-----------+      +-----------+      +-------------------+
-|    If           |        |  Check    |      |  Check    |      | Create GPT        |
-|ID_TYPE is empty |   Yes  | GPT label |  No  | FS label  |  No  | partition         |
-|    or           +------->+           +----->+           +----->+ table and         |
-|Virtual/Ephemeral|        |           |      |           |      | one partition     |
-|    Disk         |        |           |      |           |      | spanning the disk |
-+-----------------+        +-----------+      +-----------+      +-------------------+
++----v-------------+        +-----------+      +-----------+      +-------------------+
+|    If            |        |  Check    |      |  Check    |      | Create GPT        |
+|ID_TYPE is empty  |   Yes  | GPT label |  No  | FS label  |  No  | partition         |
+|    or ID_MODEL is+------->+           +----->+           +----->+ table and         |
+|Virtual/Ephemeral |        |           |      |           |      | one partition     |
+|Disk and empty WWN|        |           |      |           |      | spanning the disk |
++------------------+        +-----------+      +-----------+      +-------------------+
      |                        |                     |                  |
      |No                      |                     |                  |
      v                        |Yes                  |Yes               |
--- a/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
+++ b/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
@@ -23,16 +23,18 @@ status: provisional
     * [Goals](#goals)
     * [Non-Goals](#non-goals)
 * [Proposal](#proposal)
-    * [User Stories [optional]](#user-stories-optional)
-      * [Story 1](#story-1)
-      * [Story 2](#story-2)
-    * [Implementation Details/Notes/Constraints [optional]](#implementation-detailsnotesconstraints-optional)
-    * [Risks and Mitigations](#risks-and-mitigations)
+    * [User Stories(#user-stories)
+      * [Identify a physical disk](#identify-a-physical-disk)
+      * [Identify a virtual disk](#identify-a-virtual-disk)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+      * [Current Implementation](#current-implementation)
+      * [Shortcomings of current implementation](#shortcomings-of-current-implementation)
+      * [Proposed Implementation](#proposed-implementation)
+      * [Workflow](#worflow)
 * [Graduation Criteria](#graduation-criteria)
 * [Implementation History](#implementation-history)
-* [Drawbacks [optional]](#drawbacks-optional)
-* [Alternatives [optional]](#alternatives-optional)
-* [Infrastructure Needed [optional]](#infrastructure-needed-optional)
+* [Drawbacks](#drawbacks)
+* [Infrastructure Needed](#infrastructure-needed)
 
 ## Summary
 
@@ -136,7 +138,7 @@ NDM takes care of data clean up, after a BD is released from a BDC. i.e. a compl
 wipe of the disk will be done `wipefs -fa`. Since in case of NDM created GPT labels, 
 only partition is being wiped, it won't cause the labels to be removed.
 
-#### Workflow
+##### Workflow
 +-----------+
 |UID = WWN +|
 |    Model +|
@@ -165,10 +167,6 @@ only partition is being wiped, it won't cause the labels to be removed.
 +-----------+
 
 
-### Risks and Mitigations
-
-- 
-
 ## Graduation Criteria
 
 - NDM should be able to uniquely identify the disks, across reboots, across different
@@ -178,27 +176,20 @@ only partition is being wiped, it won't cause the labels to be removed.
 
 ## Implementation History
 
-Major milestones in the life cycle of a OEP should be tracked in `Implementation History`.
-Major milestones might include
+- Owner acceptance of `Summary` and `Motivation` sections - YYYYMMDD
+- Agreement on `Proposal` section - YYYYMMDD
+- Date implementation started - YYYYMMDD
+- First OpenEBS release where an initial version of this OEP was available - YYYYMMDD
+- Version of OpenEBS where this OEP graduated to general availability - YYYYMMDD
+- If this OEP was retired or superseded - YYYYMMDD
 
-- the `Summary` and `Motivation` sections being merged signaling owner acceptance
-- the `Proposal` section being merged signaling agreement on a proposed design
-- the date implementation started
-- the first OpenEBS release where an initial version of the OEP was available
-- the version of OpenEBS where the OEP graduated to general availability
-- when the OEP was retired or superseded
-
-## Drawbacks [optional]
+## Drawbacks
 
 - Do not have a mechanism to identify disks that are attached in multipath mode.
 - Cannot identify disks and create BDs if the same disk is connected to 2 nodes 
   at the same time.
 
-## Alternatives [optional]
-
-- NA
-
-## Infrastructure Needed [optional]
+## Infrastructure Needed
 
 - Test setup with different types of disks in different configurations to test and
   ensure the unique ID generation process.
--- a/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
+++ b/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
@@ -30,7 +30,7 @@ status: provisional
       * [Current Implementation](#current-implementation)
       * [Shortcomings of current implementation](#shortcomings-of-current-implementation)
       * [Proposed Implementation](#proposed-implementation)
-      * [Workflow](#worflow)
+      * [Workflow](#workflow)
 * [Graduation Criteria](#graduation-criteria)
 * [Implementation History](#implementation-history)
 * [Drawbacks](#drawbacks)
@@ -139,6 +139,7 @@ wipe of the disk will be done `wipefs -fa`. Since in case of NDM created GPT lab
 only partition is being wiped, it won't cause the labels to be removed.
 
 ##### Workflow
+```
 +-----------+
 |UID = WWN +|
 |    Model +|
@@ -165,7 +166,7 @@ only partition is being wiped, it won't cause the labels to be removed.
 |  Label    |
 |           |
 +-----------+
-
+```
 
 ## Graduation Criteria
 
--- a/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
+++ b/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
@@ -122,7 +122,7 @@ method to be used. As an inital step, 3 methods will be provided
 
 1. We try to generate the UUID using the current implementation. 
 
-2. If the `ID_TYPE` is empty or `ID_MODEL` is `Virtual_disk` or `EphemeralDisk` and `ID_WWN` is empty 
+2. If the `ID_TYPE` is empty or `ID_MODEL` is `Virtual_disk` or `EphemeralDisk` or `QEMU_HARDDISK` and `ID_WWN` is empty 
 
     1. Check if the disk has a GPT label available on it.
     	- If a GPT label is available we will use only that label for UID generation
--- a/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
+++ b/contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
@@ -9,7 +9,7 @@ owners:
   - "@vishnuitta"
 editor: "@akhilerm"
 creation-date: 2019-07-05
-last-updated: 2019-09-10
+last-updated: 2020-09-01
 status: provisional
 ---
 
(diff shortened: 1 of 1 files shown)


contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+* [Implementation History](#implementation-history)
+* [Drawbacks [optional]](#drawbacks-optional)
+* [Alternatives [optional]](#alternatives-optional)
+* [Infrastructure Needed [optional]](#infrastructure-needed-optional)
+
+## Summary
+
+This proposal brings out the design details to implement a Disk identification
+solution which can be used to uniquely identify a disk in the cluster. The main
+aim is to identify the data.
```
gila: The last part does not make sense to me. We are clearly speaking about disks here. Let us keep it at that?

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Motivation
+
+Uniquely identifying a disk across a cluster is a major challenge which needs to 
+be tackled. The identification is necessary, due to the fact that disk can move
+from node to node and even attached to different ports on the same node. The unique
+identification solution should help the NorthBound users of NDM to track the disk
+and thus the data movement within the cluster.
```
gila: *should help* -- should will help. 

(ps. Northbound? Can we dispense of the hipster talk and just say, consumers?)

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Goals
+
+- A unique disk identification mechanism that will work within the cluster on
+  disks having atleast a GPT label
```
gila: The fact being able to uniquely ID a disk is orthogonal to clusters. IOW, it will work across clusters too, not just within the same cluster.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+
+- Identification of disks without GPT labels
+- Identification of disks which removed the labels after claiming
+
+## Proposal
+
+### User Stories
+
+#### Identify a physical disk
+NDM should be able to generate a unqiue id for a physical disk attached to the node
```
gila: We should mention here how this is done:

for SATA disks: xxxx

for SAS disk: yyyyy

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+and this id should not change even if the disk is attached at a different port on the
+same host or moved to a different node.
+
+#### Identify a virtual disk
```
gila: not between nodes and clusters?

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+NDM should be able to generate a unique id for Virtual Disks(which may not have all 
+relevant information in VPD). This id should also persist between restarts of the node,
+attaching to a different machine or a different SCSI port of the same machine.
+
+### Implementation Details/Notes/Constraints
+
+The implementation details for the disk identification. The current implementation 
+works well for physical disks which have a unique WWN and model assigned by the
```
gila: WWN and/or  serial

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+NDM should be able to generate a unique id for Virtual Disks(which may not have all 
+relevant information in VPD). This id should also persist between restarts of the node,
+attaching to a different machine or a different SCSI port of the same machine.
+
+### Implementation Details/Notes/Constraints
+
+The implementation details for the disk identification. The current implementation 
+works well for physical disks which have a unique WWN and model assigned by the 
+manufacturer. But the implementation fails for `Virtual_disk`, because the fields
+are provider dependent and are not guaranteed to be unique. Also when the disk path
```
gila: The sentence does not seem to be correct.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+The following data is fetched from the disk: 
+1. `ID_WWN`
+2. `ID_MODEL`
+3. `ID_SERIAL_SHORT`
+4. `ID_VENDOR`
+5. `ID_TYPE`
+
+If the `ID_MODEL` is not `Virtual_disk` or `EphemeralDisk`:
+	- field 1-4 are appended
+	- md5 hash of the result is generated which will be UID for the disk by NDM
```
gila: If the disk has a WWN, its already unique isn't it?

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+	  eg: disk-xxxx-1 (/dev/sdX1), disk-xxxx-2 (/dev/sdX2). 
+	  This is guaranteed to be unique since the partition numbers won't change
+	  even if the device path changes
+2. If GPT label is not available on the disk, it can be logged to the user that
+   "the device couldn't be identified uniquely". Either NDM itself can take up
+   the GPT label creation part or ask the user to create a label so that the disk
+   UID can be generated 
+
+Data Cleanup:
+NDM takes care of data clean up, after a BD is released from a BDC. i.e. a complete
```
gila: I'm not sure why we wipe all labels and create new ones, why not only remove the partitions? or even better why not only remove the filesystems on the partitions?

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+- the `Summary` and `Motivation` sections being merged signaling owner acceptance
+- the `Proposal` section being merged signaling agreement on a proposed design
+- the date implementation started
+- the first OpenEBS release where an initial version of the OEP was available
+- the version of OpenEBS where the OEP graduated to general availability
+- when the OEP was retired or superseded
+
+## Drawbacks [optional]
+
+- Support will be available only for either physical disks or disks having
```
gila: physical disks that have .... (like serial)

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,184 @@
...
+  a GPT label on them. This will result in many disks that currently NDM supports
+  going into a not-supported mode.
+- In cases where the complete disk is formatted with a filesystem, as in the case of
+  local SSDs in GKE, there are not GPT labels available on the disk. This will result
```
gila: :s/not/no

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
akhilerm: we will wipe all labels on the blockdevice. either a partition or a complete device.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,196 @@
...
+    	- Once the GPT label is available, it is possible to identify partitions
+    	  also on the disk (if it exists). This will work by adding the partition
+    	  number along with the generated UID.
+    	  eg: disk-xxxx-1 (/dev/sdX1), disk-xxxx-2 (/dev/sdX2). 
+    	  This is guaranteed to be unique since the partition numbers won't change
+    	  even if the device path changes
+    2. Check if the disk has a Filesystem label available on it. This is in cases where the 
+       complete disk is formatted with a filesystem without a partition table. If filesystem
+       label is available we use it for generating the UUID.
+    3. If GPT and FS label is not available on the disk, we create a GPT partition table on the disk,
```
kmova: This proposal looks good. Has the GPT partition creation been prototyped? What additional changes need to be made to the NDM DaemonSet. Also as a note, we may want to mention when the NDM DaemonSet is deleted, the partition created will remain as is.

akhilerm: @kmova  The partition creation has not been prototyped yet.  For the NDM daemonset, whenever a UID generation fails, a create partition has to be invoked. We will not have to create UID for the partition in the same event, as it will come as a separate udev event.

Also, will mention that partition will remain on the disk, when daemonset is deleted as a drawback

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,196 @@
...
+- Owner acceptance of `Summary` and `Motivation` sections - YYYYMMDD
+- Agreement on `Proposal` section - YYYYMMDD
+- Date implementation started - YYYYMMDD
+- First OpenEBS release where an initial version of this OEP was available - YYYYMMDD
+- Version of OpenEBS where this OEP graduated to general availability - YYYYMMDD
+- If this OEP was retired or superseded - YYYYMMDD
+
+## Drawbacks
+
+- Do not have a mechanism to identify disks that are attached in multipath mode.
```
vishnuitta: isn't this the case even with current implementation?
How about limitation that north bound user of NDM will not be able to make any more partitions?

akhilerm: Added this into list of limitations

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,214 @@
...
+the device identified as a new disk. This will lead to data corruption as the disk that 
+was being claimed by some users has now came up with a new UID, but with the old data.
+Also, since this disk is now in an Unclaimed state, it will be available for others to
+claim possibly leading to data loss
+
+#### Proposed Implementation
+
+1. We try to generate the UUID using the current implementation. 
+
+2. If the `ID_TYPE` is empty or `ID_MODEL` is `Virtual_disk` or `EphemeralDisk` and `ID_WWN` is empty
```
pawanpraka1: why do we need to check ID_TYPE pr ID_MODEL? Should we just check if ID_WWN is empty here?

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,214 @@
...
+    	  number along with the generated UID.
+    	  eg: disk-xxxx-1 (/dev/sdX1), disk-xxxx-2 (/dev/sdX2). 
+    	  This is guaranteed to be unique since the partition numbers won't change
+    	  even if the device path changes
+    2. Check if the disk has a Filesystem label available on it. This is in cases where the 
+       complete disk is formatted with a filesystem without a partition table. If filesystem
+       label is available we use it for generating the UUID.
+    3. If GPT and FS label is not available on the disk, we create a GPT partition table on the disk,
+       and then create a single partition which spans the complete disk. This new partition will
+       be consumed by the users. This will help to identify a virtual disk for movements across
```
pawanpraka1: can you provide a sample example how the block device will look like after the partition and which path the data will be written..

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,214 @@
...
+
+#### Current Implementation
+The following data is fetched from the disk: 
+1. `ID_WWN`
+2. `ID_MODEL`
+3. `ID_SERIAL_SHORT`
+4. `ID_VENDOR`
+5. `ID_TYPE`
+
+If the `ID_MODEL` is not `Virtual_disk` or `EphemeralDisk`:
```
pawanpraka1: and "ID_TYPE is not empty" also.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,214 @@
...
+	- field 1-4 are appended
+	- md5 hash of the result is generated which will be UID for the disk by NDM
```
NajmudheenCT: What if the backing LUN is a shared resource and attached to 2 nodes.?

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
footplus: If we go the ID_MODEL identification route, we may add also `QEMU_HARDDISK`, which is presented by KVM-based hosts.

```suggestion
2. If the `ID_TYPE` is empty or `ID_MODEL` is `Virtual_disk` or `EphemeralDisk` or `QEMU_HARDDISK` and `ID_WWN` is empty 
```

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
akhilerm: In case of shared resource, like a single disk attached to 2 nodes. This design will not work? Because in the resulting object that we store in `etcd` will have 2 hostname and nodename for the same disk.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
akhilerm: For `QEMU_HARDDISK`, we will follow the same path as of `Virtual_Disk`, if we are unable to get a WWN

kmova:
Had a chat with @akhilerm on some possible upgrade scenarios to be considered. The design will be updated.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,237 @@
...
+|    Model +|
+|   Serial +|
+|   Vendor  |
+|           |
++-----------+
+     |
+     |
++----v-------------+        +-----------+      +-----------+      +-------------------+
+|    If            |        |  Check    |      |  Check    |      | Create GPT        |
+|ID_TYPE is empty  |   Yes  | GPT label |  No  | FS label  |  No  | partition         |
```
richardelling: Shared, dual-port NVMe devices might not have an ID_TYPE. They do have a WWN, serial, vendor, and model. We can infer they are NVMe by matching the DEVNAME, but I see no other udev environment variable that will identify them as being NVMe.

The current code sees lack of ID_TYPE and adds the hostname into the UID. This is a problem because the drive appears once per host. Is it sufficient to use a valid, non-known-to-be-ephemeral WWN and SERIAL?

akhilerm: In the new algorithm, we check for WWN and consider it to be unique. So shouldn't the problem with dual Port NVMe be mitigated?

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,237 @@
...
+|    or ID_MODEL is+------->+           +----->+           +----->+ table and         |
+|Virtual/Ephemeral |        |           |      |           |      | one partition     |
+|Disk and empty WWN|        |           |      |           |      | spanning the disk |
++------------------+        +-----------+      +-----------+      +-------------------+
+     |                        |                     |                  |
+     |No                      |                     |                  |
+     v                        |Yes                  |Yes               |
++-----------+                 |                     |                  |
+|if DEVTYPE |                 |                     |                  |
```
kmova: n_ : What happens if dev doesn't have partition. Probably goes to creating the partition and using that for generating the UUID.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
```diff
@@ -0,0 +1,237 @@
...
+  at the same time.
+- The partition created on the disk for identification will remain even after NDM
+  is uninstalled from the node.
+- Consumers of NDM will not be able to create a partition on blockdevices on which
+  NDM already created a partition for identification. 
+
+## Infrastructure Needed
+
+- Test setup with different types of disks in different configurations to test and
+  ensure the unique ID generation process.
```
kmova: The implementation looks good to proceed. As the implementation progresses, few negative scenarios need to be addressed before graduating this feature:
- What happens if the partition or filesystem is deleted? Even though this is not expected of the user, this kind of case should be easy to detect. 
- The upgrade strategy from older version device ids to the new IDs needs to be outlined.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
akhilerm: yes.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
richardelling: the new algorithm sounds good

akhilerm:
@kmova . Updated the OEP with latest details regarding implementation and upgrade of blockdevices.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
akhilerm: Done.

contribute/design/1.x/ndm/20190705-ndm-disk-identification.md
akhilerm: Yes. Only ID_WWN will be checked.

kmova (approved):
The design has been implemented in openebs 2.0.0 = with NDM 0.8.0.

PR #2678, opened 2019-07-18
WIP: chore(upgrade): pre upgrade scripts for 1.0 to 1.1 upgrade

akhilerm:
pre-upgrade scripts for 1.0 to 1.1 upgrade

### DIFF
--- a/k8s/upgrades/1.0.0-1.1.0/patch_remove_bdc_finalizer.json
+++ b/k8s/upgrades/1.0.0-1.1.0/patch_remove_bdc_finalizer.json
@@ -1,0 +1,3 @@
+[
+  {"op": "remove", "path": "/metadata/finalizers"}
+]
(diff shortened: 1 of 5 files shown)


k8s/upgrades/1.0.0-1.1.0/patch_remove_bdc_finalizer.json
```diff
@@ -0,0 +1,3 @@
+[
+  {"op": "remove", "path": "/metadata/finalizers"}
```
vishnuitta: 'm bit skeptical with deleting finalizers.. If finalizers are removed first and upgrade script breaks, then, we have a BDC without finalizers

kmova:
An alternate implementation to eliminate the need for pre-upgrade step is being experimented. Will revisit this PR.

kmova:
The pre-upgrades are mainly required to modify the existing custom resources with updated information like adding new finalizers. This is now implemented by owners of the respective custom resources. For example the cstor operator and localpv provisioner take care of adding the finalizer to BDC. ndm operator takes care of renaming the finalizer that it manages.

PR #2684, opened 2019-07-24
fix(jiva snap cleanup): exit when available snapshot is smaller than min-req-snap

shashank855:
This commit fix the following things:
- Exit when passed N or n is passed as an runtime argument.
- Exit when available snapshots is less than minimum-required-snapshot

**Special Note for reviewers**
- Kindly check the output message getting printed for the following condition 
 `if [ $(eval $snapshot_number_cmd) -le $(($min_required_snapshot + $number_of_snapshot)) ];`

### DIFF
--- a/k8s/jiva/snapshot-cleanup.sh
+++ b/k8s/jiva/snapshot-cleanup.sh
@@ -10,7 +10,7 @@ warning()
     echo "WARNING: Snapshot cleanup involves disconnecting the application from the storage. Also, while the snapshot cleanup is in progress - you will need to ensure that the connectivity to the Kubernetes Clusters is active. In case of unexpected disconnect, you will have to run the following command to restore the volume service. snapshot-cleanup.sh <pv-name> restore_service."
     echo "Do you want to continue (Y/N)"
     read access
-    if [ $access == "N" ]; then
+    if [ $access == "N" ] || [ $access == "n" ]; then
     exit 1;
     fi
 }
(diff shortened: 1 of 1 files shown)


k8s/jiva/snapshot-cleanup.sh
```diff
@@ -90,9 +96,9 @@ restart_ctrl()
 validate_snap_delete()
 {
     del_snapshot_name=$1
-    validate_cmd='kubectl exec -it $ctrl_pod_name -n $pvc_namespace -- jivactl snapshot ls | grep $del_snapshot_name'
+    validate_cmd="kubectl exec -it $ctrl_pod_name -n $pvc_namespace -- jivactl snapshot ls | grep $del_snapshot_name"
     eval $validate_cmd
-    if [ $? != "0" ]; then
+    if [ $? == "0" ]; then
```
utkarshmani1997: Good catch

k8s/jiva/snapshot-cleanup.sh
```diff
@@ -42,8 +42,14 @@ delete_jiva_snapshot()
        exit 1;
     fi
 
-    snapshot_number_cmd='kubectl exec -it $ctrl_pod_name -n $pvc_namespace -- jivactl snapshot ls | grep -v ID | wc -l'
-    snapshot_name_cmd='kubectl exec -it $ctrl_pod_name -n $pvc_namespace -- jivactl snapshot ls | grep -v ID | tail -1'
+    snapshot_number_cmd="kubectl exec -it $ctrl_pod_name -n $pvc_namespace -- jivactl snapshot ls | grep -v ID | wc -l"
+    snapshot_name_cmd="kubectl exec -it $ctrl_pod_name -n $pvc_namespace -- jivactl snapshot ls | grep -v ID | tail -1"
+    
+    if [ $(eval $snapshot_number_cmd) -le $(($min_required_snapshot + $number_of_snapshot)) ]; then
+        echo "Error: 0 snapshot present for deletion"
```
kmova: Remove the above echo. The next line actually helps with all the requirement information.

k8s/jiva/snapshot-cleanup.sh
```diff
@@ -42,8 +42,14 @@ delete_jiva_snapshot()
...
+        echo "Error: You have requested to delete $number_of_snapshot. There are only $(eval $snapshot_number_cmd) snapshots are present. Minimum $min_required_snapshot snpshots should be present to perform deletion"
```
kmova: `Error: You can initiate snapshot deletion only when the volume has more than $min_required_snapshot snapshots. There are only $(eval $snapshot_number_cmd) snapshots at the moment. You need not cleanup any more snapshots.`

PR #2689, opened 2019-07-26
fix(chart): make seachest probe configurable

akhilerm:
NDM seachest probe has been made configurable using `values.yaml`. By default it will be disabled.


k8s/charts/openebs/values.yaml
```diff
@@ -85,6 +85,8 @@ ndm:
     excludeVendors: "CLOUDBYT,OpenEBS"
     includePaths: ""
     excludePaths: "loop,fd0,sr0,/dev/ram,/dev/dm-,/dev/md"
+  probes:
```
kmova: @akhilerm -- add this to the  README table also please.

akhilerm: Added @kmova

PR #2709, opened 2019-08-14
fix(operator yaml): update deployment upgrade strategy as Recreate

mittachaitu:
This PR updates the control plane deployment upgrade strategy to `Recreate`.
When we set the upgrade strategy to `Recreate` deployment controller will
make sure during upgrade time(patching pod template spec) after deleting old pod
new pod with the changes will come to running state.

### DIFF
--- a/k8s/charts/openebs/templates/deployment-admission-server.yaml
+++ b/k8s/charts/openebs/templates/deployment-admission-server.yaml
@@ -13,7 +13,7 @@ metadata:
 spec:
   replicas: {{ .Values.webhook.replicas }}
   strategy:
-    type: {{ .Values.webhook.upgradeStrategy }}
+    type: "Recreate"
     rollingUpdate: null
   selector:
     matchLabels:
--- a/k8s/charts/openebs/templates/deployment-admission-server.yaml
+++ b/k8s/charts/openebs/templates/deployment-admission-server.yaml
@@ -12,6 +12,9 @@ metadata:
     openebs.io/version: {{ .Values.release.version }}
 spec:
   replicas: {{ .Values.webhook.replicas }}
+  strategy:
+    type: {{ .Values.webhook.upgradeStrategy }}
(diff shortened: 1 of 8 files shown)


k8s/charts/openebs/templates/deployment-local-provisioner.yaml
```diff
@@ -12,6 +12,9 @@ metadata:
     openebs.io/version: {{ .Values.release.version }}
 spec:
   replicas: {{ .Values.provisioner.replicas }}
+  strategy:
+    type: {{ .Values.provisioner.upgradeStrategy }}
```
kmova: Can this be changed to local PV provisioner. Also update the above replicas to use the localpvprovisioner (this has been fixed in upstream)

k8s/charts/openebs/templates/deployment-admission-server.yaml
```diff
@@ -12,6 +12,9 @@ metadata:
     openebs.io/version: {{ .Values.release.version }}
 spec:
   replicas: {{ .Values.webhook.replicas }}
+  strategy:
+    type: {{ .Values.webhook.upgradeStrategy }}
+    rollingUpdate: null
```
kmova: should rollingUpdate be set to null .. if the upgradeStrategy is set to  `Recreate`. Having said that, should we even give this option to select the upgradeStrategy or hardcode it to "Recreate" ?

k8s/charts/openebs/templates/deployment-local-provisioner.yaml
mittachaitu: I will update it

k8s/charts/openebs/templates/deployment-admission-server.yaml
mittachaitu: We will hardcode it to `Recreate`

PR #2732, opened 2019-09-04
refact(cspc-operator): added cspc crd yaml to cspc-operator

shubham14bajpai:
Signed-off-by: shubham <shubham.bajpai@mayadata.io>

This PR adds CSPC CRD yaml to cspc-operator for installation. The PR also includes a script to delete the CRD if present and apply the cspc-operator yaml.
**Note: Before applying the above operator yaml the old CRD needs to be deleted**

### DIFF
--- a/k8s/cspc_install.sh
+++ b/k8s/cspc_install.sh
@@ -10,4 +10,4 @@ fi
 
 echo "Installing cspc operator"
 
-kubectl apply -f cspc-operator.yaml
+kubectl apply -f https://raw.githubusercontent.com/openebs/openebs/master/k8s/cspc-operator.yaml 
--- a/k8s/cspc_install.sh
+++ b/k8s/cspc_install.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+echo "Running pre-installation steps"
+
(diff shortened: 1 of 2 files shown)


vishnuitta:
lets have a script to delete old CSPC crd yaml if no CR exists and applies the new one

k8s/cspc_install.sh
```diff
@@ -0,0 +1,13 @@
...
+
+x=$(kubectl get crd cstorpoolclusters.openebs.io -o jsonpath="{.spec.scope}" 2>&1 )
+
+if [[ $x == 'Cluster' ]]; then
+    kubectl delete crd cstorpoolclusters.openebs.io
+fi
+
+echo "Installing cspc operator"
+
+kubectl apply -f cspc-operator.yaml
```
vishnuitta: better to change it to full raw github link

vishnuitta: again, it depends on our doc steps..
if our doc says to download and run, then, this should be fine

shubham14bajpai: added the raw github link instead as there is nothing that user can customize in this yaml

PR #2760, opened 2019-10-04
Update README.md

adimessi:


kmova:
@adimessi .. could you open a new PR with some real fix. The whitespace is not required in this case.

PR #2769, opened 2019-10-07
updated README.md

SanyuktaSaha:
added more ways of installation of openebs.


README.md
```diff
@@ -55,7 +55,18 @@ kubectl apply -f https://openebs.github.io/charts/openebs-operator.yaml
 helm repo update
 helm install --namespace openebs --name openebs stable/openebs
 ```
+## Set cluster-admin user context and RBAC
```
kmova: @SanyuktaSaha Thanks for this PR! It will be good to push the specifics of creating the admin context into the docs itself. However can you put in a note above.. say on line 45, where you mention that the installation should be done with admin context. 

When providing an example in the top readme, it has to be very specific and need to work on all environments.

PR #2783, opened 2019-10-10
Update content security policy 

Apollo9999:
Content security policy details required


Apollo9999:
Please  ignore my commit.Unfortunately the commit changes made but not intentional

PR #2798, opened 2019-10-15
oep(cstor_ca): proposal for cstor to be CA aware

vishnuitta:
This OEP is having proposal for high level design to make cStor storage engine work with cluster autoscaler.

Signed-off-by: Vitta vitta@mayadata.io

### DIFF
--- a/contribute/design/1.x/autoscale/cstor/20191014-cstor-autoscale.md
+++ b/contribute/design/1.x/autoscale/cstor/20191014-cstor-autoscale.md
@@ -51,19 +51,20 @@ status: provisional
 
 ## Summary
 
-This proposal includes design for building blocks of cStor which can be
-triggered by an administrator or by an operator
+This proposal includes high level design for building blocks of cStor which can
+be triggered by an administrator or by an operator
 - to add a pool to existing cStor pool cluster
 - to remove a pool from cStor pool cluster with data consistency
 either when CA adds/wants-to-remove a node (or)
 admin adds/wants-to-remove a node.
+- move pool across nodes
 
 ## Motivation
 
 User wants OpenEBS to work natively with K8S cluster autoscaler. This minimizes
 operational cost.
 More usescases are available at:
-https://docs.google.com/document/d/1z_HdF7p_BNYO1MJnfJqFAHBrXEbqCjistyRXTpLbnFg/edit
+https://docs.google.com/document/d/1z_HdF7p_BNYO1MJnfJqFAHBrXEbqCjistyRXTpLbnFg/edit?usp=sharing
 
 ### Goals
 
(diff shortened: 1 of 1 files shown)


contribute/design/1.x/autoscale/cstor/20191014-cstor-autoscale.md
```diff
@@ -0,0 +1,339 @@
...
+Second is - by moving replica from old one to new replica
+
+Admin (or) operator need to identify the list of replicas i.e., CVRs on pool DP.
+For each identified CVR (CVR1), follow either the first mode or second.
+
+__Add a new replica and remove old one__ way:
+- Identify new pool (NP) other than DP which can take the replica
+- Perform replica scaleup steps
+  - create a new CVR (CVR2) on NP
+  - Increase the desired replication factor by 1 in CV CR related to this volume.
```
mittachaitu: So to support CA should we need to have more than 2 replicas?

contribute/design/1.x/autoscale/cstor/20191014-cstor-autoscale.md
```diff
@@ -0,0 +1,339 @@
...
+Above steps to move replica are also available at
+https://github.com/openebs/openebs/blob/master/contribute/design/1.x/replica_scaleup/dataplane/20190910-replica-scaleup.md
+
+#### Movement of pools on remote disks from one node to another
+
+This scenario is about moving pools on remote disks from old node to another,
+instead of distributing replicas on pool of old node to other pools.
+
+Steps to perform are as follows:
+- Identify the node N1 on which replicas of old pool doesn't exists
```
mittachaitu: Can we ignore this step? Are we supporting distributing volume replicas across different CSPC?
It leads to another way of understanding consider below case
1. Four node cluster and created CSPI(cStor pools) on all four nodes.
2. I have only single replica volumes in all the pools. (This point leads to the assumption of moving cspi1 to node3 which has cspi3).

contribute/design/1.x/autoscale/cstor/20191014-cstor-autoscale.md
```diff
@@ -0,0 +1,339 @@
...
+- Delete CSPI-mgmt deployment of old pool
```
mittachaitu: Deleting CSPI-mgmt deployment might not require if we just patch the node selector on deployment with node information. If we do this we can avoid last step `Create CSPI-mgmt deployment on new node N1`?

contribute/design/1.x/autoscale/cstor/20191014-cstor-autoscale.md
```diff
@@ -0,0 +1,339 @@
...
+- Patch CSPI CR with the new node information
+- Detach disks from old node and attach them to new node
+- Create CSPI-mgmt deployment on new node N1
+
+Another approach for the same would be:
+Let cStor pool pod i.e., CSPI-mgmt deployment consume disks using PVC instead
```
mittachaitu: It shouldn't be OpenEBS(cStor, jiva) PVC right?

PR #2807, opened 2019-10-22
oep(remove_replicas): proposal to enable user to remove replicas from volume

mittachaitu:
This PR adds a proposal to enable the user to remove replicas from a volume without IO disruption.


mittachaitu:
I think this PR needs to be merged
cc: @vishnuitta @kmova

PR #2810, opened 2019-10-23
Implementation of charts done to fetch articles data from server using Java script

Apollo9999:
Please implement d3.js chart library, and put chart.js script in static/js/ folder


kmova:
This is not relevant.

PR #2819, opened 2019-10-26
Thing Heading and subheading on the page need to highlighted

shahidafrid526:
made changes to the heading so that it looks attractive to thee users.

PR #2820, opened 2019-10-26
Thing Heading and subheading on the page need to highlighted

shahidafrid526:
made changes to the heading so that it looks attractive to the users.


kmova:
Similar to https://github.com/openebs/openebs/pull/2819

PR #2826, opened 2019-10-30
Update README.md

snehalbiju12:
Incorrect preposition use . Add 'of' before 'your'


kmova:
The correct text is correct @snehalbiju12. No need to change it.

PR #2850, opened 2019-12-06
Tighten OpenEBS RBAC permissions.

r0mant:
### What problem is being solved

The default set of RBAC permissions that comes with OpenEBS is very broad: it allows cluster-wide access to all core objects, ability to exec into any container and so on which can pose a security risk.

### How this change solves the problem

This change updates both `openebs-operator.yaml` and the Helm chart in an attempt to limit OpenEBS permissions in the following ways:

* Restrict CRUD permissions for core objects to OpenEBS specific namespace.
* Allow read-only access to some cluster-wide objects.
* Allow cluster-wide access to OpenEBS specific resources.
* Remove seemingly unnecessary permissions (more on that below).

### Notes

Note that I've removed a few permissions such as "ingresses", "horizontalpodautoscalers" and a couple of others that _seemed_ unnecessary - I searched through OpenEBS code and couldn't find any mention on those being used in any way - but if they are in fact needed for some use-cases, I'd be happy to bring them back (maybe a comment/explanation on what they're used for would be in order).

I have run a few tests with Local Provisioner (both host path and device modes) and cStor engine after making these changes and things worked fine. It would be helpful to run this through your CI as well to make sure nothing is missed.

Let me know what you think.

### DIFF
--- a/k8s/charts/openebs/templates/clusterrole.yaml
+++ b/k8s/charts/openebs/templates/clusterrole.yaml
@@ -9,57 +9,70 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 rules:
+# Allow OpenEBS read-only access to a few cluster-wide objects.
 - apiGroups: ["*"]
-  resources: ["nodes", "nodes/proxy"]
+  resources: ["nodes"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["*"]
+  resources: ["nodes/proxy"]
+  verbs: ["get"]
+- apiGroups: ["*"]
+  resources: ["namespaces"]
+  verbs: ["get", "watch"]
+# Allow OpenEBS to manage persistent storage cluster-wide.
+- apiGroups: ["*"]
+  resources:
+  - storageclasses
+  - persistentvolumeclaims
+  - persistentvolumes
   verbs: ["*"]
+# Allow OpenEBS to create events.
 - apiGroups: ["*"]
-  resources: ["namespaces", "services", "pods", "pods/exec", "deployments", "deployments/finalizers", "replicationcontrollers", "replicasets", "events", "endpoints", "configmaps", "secrets",  "jobs", "cronjobs" ]
-  verbs: ["*"]
-- apiGroups: ["*"]
-  resources: ["statefulsets", "daemonsets"]
-  verbs: ["*"]
-- apiGroups: ["*"]
-  resources: ["resourcequotas", "limitranges"]
-  verbs: ["list", "watch"]
-- apiGroups: ["*"]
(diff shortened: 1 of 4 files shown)


prateekpandey14:
@r0mant thanks for doing this, we are checking some of the permission issue related to Jiva and other k8s distribution including like GKE, openshift, AKS etc.
We will get back to you with changes if any thing required here.

prateekpandey14:
@r0mant sorry for the delay, started looking into the above changes, the RBAC clusterwide permission for all the k8s resources has been limited to `openebs` namespace , which will not work in case of **Jiva** and **Local-PV** as they can be deployed in any namespace as well, other then `openebs`.

I am currently looking into this and will get back with more comments.

prateekpandey14:
Some more updates based on GKE and  openshift  based clusters:

- Able to create cStor volumes successfully in openebs namespace.
- Able to create local PV using hostpath successfully.
- Unable to provision jiva volumes in any namespace other than openebs
- Unable to provision localPV using device.

- ( specific to openshift) Admission webhook crashing due to permission issue required to set the  deployemnt ownerReference inn webhook resoureces in openshift based clusters
```go
F1227 01:08:13.164069       1 main.go:85] failed to create validator{openebs-validation-webhook-cfg}: validatingwebhookconfigurations.admissionregistration.k8s.io "openebs-validation-webhook-cfg" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>failed to initialize validation server
```

akhilerm:
When provisioning fails, the controller tries to patch that information onto the PVC. This will result in 
```
'events "demo-vol1-claim.15e612138a6253cb" is forbidden: User "system:serviceaccount:openebs:openebs-maya-operator" cannot patch resource "events" in API group "" in the namespace "giri"' (will not retry!)
```
So we will need patch permission also for updating the events.

kmova:
For fixing the issues observed with Jiva, the following enhancements have to be made to the provisioners:
- Set the `DeployInOpenEBSNamespace` as default for Jiva Storage Classes. 
- Once upgraded to the new version with reduced RBAC policies, API server will not be able to operate on the Jiva volumes created in another namespace. So Jiva pods should be migrated from application namespace to openebs namespace prior to upgrading with the changes in this PR. 

Keeping this on hold, to ensure backward compatibility is maintained. 

However for fresh installations, the steps provided in this PR are good as long as the Jiva Storage Classes are configured with "DeployInOpenEBSNamespace" property as follows:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: jiva-pods-in-openebs-ns
  annotations:
    openebs.io/cas-type: jiva
    cas.openebs.io/config: |
      - name: DeployInOpenEBSNamespace
        enabled: "true"
provisioner: openebs.io/provisioner-iscsi
```

akhilerm:
@prateekpandey14 @kmova Were the backward compatibility issues for deploying jiva in openebs fixed?

PR #2865, opened 2019-12-21
docs(examples): add busybox based examples and cleanup

kmova:
Added busybox example YAMLs. 
Clean-up unused YAMLs and fix the references to the latest storage classes in some sample YAMLs.


k8s/demo/busybox/localpv-device-pod.yaml
```diff
@@ -0,0 +1,60 @@
...
+# limitations under the License.
+#
+# Example for launching a busybox pod using localpv with device.
+# 
+# This example uses the default localpv with hostdevice 
+# storage class installed by OpenEBS. 
+#
+# Prior to running this YAML, verify that the storage class 
+# mentioned below is available. 
+#
```
gprasath: Can we add one more line saying that a device matching the storage requirement should be attached to node

PR #2882, opened 2020-01-10
feat(baseDir, yaml): add support to take baseDir as configurable value in operator yaml

mittachaitu:
This PR adds a persistent storage path to store OpenEBS components related files(currently used for storing only core files).


mittachaitu:
As per discussion, changes are taking in OpenEBS version 1.7

PR #2892, opened 2020-01-20
chore(travis): add overprovisioning artifacts and test case for travis

sonasingh46:
This PR adds an example of SPC yaml that has overprovisioning disabled.
It also adds other artifacts to add overprovisioning test cases in Travis. 
This will be used in overprovisioning test case run-in Travis.
Signed-off-by: Ashutosh Kumar <ashutosh.kumar@mayadata.io>


k8s/ci/test-script.sh
```diff
@@ -79,6 +79,71 @@ sleep 10
...
+		else
+      			echo "Waiting for few iterations to check that PVC test-pvc-10gigs does not get bound after sometime"
+                        kubectl get pvc test-pvc-10gigs
+			if [ "$i" == "$PVC10G_MAX_RETRY" ] && [ "$PVC1GStatus" != "Bound" ]; then
+				echo "PVC test-pvc-10gigs NOT bound and hence test case passed"
+			fi
+		fi
+      			sleep 5
+		done
+}
```
prateekpandey14: above functions are reduntent can we make it generic an use it to check, like 
`checkForPVCStatus() <pvc-name> `

k8s/ci/test-script.sh
```diff
@@ -79,6 +79,71 @@ sleep 10
...
+
+echo "------------------------ Patch ndm daemonset to set SPARSE_FILE_COUNT to 2 --------------- "
+kubectl patch ds openebs-ndm -n openebs --patch "$(cat patch.yaml)"
+
+sleep 10
+
+echo "Create PVC with 1G capacity request "
+kubectl apply -f https://raw.githubusercontent.com/openebs/openebs/master/k8s/overprovisioning-artifacts/spc/pvc1g.yaml
+echo "Create PVC with 10G capacity request "
+kubectl apply -f https://raw.githubusercontent.com/openebs/openebs/master/k8s/overprovisioning-artifacts/spc/pvc10g.yaml
```
kmova: Can the files be under the `k8s/ci/overprovisioning` directory?

k8s/ci/test-script.sh
sonasingh46: The exit criteria are different for both the functions. 
Putting both in the same function will cause bloating up of the single function due to
if conditions for deciding the exit criteria based on PVC name and the same code has to be written again.

PR #2900, opened 2020-01-28
oep(snapshot): updating the OEP with snapshot and clone details.

pawanpraka1:
Adding the details how ZFSPV will take snapshot and clone

Signed-off-by: Pawan <pawan@mayadata.io>

### DIFF
--- a/contribute/design/1.x/csi/20190805-csi-zfspv-volume-provisioning.md
+++ b/contribute/design/1.x/csi/20190805-csi-zfspv-volume-provisioning.md
@@ -327,13 +327,18 @@ spec:
 status:
   state: Ready
 ```
+### 5.1 Alternate Design
 
 We can use GRPC call also to create the snapshot, The controller plugin will call node plugin's grpc server to create the snapshot. Creating a ZFSSnapshot custom resource has advantages :-
 
 - Even if Node plugin is down we can get the info about snapshots
 - No need to do zfs call everytime to get the list of snapshots
 - Controller plugin and Node plugin can run without knowledge of each other
-- Reconciliation is easy to manage
+- Reconciliation is easy to manage as K8s only supports a reconciliation based snapshots as of today. For example, the operation to create a snapshot is triggered via K8s CR. Post that, K8s will be re-trying till a successful snapshot is created. In this case, having an API doesn't add much value.
+
+However, there will be use cases where snapshots have to be taken while the file system is frozen to get an application-consistent snapshot. This can required that a snapshot API should be available with blocking API calls that can be aborted after an upper time limit.
+
+Maybe, in a future version, we should add a wrapper around the snapshot functionality and expose it through node agent API. This can be tracked in the backlogs.
 
 ### 6. CSI Clone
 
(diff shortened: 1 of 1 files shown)


contribute/design/1.x/csi/20190805-csi-zfspv-volume-provisioning.md
```diff
@@ -261,6 +258,123 @@ When CSI will get volume destroy request, it will destroy the created zvol and a
...
+it will create a ZFSSnapshot custom resource. This custom resuorce will get all the details from the ZFSVolume CR, ownerNodeID will be same here as snapshot will also be there on same node where the original volume is there. The controller will create the request with status as Pending and set a label "openebs.io/volname" to the volume name from where the snapshot has to be created.
+
+```yaml
+apiVersion: openebs.io/v1alpha1
+kind: ZFSSnapshot
+metadata:
+  name: zfspv-snap
+  namespace: openebs
+  labels:
+    openebs.io/volname: pvc-34133838-0d0d-11ea-96e3-42010a800114
```
kmova: adding the details of the volumesnapshot that created this snap will be useful. 
Also let us keep the labels uniform with other engines - openebs.io/persistent-volume.

contribute/design/1.x/csi/20190805-csi-zfspv-volume-provisioning.md
```diff
@@ -261,6 +258,123 @@ When CSI will get volume destroy request, it will destroy the created zvol and a
 ### 4. CSI ZFSPV property change
 There will be a watcher watching for this ZFSVolume custom resource in the agent. We can update the ZFSVolume custom resource with the desired property and the watcher of this custom resource will apply the changes to the corresponding volume.
 
+### 5. CSI SnapShot
+
```
kmova: Could you add a note on why this approach of CR was selected over - possibly having the CSI controller directly invoking the snapshot command on the corresponding CSI node-agent.

pawanpraka1: added, Please check.

contribute/design/1.x/csi/20190805-csi-zfspv-volume-provisioning.md
pawanpraka1: updated the label. Volumesnapshot yaml is already there.

contribute/design/1.x/csi/20190805-csi-zfspv-volume-provisioning.md
```diff
@@ -261,6 +258,141 @@ When CSI will get volume destroy request, it will destroy the created zvol and a
...
+  blocksize: 4k
+  capacity: "4294967296"
+  compression: "on"
+  dedup: "on"
+  ownerNodeID: zfspv-node1
+  poolName: zfspv-pool
+status:
+  state: Ready
+```
+
```
kmova: It will be good to push this into a sub-section called 5.1 and name it: Alternate Design. 

In addition to the notes currently mentioned, it would be good to include the following information as well: 
It appears that K8s only supports a reconciliation based snapshots as of today. For example, the operation to create a snapshot is triggered via K8s CR. Post that, K8s will be re-trying till a successful snapshot is created. In this case, having an API doesn't add much value. 

However, there will be use cases where snapshots have to be taken while the file system is frozen to get an application-consistent snapshot. This can required that a snapshot API should be available with blocking API calls that can be aborted after an upper time limit. 

Maybe, in a future version, we should add a wrapper around the snapshot functionality and expose it through node agent API. This can be tracked in the backlogs.

PR #2909, opened 2020-02-11
travis(readonly,spc): test case for pool read-only threshold limit for SPC/CSP

mynktl:
Signed-off-by: mayank <mayank.patel@mayadata.io>


k8s/ci/test-script.sh
```diff
@@ -100,6 +100,8 @@ wget https://raw.githubusercontent.com/openebs/openebs/master/k8s/ci/overprovisi
 echo "------------------------ Patch ndm daemonset to set SPARSE_FILE_COUNT to 2 --------------- "
 kubectl patch ds openebs-ndm -n openebs --patch "$(cat patch.yaml)"
 
+kubectl apply -f https://raw.githubusercontent.com/openebs/openebs/master/k8s/sample-pv-yamls/spc-sparse-single.yaml
```
prateekpandey14: we are already using the same `spc-sparse-single.yaml` to create spc based pool at line 76, i think we can use  the same pool to compare the threshold values.

PR #2935, opened 2020-02-19
oep(ndm): proposal for unique disk accessing

mittachaitu:
Enhancemet proposal for NDM. This proposal has a solution to uniquely
access the disks in a cluster.

### DIFF
--- a/contribute/design/1.x/ndm/20200219-ndm-disk-accessing.md
+++ b/contribute/design/1.x/ndm/20200219-ndm-disk-accessing.md
@@ -1,0 +1,175 @@
+---
+oep-number: NDM 0002
+title: Uniquely accessing disk in cluster
+authors:
+  - "@mittachaitu"
+owners:
+  - "@gila"
+  - "@kmova"
+  - "@vishnuitta"
+  - "@pawanpraka1"
+  - "@akhilerm"
+editor: "@mittachaitu"
+creation-date: 2020-02-19
+last-updated: 2020-02-19
+---
+
+# Disk Identification using NDM
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [User Stories](#user-stories)
+      * [Accessing Uniquely Identified Disk](#accessing-uniquely-identified-disk)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    * [Current Implementation](#current-implementation)
+    * [Shortcomings of current implementation](#shortcomings-of-current-implementation)
+    * [Proposed Implementation](#proposed-implementation)
+    * [Workflow](#workflow)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+* [Drawbacks](#drawbacks)
+* [Infrastructure Needed](#infrastructure-needed)
+
+## Summary
+
+This proposal brings out the design details to implement a uniquely access the
+uniquely identified disks across the cluster.
+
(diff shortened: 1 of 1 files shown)


contribute/design/1.x/ndm/20200219-ndm-disk-accessing.md
```diff
@@ -0,0 +1,184 @@
+---
+oep-number: NDM 0002
+title: Uniquely accessing disk in cluster
```
vishnuitta: Unique access path of any kind of disks for NDM consumers

vishnuitta: Fixed access path to disks for NDM consumers

contribute/design/1.x/ndm/20200219-ndm-disk-accessing.md
```diff
@@ -0,0 +1,184 @@
...
+- Configurable device accessing mechanism.
+- A unique disk accessing mechanism that will work across the clusters.
+
+### Non-Goals
+
+- NA
+
+## Proposal
+
+### User Stories
```
vishnuitta: what about user stories like:
- removing the requirement to run in Privileged mode for NDM consumers
- removing the need for restart of app pod to access disks that are attached after non-privileged pod came up

contribute/design/1.x/ndm/20200219-ndm-disk-accessing.md
```diff
@@ -0,0 +1,184 @@
...
+  claimed disks in the specified path.
+
+##### Advantages
+
+- Application can uniquely access the disks.
+- Application doesn't required to run in privileged mode to access the disks.
+- Dynamically attached disks can be easily consumed by the application without
+  restarting application pod.
+- Application consuming multiple blockdevices no need to mount all of them (or)
+  no need to mount the /dev directory inside the application.
```
akhilerm: @mittachaitu . Suppose the application creates a partition on the disk. How will the application access partition it created ?

contribute/design/1.x/ndm/20200219-ndm-disk-accessing.md
```diff
@@ -0,0 +1,184 @@
...
+  - "@kmova"
+  - "@vishnuitta"
+  - "@pawanpraka1"
+  - "@akhilerm"
+editor: "@mittachaitu"
+creation-date: 2020-02-19
+last-updated: 2020-02-19
+---
+
+# Disk Identification using NDM
```
akhilerm: Disk `Accessing` using NDM

contribute/design/1.x/ndm/20200219-ndm-disk-accessing.md
mittachaitu: It will be possible with the following approach but not sure how far it is correct:
1. Given block file(pointing to entier disk) to the application by mounting blockfile directory(which contains the only disk it can access) into the container.
2. Application created partitions on that disk.
3. The application has to claim a partitioned block device(by specifying the same access path whatever specified to claim entier disk).
4. Now the path to access the partition exists inside the container. 
Not sure whether it is correct or not just kept my thoughts.

mittachaitu:
> What will happen if the disk moves from one node to another. ?
> i.e The blockdevice was claimed. But it moved to another node. Should the unique access path be created there also?

Yes, access path needs to be created there also and the application also should be moved to that node to access that disk.

contribute/design/1.x/ndm/20200219-ndm-disk-accessing.md
mittachaitu: Currently in Kubernetes world accessing partitioned disk that was created by the application itself is not straight forward.

akhilerm:
Unique disk accessing feature is not yet implemented and the design is still in the review phase.

PR #2997, opened 2020-04-13
chore(operator): update liveness probes regular expression

prateekpandey14:
Refactor the liveness check to check for the exact command name instead of using `pgrep -f`. 

`pgrep -f` can result in false positive is there another process with some arguments that can match the command name. 

**Note:** (from the pgrep man pages)
The process name used for matching is limited to the 15 characters present in the output of `/proc/pid/stat`.

- So updated the regular expression with 15 characters command name ( if applicable) for better handling.
- for NDM daemontset just using `ndm`  with pgrep as its only 3 char cmd name and don't need any regular expression.

### DIFF
--- a/k8s/openebs-operator.yaml
+++ b/k8s/openebs-operator.yaml
@@ -315,8 +315,10 @@ spec:
         livenessProbe:
           exec:
             command:
-            - pgrep
-            - "^openebs-provisi.*"
+            - sh
+            - -c
+            - pcount=$(pgrep "^openebs-provisi.*" | wc -l); if [ $pcount != 1 ]; then exit
+              1; else exit 0; fi;
           initialDelaySeconds: 30
           periodSeconds: 60
 ---
@@ -357,8 +359,10 @@ spec:
           livenessProbe:
             exec:
               command:
-              - pgrep
-              - "^snapshot-contro.*"
+              - sh
+              - -c
+              - pcount=$(pgrep "^snapshot-contro.*" | wc -l); if [ $pcount != 1 ]; then exit
+                1; else exit 0; fi;
             initialDelaySeconds: 30
             periodSeconds: 60
         # OPENEBS_MAYA_SERVICE_NAME provides the maya-apiserver K8s service name,
@@ -384,8 +388,10 @@ spec:
           livenessProbe:
             exec:
               command:
-              - pgrep
-              - "^snapshot-provis.*"
+              - sh
+              - -c
+              - pcount=$(pgrep "^snapshot-provis.*" | wc -l); if [ $pcount != 1 ]; then exit
+                1; else exit 0; fi;
             initialDelaySeconds: 30
             periodSeconds: 60
 ---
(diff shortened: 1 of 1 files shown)


k8s/openebs-operator.yaml
```diff
@@ -512,8 +509,7 @@ spec:
           exec:
             command:
             - pgrep
-            - -f
-            - ".*ndm"
+            - "^ndm.*"
```
vishnuitta, line 512: do we need regular expression here also?

prateekpandey14: command name is only `ndm` , so i think we don't need * at the end

k8s/openebs-operator.yaml
```diff
@@ -512,8 +515,7 @@ spec:
           exec:
             command:
             - pgrep
-            - -f
-            - ".*ndm"
+            - "ndm"
```
prateekpandey14, line 518: @kmova @akhilerm just using `ndm` to grep here as its only 3 char cmd name.

akhilerm: yes. That will work

kmova: how about we keep it consistent with other liveness checks.

k8s/openebs-operator.yaml
```diff
@@ -315,9 +315,10 @@ spec:
         livenessProbe:
           exec:
             command:
-            - pgrep
-            - -f
-            - ".*openebs"
+            - sh
+            - -c
```
pawanpraka1, line 319: how about ```test `pgrep -c ^openebs-provisi.*` == 1```

k8s/openebs-operator.yaml
```diff
@@ -592,7 +615,19 @@ spec:
             - name: OPERATOR_NAME
               value: "node-disk-operator"
             - name: CLEANUP_JOB_IMAGE
-              value: "quay.io/openebs/linux-utils:1.4.0"
+              value: "openebs/linux-utils:3.9"
```
kmova: This image should be quay.io/openebs/linux-utils:1.9.0 or quay.io/openebs/linux-utils:ci.

prateekpandey14: yes removed the change in earlier commit

k8s/openebs-operator.yaml
```diff
@@ -312,12 +312,19 @@ spec:
         # This is supported for openebs provisioner version 0.5.3-RC1 onwards
         #- name: OPENEBS_MAYA_SERVICE_NAME
         #  value: "maya-apiserver-apiservice"
+         # Process name used for matching is limited to the 15 characters
+         # present in the pgrep output.
+         # So fullname can't be used here with pgrep (>15 chars).A regular expression
+         # that matches the enter command name has to specified.
```
akhilerm, line 318: typo. `entire command`

prateekpandey14: fixed in new commit

k8s/openebs-operator.yaml
prateekpandey14: changed to `test `pgrep -c ^openebs-provisi.*` = 1` ( test support single `=` expression example string1 = string2 )

PR #3019, opened 2020-04-27
chore(YAML): Add minio spec using Jiva StoragePool

ranjithwingrider:
- Add YAML spec for Minio-Standalone using default Jiva SC

Signed-off-by: Ranjith R <ranjith.raveendran@mayadata.io>

### DIFF
--- a/k8s/demo/minio/minio-standalone-jiva-storagepool.yaml
+++ b/k8s/demo/minio/minio-standalone-jiva-storagepool.yaml
@@ -49,7 +49,7 @@ metadata:
     openebs.io/target-affinity: minio      
     app: minio-storage-claim1
 spec:
-  storageClassName: openebs-jiva-gpd-3repl
+  storageClassName: openebs-jiva-3rep
   accessModes:
     - ReadWriteOnce
   resources:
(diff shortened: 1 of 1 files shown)


k8s/demo/minio/minio-standalone-jiva-storagepool.yaml
```diff
@@ -0,0 +1,73 @@
...
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: minio-pv-claim1
+  labels:
+    openebs.io/target-affinity: minio      
+    app: minio-storage-claim1
+spec:
+  storageClassName: openebs-jiva-gpd-3repl
```
muratkars, line 52: Why do we have gpd in the storageclass name? (openebs-jiva-gpd-3repl)
Please point me if this is mentioned somewhere otherwise we should rename it `openebs-jiva-3repl` since it doesn't matter which cloud vendor was used.

ranjithwingrider: Yes. You are correct. It is not intentional. I was taken sample YAML file from docs. I will rename it..

PR #3029, opened 2020-05-07
chore(yaml): update operator yaml

akhilerm:
- add ndm feature-gate flag
- update linux-utils to latest image

### DIFF
--- a/k8s/openebs-operator.yaml
+++ b/k8s/openebs-operator.yaml
@@ -487,6 +487,11 @@ spec:
       containers:
       - name: node-disk-manager
         image: quay.io/openebs/node-disk-manager-amd64:ci
+        args:
+          - -v=4
+        # The feature-gate is used to enable the new UUID algorithm.
+        # This is a feature currently in Alpha state
+        #  - --feature-gates="GPTBasedUUID"
         imagePullPolicy: Always
         securityContext:
           privileged: true
@@ -612,7 +617,7 @@ spec:
             - name: OPERATOR_NAME
               value: "node-disk-operator"
             - name: CLEANUP_JOB_IMAGE
-              value: "quay.io/openebs/linux-utils:1.4.0"
+              value: "quay.io/openebs/linux-utils:1.9.0"
           # Process name used for matching is limited to the 15 characters
           # present in the pgrep output.
           # So fullname can be used here with pgrep (cmd is < 15 chars).


k8s/openebs-operator.yaml
```diff
@@ -525,7 +530,7 @@ spec:
           value: "10737418240"
         # Specify the number of sparse files to be created
         - name: SPARSE_FILE_COUNT
-          value: "1"
+          value: "0"
```
prateekpandey14, line 533: changing to '0' will break the CI in maya and other repos, let it be '1'.  We are mantaining the count '0' in charts and release yamls that would be good enough.

akhilerm: Done.

k8s/openebs-operator.yaml
```diff
@@ -45,7 +45,7 @@ rules:
   resources: ["customresourcedefinitions"]
   verbs: [ "get", "list", "create", "update", "delete", "patch"]
 - apiGroups: ["*"]
-  resources: [ "disks", "blockdevices", "blockdeviceclaims"]
+  resources: ["blockdevices", "blockdeviceclaims"]
```
kmova, line 48: @akhilerm -- let us retain the option to `disks`. And probably => have some code in the upcoming or this release, that actually deletes the older disks?

akhilerm: @kmova  But we removed the controller for disk, and deleted the APIs.

kmova: yep.. as part of m-apiserver upgrade, can we clean up old stale entries or CRDs. were there any finalizers on disk resources?

kmova: If we don't do this, then customers might have disk entries that need to be manually cleaned up.. right?

akhilerm: > were there any finalizers on disk resources?

No

akhilerm: > If we don't do this, then customers might have disk entries that need to be manually cleaned up.. right?

Yes. Entries will be there. But in that case. It would have been better if we removed entries as part of reconciliation and then in another release, remove the controller.

kmova: yes, we missed to the above. As we move from 1.x releases of openebs into 2.0, we will have to have users first upgrade to the last stable release of 1.x. In which we need to catch and cleanup all these things. When we start tagging 2.0, we can remove all the deprecated resources.

akhilerm: But for "catching up" those disk resources, we need the API right? We removed it in this release (1.10).

akhilerm: Done.

Issue #2946, opened 2020-02-29
kubectl plugin to query openebs component status


kmova:
OpenEBS is completely Kubernetes native and is implemented using microservices. The microservices are deployed as Kubernetes deployments, statefulset or daemon sets and so forth. 

In order to know the status of the components, a user has to run multiple kubectl commands to check that all the OpenEBS related microservices are up and running. 

For instance, a cStor volume comprises of one or more cStor Pool Deployments and a cStor Target Deployment. To know the status of a cStor volume, the user is expected to run the following kubectl commands:
- `kubectl get cstorvolume -n openebs` (check for the status of the volume and if not-healthy, check for the replica and target status. 
- `kubectl get pods -n openebs` to check if the replica and target pods are running
- `kubectl get csp -n openebs` and `kubectl get cvr -n openebs` for checking if the replica has any additional issues, if the node that was running a csp might be down and so forth. 

The proposal is to provide a kubectl plugin for openebs that will allow running simple CLI commands that can perform required kubectl queries and present the information in an easy-to-consume manner. 

Using kubectl plugin, it would be great if a user can type a command like:

`kubectl openebs volume-status --name <cstor-pv-name>`

The output for this command could be something like:

```
Volume Status
----------------
Name              :   pvc-dc3cb979-51ec-11e9-803f-42010a800179
PVC Name          :   cstor-volume
Capacity          :   8G
Used              :   5G
Target Status     :   Running
Target Node       :   node2
Replica Count     :   3

Replica Details :
-----------------
NAME          STATUS      POOL NAME           NODE    POOL CAPACITY
----          ------      ---------           -----   -------------
replica-1     Running     cstor-pool-ejs2     node1   50%
replica-2     Running     cstor-pool-gf1d     node2   80% 
replica-3     Running     cstor-pool-m8cy     node3   60%


Application Pods using the Volume
----------------------------------
NAME          STATUS      NODE 
----          ------      -----
demo-app      Running     node1
```

Similarly, there can be many other CLI options like:

`kubectl openebs cstor-status --name <cstor-pool-cluster-name>`
`kubectl openebs status`
`kubectl openebs upgrade volume --name <cstor-volume-name> --to-verstion <new-version>`

Raman-Kumar:
@kmova 
Can you suggest me some docs or resources to properly understand 
the **codebase** of **openebs** and **contributing** to it?

kmova:
Hi @Raman-Kumar -- this is a good place to start: 
[...]

PR #3035, opened 2020-05-13
A draft proposal for a CLI tool

vaniisgh:
A draft proposal to tackle #2946

Hey everyone!
I have recently learnt about CAS and OpenEBS. I'm new to Cloud-Native development but even I can see that OpenEBS is a great solution for containerized applications to store their data. 
I saw the issue on the Community Bridge ideas page and would love to help with the development of a CLI tool in any way possible. I have scraped together a draft proposal for the same. In the next few days, I shall try my best to refine it to bring it up to a good standard as well as maybe implement one or two commands for the same.
In the meanwhile, I would love some review so that I can understand if I am thinking in the right direction.
Please forgive me for my mistakes, I probably have made quite a few. 
Looking forward to your comments,

thanks & regards,
vani

[rendered](https://github.com/openebs/openebs/blob/e914a80ba3b9267b671b5b73d4a819f1ed5c0542/contribute/design/1.x/cli_tool/cli_proposal.md)

Signed-off-by: Vani Singh <vanisingh@live.co.uk>

### DIFF
--- a/contribute/design/1.x/cli_tool/cli_proposal.md
+++ b/contribute/design/1.x/cli_tool/cli_proposal.md
@@ -1,65 +1,92 @@
 ---
-title: Easy to use cli for OpenEBS
+title: Easy to use cli for OpenEBS + ( mini test suite + doccumentation)
 authors:
   - "@vaniisgh"
 creation-date: 2020-05-13
-last-updated: 2020-05-13
+last-updated: 2020-05-14
 status: proposed
 ---
 
 ## Easy to use command-line interface (CLI) for OpenEBS
 
 ## Table of Contents
-    * [Table of Contents](#table-of-contents)
-    * [Summary](#summary)
-    * [Proposal](#proposal)
-    * [Proposed Usage Options](#proposed-usage-options)
-    * [Proposed Implementation](#proposed-implementation)
 
+  * [Table of Contents](#table-of-contents)
+  * [Summary](#summary)
+  * [Proposal](#proposal)
+  * [Proposed Usage Options](#proposed-usage-options)
+  * [Proposed Implementation](#proposed-implementation)
 
-### Summary 
+### Summary
 
-OpenEBS is completely Kubernetes native and is implemented using microservices. 
-Currently OpenEBS can be installed via kubectl or helm chart. And managed via 
+OpenEBS is completely Kubernetes native and is implemented using microservices.
+Currently OpenEBS can be installed via kubectl or helm chart. And managed via
 a few custom resources.
-Often these require the user to have detailed knowledge of the underlying 
-infrastructure of the deployment and use long shell commands to configure the 
-package. This intern takes away from the ease of use of the package and may 
-discourage a certain group of users from exploring the features of available.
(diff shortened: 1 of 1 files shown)


harshthakur9030:
Hey @vaniisgh  
* In the proposed usage section, you have mentioned `kubectl openebs`. I think the project requires it's own binary since kubectl maintainers won't be adding support for external projects. 
* In the implementation proposed, you've used spf13/cobra. Can you take a look at [this](https://github.com/urfave/cli/blob/master/docs/v2/manual.md)? It provides features to load json/yaml files and some other features out of the box that I believe will be useful. What do you think?

vaniisgh:
Hey @harshthakur9030 
- the project should, of course, have its own binary, it is intended to be used as a plugin, I thought it could be installed using a simple curl command then the can be added to the user's path variable,  hence shouldn't need added support from kubectl. ( It could also be installed as a helm package ?)
- Thanks for the suggestion I will check CLI v2 manual out :) actually, I have used cobra before and simply thought i'd use it because I have seen the mayactl, as well as a few other CLI tools, use it and thought it would fulfill all the requirements :) 

I am pretty new to all this so do correct me if I am wrong :)

harshthakur9030:
Thanks for replying. 
* Would it not be simpler to just use the binary name instead of prepending it with kubectl? As for helm, it is to be used to install components inside a cluster. I guess the idea is to install openebs components with just one binary and assume users might not have helm.
* Sure. 
I am new to this too. :)

vaniisgh:
- it seems simpler only for some operations like say installing OpenEBS and onto a cluster and then the intended usage to install a default configuration onto an empty cluster could be :
``` openebs install | kubectl apply -f -```
but at the end of the day, it is a [kubectl plugin], it will leverage kubectl in some operations say to upgrade a volume, or to monitor a volume. 
It needs to be a one-stop solution to all the user's needs when it comes to OpenEBS.

- as for the helm chart, OpenEBS exists as helm chart already, So the CLI plugin will probably be shipped in it. 

Of-course I am still working on the proposal ... I will try to sort this out in my next commit.

this is just my understanding at the moment, I will need to read more and update the proposal to make things clearer :)

thanks again for taking the time to go through it :) 

[kubectl plugin]:https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/

harshthakur9030:
Thanks for sharing, I didn't know about the plugin ecosystem around kubectl. . Now that I've read a bit about it, plugin does seem the better option.

vaniisgh:
Thanks for your time & review :)
feedback is always invaluable :)

harshthakur9030:
After looking at a few projects like [openfaas](https://docs.openfaas.com/deployment/kubernetes/), the installation part is better handled by Helm or Arkade and if need be people can use the operator yaml. The idea of plugin architecture is nice and I found folks at knative have built their [client](https://github.com/knative/client) whose design we can follow. CLI created  can warn users when they're not on latest version and suggest to upgrade using package managers. What are your thoughts? @vaniisgh  @kmova

PR #3096, opened 2020-08-07
chore(yaml): update operator yaml

akhilerm:
- use apigroups for crd instead of individual resource names
- enable NDM feature flag by default
- add /dev mount to NDM volume mounts
- sync changes from operator yaml in openebs/charts

### DIFF
--- a/k8s/openebs-operator.yaml
+++ b/k8s/openebs-operator.yaml
@@ -203,7 +203,7 @@ spec:
         - name: OPENEBS_IO_CSTOR_POOL_EXPORTER_IMAGE
           value: "openebs/m-exporter:ci"
         - name: OPENEBS_IO_HELPER_IMAGE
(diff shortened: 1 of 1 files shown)


k8s/openebs-operator.yaml
```diff
@@ -743,7 +747,7 @@ spec:
         - name: OPENEBS_IO_INSTALLER_TYPE
           value: "openebs-operator"
         - name: OPENEBS_IO_HELPER_IMAGE
-          value: "quay.io/openebs/linux-utils:1.4.0"
+          value: "openebs/linux-utils:1.12.0"
```
kmova, line 750: Can you also update this with `ci` images, so we can test the changes wit multi-arch image. @akhilerm

akhilerm: Done.

PR #3116, opened 2020-08-31
feat(oep): zpool/zfs command execution without exec into the pool pod

mittachaitu:
This PR adds proposal to execute ZPOOL/ZFS commands without
exec into the cStor pool pod.

### DIFF
--- a/contribute/design/1.x/cstor/zcmd-operation.md
+++ b/contribute/design/1.x/cstor/zcmd-operation.md
@@ -34,7 +34,7 @@ via CSP custom resource.
 
 ## Motivation
 
-- Zpool and ZFS command execution via CSP custom resource
+- Zpool and ZFS command execution via CSP custom resource is one of the use case where migration job needs to execute zpool commands even if user has tighten the RBAC rules(Here example: Not allowing other process to exec into cStor pool pods).
 
 ### Goals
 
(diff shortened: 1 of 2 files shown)


contribute/design/1.x/cstor/zcmd-operation.md
```diff
@@ -0,0 +1,232 @@
...
+	Capacity CStorPoolCapacityAttr `json:"capacity"`
+
+	// LastTransitionTime refers to the time when the phase changes
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+
+	//ReadOnly if pool is readOnly or not
+	ReadOnly bool `json:"readOnly"`
+
+    // CommandResult contains the output of command requested in spec section
+    CommandResult string `json:"commandResult"`
```
shubham14bajpai, line 220: Few questions
 - what is the maximum limit for output?
 - if a command with trailing output is provided what will be the behaviour? Will the truncated output come here or it wait for the entire output?

mittachaitu: > what is the maximum limit for output?

As JSON string doesn't contain any limits AFAIK... but when unmarshaled into go struct it will depend on pod memory limits...

> if a command with trailing output is provided what will be the behaviour? Will the truncated output come here or it wait for the entire output?

Can you elaborate on what is meant by trailing output?

shubham14bajpai: > Can you elaborate on what is meant by trailing output?

If a command takes more time to  execute than the timeout of the sync then the output will come as truncated or it will not come?

mittachaitu: It will be an atomic operation upon failure/success only we will update the output

shubham14bajpai: Then it is good. What if we want to execute that takes more time than the sync period? Do we need to change the sync period as well?

mittachaitu: Actually there is maxRetryCount(new filed) controller will retry the command for specified no.of times and the sync period can have any value and high-level operators shouldn't worry about the sync period.

Q: I am thinking to add another field in status to represent whether the command succeeded/failed. WDYT?

contribute/design/1.x/cstor/zcmd-operation.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Goals
+
+- High level operators which requires execution of Zpool/ZFS command can be executed via CStorPool(CSP) custom resource.
+
+## Proposal
+
+### Stories
+
+High level operator should able to execute Zpool/ZFS commands even if user has tighten the RBAC rules.
```
sonasingh46, line 47: tighten RBAC rules ?? Which component is impacted and how due to this?

sonasingh46: Also, how a high-level operator executes zpool/zfs command ?? It is always the cstor-pool-mgmt. 
Are we planning to have any operator that will run as a sidecar like cstor-pool-mgmt?

contribute/design/1.x/cstor/zcmd-operation.md
```diff
@@ -0,0 +1,232 @@
...
+			- [WrokFlow](#wrokflow)
+
+## Summary
+
+This proposal brings out the design details to implement zpool/zfs commands execution
+via CSP custom resource.
+
+## Motivation
+
+- Zpool and ZFS command execution via CSP custom resource
```
sonasingh46, line 37: If we can have an elaboration of why this is required, What are the motivations? I can see that SPC --> CSPC migration kind of needs this. If we can put the exact story of that -- it would be a good example.

contribute/design/1.x/cstor/zcmd-operation.md
mittachaitu: > how a high-level operator executes zpool/zfs command ??

There is a section called #steps-to-be-performed-by-high-level-operator which will address How part.

contribute/design/1.x/cstor/zcmd-operation.md
shubham14bajpai: Yes makes sense, we cannot assume empty output as a failure.

contribute/design/1.x/cstor/zcmd-operation.md
mittachaitu: Done

mittachaitu:
I will rework and update this OEP with annotation approach as well as RBAC approach as discussed in standup call

PR #3124, opened 2020-09-09
fix(monitoring): Update openebs-monitoring-pg.yaml

Beheadedstraw:
- Fixes for new K8s API deprecation. 
- Weird white spaces removed.


k8s/openebs-monitoring-pg.yaml
```diff
@@ -311,8 +319,8 @@ metadata:
   name: openebs-prometheus-service
   namespace: openebs
 spec:
-  selector: 
-    name: openebs-prometheus-server
+  selector:
+    name: openebs-prometheus-service
```
kmova, line 323: @Beheadedstraw, this should be the label on the prometheus deployment.

PR #3212, opened 2020-10-04
Improve readme with extra helm command which created confusion

kashish2909:


prateekpandey14:
@kashish2909 openebs.io install section have the steps , this change is not required, its up to user to add the other available charts

PR #3262, opened 2020-10-18
README Improved.md

ritikofficial:


kmova:
This PR is not required. OpenEBS is right way to pronounce.

PR #3269, opened 2020-10-19
German Language Translation

JoelJJoseph:


kmova:
@JoelJJoseph - can you help with following changes in this PR:

PR #3288, opened 2020-10-30
Update README.md

vinayak017:
There were some selling mistakes in this file. So I update it.


kmova:
The changes are not required. Almost all the corrections are invalid.

Issue #2855, opened 2019-12-11
Add support to disable over provisioning on cStor Pools


kmova:
## Description

It was observed in some cases that even though the application only effectively writes about 5GB of data, the underlying pool shows more than 50GB used. Sometimes the pool can be 100% full after which the pool stops serving IO causing the applications to fail. 

## Context
Over provisioning is enabled by default on cStor volumes. Example, users will be able to create a volume of 200GB on cStor Pool with disk capacity of 100G. This allows for cases where user can save cost on the underlying storage spend and increase the capacity of the pools by adding more disks or expanding the disks, as volume starts consuming the space. 

The usage can be monitored by looking at the volume and pool metrics. 

While the above is applicable for the write-intensive workloads like Elastic, the workloads like postgresql that will potentially modify the existing blocks can end up quickly consuming more space from the pool. 

In case of postgres, that typically installs Ext4 on top of the cstor volumes will result in the following:
- Ext4 is mounted without unmap support
- As data is written and modified, Ext4 tends to use new blocks of storage - without sending any delete (unmap) requests back to the underlying block storage - like cStor. 
- As the volumes are over provisioned, cStor keeps allocating new blocks from the pool. 

This will result in a case where the application would have effectively used only 3GB of data on ext4, but about 100GB of data is already utilized in the pool because volume (ext4) quota is set at 200GB. 

## Workaround
Enable storage class based resource quota to set a limit on the sum of capacity allocated to all the PVCs to be within the available capacity of the underlying cStor Pools. https://kubernetes.io/docs/concepts/policy/resource-quotas/#enabling-resource-quota

## Possible solution
* Provide an option at the cStor Storage Class to disable over provisioning

vishnuitta:
disabling thin provisioning may sound like solves this issue, but, user faces errors during provisioning time.. it should be part of best practices for user to look at pool usage and take necessary actions.
unmap is worth to have but this will be useful in the cases when user anticipated volume size wrongly with workloads that does writes and deletes (like above)

vishnuitta:
When pool becomes full, cstor-pool container crashes continuously. This ends up in OS disk space consumption to store cores.
This PR https://github.com/openebs/cstor/pull/282 disables dumping core by default.

vishnuitta:
request made in this issue https://github.com/openebs/openebs/issues/2673 is to disable thin provisioning, which is same as this issue

sonasingh46:
[...]

PR #3397, opened 2021-05-25
oep(cstor_replica_placement): oep for cstor replica placement

sonasingh46:
Addresses the following issue:
https://github.com/openebs/openebs/issues/2855
Signed-off-by: Ashutosh Kumar <sonasingh46@gmail.com>

### DIFF
--- a/contribute/design/3.x/cstor_replica_placement/placement.md
+++ b/contribute/design/3.x/cstor_replica_placement/placement.md
@@ -1,0 +1,111 @@
+---
+oep-number: CStor Volume Replica Placement
+title: Placement of CStor Volumes to a CStor Pool by enforcing constraints and policies
+authors:
+  - "@sonasingh46"
+owners:
+  - "@kmova"
+  - "@sonasingh46"
+  - "@prateekpandey14"
+creation-date: 2021-05-25
+last-updated: 2021-05-25
+status: provisional
+---
+
+# CStor Pool Migration
+
+## Table of Contents
+
+- [CStor Volume Replica Placement](#cstor-volume-replica-placement)
+  - [Table of Contents](#table-of-contents)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
+      - [Support to disable overprovisioning on cstor pool](#support-to-disable-overprovisioning-on-cstor-pool)
+    - [Proposed Implementation](#proposed-implementation)
+      - [cvr-scheduler](#cvr-scheduler)
+    - [Steps to perform user stories](#steps-to-perform-user-stories)
+    - [Schema changes](#schema-changes)
+
+## Summary
+
+This proposal brings out the design details to implement volume placement on
+CStor pool.
+
+## Motivation
+
+In current impementation, the CVR (CStor Volume Replica) gets bound to a CSPI
+(CStor Pool Instance) by the cvr controller in a random fashion. As a result of
+this, users have no way to specify to bind a CVR to a preferred CSPI. 
+
+- CStor volume placement should be supported for the following cases
+(not exhaustive):
+  1. Disable overprovisoning on CSPI
+  2. Migration of a CVR from one CSPI to another
+  3. Selecting the best CSPI(s) for the CVR(s) based on node and zone locality
+
+
+### Goals
+
+- The bind of a CVR to a CSPI should be deterministic and users should be able to
+specify policies for the binding. As a first cut implementation, the policy of 
+interest here is the overprovisioning restriction. 
+
+### Non-Goals
+
+- Migrating a CVR from one CSPI to another
+- Policies around node and zone locality
+
+## Proposal
+
+### User Stories
+
+#### Disable overprovisioning on CStor pools
+As an OpenEBS user, I should be able to disable overprovisioning on CStor pools.
+
+#### Migrating a CVR from one CSPI to another
+As an OpenEBS user, I should be able to migrate volume from one CStor pool to
+another CStor pool.
+
+#### Enforce policies on CVR binding
+As an OpenEBS user, I should be able to enforce policies e.g. node,zone locality
+to bind to a most suitable(best effort) CSPI.
+
+### Proposed Implementation
+
+#### CVR-Scheduler
+
+In current impementation, the CVR (CStor Volume Replica) gets bound to a CSPI
+(CStor Pool Instance) by the cvr controller in a random fashion. See this [link]
+(https://github.com/openebs/cstor-operators/blob/master/pkg/controllers/cstorvolumeconfig/volume_operations.go#L396)
+
+A separate controller i.e cvr-scheduler watches for CVR(s). Once the CVR is 
+created, it is unassigned, meaning it does not belong to any CSPI. Now, this 
+cvr-scheduler enforces policies and requirements e.g. overprovisioning etc and 
+picks up appropriate or most suitable CSPI for the particular CVR and assigns 
+this CVR to a CSPI by patching the CVR. This patching will basically put the 
+selected CSPI name and UUID on the CVR.
+
+The replica controller that also watches for the CVRs. It ignores the CVR that 
+has not been assigned any CSPI. Once the CVR is assigned to a CSPI, the replica 
+controller processes the CVR to carry out the volume provisioning.
+
+In case a quick binding is required, there will be an option to by pass the cvr
+scheduler and the usual way of randomised binding will prevail via the replica
+controller.
+
+The policies for the cvr-scheduler (e.g. overprovisioning, binding) can be passed
+to the CVR(s) via the CSPC.
+So, the CSPC is the user-interacting API where the policies can be mentioned for
+CVR bindings.
+
+### Steps to perform user stories
+
+Add annotations on CSPC to apply policies for CVR binding. 
+
+### Schema changes
+
+No schema changes are required. Policies can be passed via the annotations.


contribute/design/3.x/cstor_replica_placement/placement.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### CVR-Scheduler
+
+In current impementation, the CVR (CStor Volume Replica) gets bound to a CSPI
+(CStor Pool Instance) by the cvr controller in a random fashion. See this [link]
+(https://github.com/openebs/cstor-operators/blob/master/pkg/controllers/cstorvolumeconfig/volume_operations.go#L396)
+
+A separate controller i.e cvr-scheduler watches for CVR(s). Once the CVR is 
+created, it is unassigned, meaning it does not belong to any CSPI. Now, this 
+cvr-scheduler enforces policies and requirements e.g. overprovisioning etc and
```
mittachaitu, line 87: How scheduler reads the policies? Is it via some new CRs? (Or) Does CVR object itself holds the name of the policies?

contribute/design/3.x/cstor_replica_placement/placement.md
```diff
@@ -0,0 +1,111 @@
...
+A separate controller i.e cvr-scheduler watches for CVR(s). Once the CVR is 
+created, it is unassigned, meaning it does not belong to any CSPI. Now, this 
+cvr-scheduler enforces policies and requirements e.g. overprovisioning etc and 
+picks up appropriate or most suitable CSPI for the particular CVR and assigns 
+this CVR to a CSPI by patching the CVR. This patching will basically put the 
+selected CSPI name and UUID on the CVR.
+
+The replica controller that also watches for the CVRs. It ignores the CVR that 
+has not been assigned any CSPI. Once the CVR is assigned to a CSPI, the replica 
+controller processes the CVR to carry out the volume provisioning.
```
mittachaitu, line 94: nit: Can we rephrase in a more clear way? Some more context like where this replica controller running? How it identifies CVR to reconcile? If we patch CVR with CSPI name & CSPI UID, Which replica controller will reconcile CVR?

contribute/design/3.x/cstor_replica_placement/placement.md
```diff
@@ -0,0 +1,111 @@
...
+
+In case a quick binding is required, there will be an option to by pass the cvr
+scheduler and the usual way of randomised binding will prevail via the replica
+controller.
```
mittachaitu, line 98: How users can pass the policy options? Is it via StorageClass parameters?

mittachaitu: Ok, Seen in the below lines.
> The policies for the cvr-scheduler (e.g. overprovisioning, binding) can be passedto the CVR(s) via the CSPC.

[Story-1:]
Are all the policies are specified via CSPC? If yes, then I have one followup question, I am a user of OpenEBS cStor and I have >=6 cStor pools(On different nodes) and I want to place my important workload in pools where volume replica count is less (Or) placing replicas in pool based on available capacity. IMO this kind of policies should be passed via storageclass since it is related to volume replica placement.


[Story-2]:
If I am not wrong as of now we are interested only in [ThickProvisioning](https://github.com/openebs/api/blob/master/pkg/apis/cstor/v1/cstorpoolcluster.go#L111-L113)(OverProvisioing). IMO that it is not a policy but it is a kind of parameter that specifies the property/nature of the pool. If it is specified then we will not allow placement of replicas in the pool if available capacity is less than the current volume replica capacity, right?

mittachaitu (commented):
Provided few questions to get more clarification

contribute/design/3.x/cstor_replica_placement/placement.md
sonasingh46: Sure. Makes sense. Will rephrase and put

contribute/design/3.x/cstor_replica_placement/placement.md
sonasingh46: With Re to [Story-1]: 
Passing policies via storage class makes sense to me. I think we should have two ways to specify policies : 
1. via CSPC ( To enforce pool level constraint and policies e.g. `overprovisioning`
2. via StorageClass ( To select most unused CStorPool )
Will make the changes and commit.

sonasingh46: With Re to [Store-2] : 
My thoughts are the following: 
We should not pull in the data layer concepts to the control plane. The Control plane just knows how to orchestrate, that's all. 
In that sense, `overprovisioning` is a mere policy, although it is a low-level property at CStor engine too. 
Plus, there is no need to get/fetch the pool's available capacity. We just need to know the total usable capacity of the pool and how many total replicas are present on the pool that request some amount of capacity. Then we calculate the remaining capacity that is guaranteed to be available for further placement. 
The catch is whenever a volume replica is placed on a pool with overprovisioning disabled, the data layer property is also set so that the replicas do not use more than requested. 

Note: Fecthcing available capacity for this kind of requirements is erroneous in my opinion due to the fact the I/O can happen in parallel. 
Correct me if I am missing something. Open to discussion.

mittachaitu: > there is no need to get/fetch the pool's available capacity. We just need to know the total usable capacity of the pool and how many total replicas are present on the pool that request some amount of capacity.

Yeah, make sense. I was under a different mindset thinking that we are populating correct usable capacity on CSPI status but it is not.

> Note: Fecthcing available capacity for this kind of requirements is erroneous in my opinion due to the fact the I/O can happen in parallel.

It is not recommended to query very frequently about zfs properties since it will degrade the performance to some more extent.

contribute/design/3.x/cstor_replica_placement/placement.md
```diff
@@ -0,0 +1,111 @@
...
+#### Disable overprovisioning on CStor pools
+As an OpenEBS user, I should be able to disable overprovisioning on CStor pools.
+
+#### Migrating a CVR from one CSPI to another
+As an OpenEBS user, I should be able to migrate volume from one CStor pool to
+another CStor pool.
+
+#### Enforce policies on CVR binding
+As an OpenEBS user, I should be able to enforce policies e.g. node,zone locality
+to bind to a most suitable(best effort) CSPI.
```
akhilerm, line 75: These 2 items are mentioned in the Non-goals section.

contribute/design/3.x/cstor_replica_placement/placement.md
akhilerm: We already have a CstorVolumePolicy CR right? Will it be used to provide the policies?

akhilerm: nvm the above comment. Read through the explanation below.

contribute/design/3.x/cstor_replica_placement/placement.md
```diff
@@ -0,0 +1,111 @@
...
+has not been assigned any CSPI. Once the CVR is assigned to a CSPI, the replica 
+controller processes the CVR to carry out the volume provisioning.
+
+In case a quick binding is required, there will be an option to by pass the cvr
+scheduler and the usual way of randomised binding will prevail via the replica
+controller.
+
+The policies for the cvr-scheduler (e.g. overprovisioning, binding) can be passed
+to the CVR(s) via the CSPC.
+So, the CSPC is the user-interacting API where the policies can be mentioned for
```
akhilerm, line 102: :+1:

akhilerm (commented):
just one question on how the policies will be passed to the CVC controller . Saw that only CSPC will be used to provide policies as this is the only user facing API.

sonasingh46:
@akhilerm 
>just one question on how the policies will be passed to the CVC controller . Saw that only CSPC will be used to provide policies as this is the only user facing API.

This is a good question and it should be present in the doc. I will put it in the design doc too. 
The CVC controller only needs to know if it can do a quick bind and go ahead. 
The polices will be attached to following two resources: 
- CSPC
- SC
And from there it will be picked up.

contribute/design/3.x/cstor_replica_placement/placement.md
sonasingh46: Yes. This first goal is to get the overprovisioning policy implemented via the framework and then keep incrementally adding new policies and constraints. 
Open to suggestions though.

mittachaitu:
High-level points that were made during the OpenEBS sync up meeting:

## Current Flow that should be handled if there is a CVR scheduler:

**PVC creations with the storageclass that cStor policy**:
- When a user creates PVC cStor-CSI driver will [create a CVC](https://github.com/openebs/cstor-csi/blob/73dc59057ce481fc9e27a272eee0ee304e11ad3e/pkg/utils/maya.go#L117-L129) with a user-specified [cStor policy](https://github.com/openebs/cstor-csi/blob/73dc59057ce481fc9e27a272eee0ee304e11ad3e/pkg/driver/controller.go#L104) on [StorageClass](https://github.com/openebs/cstor-operators/blob/master/docs/tutorial/volumes/policy.md#cstor-volume-policies). If policy is not specified then the normal provision will happen(CSI-Driver will specify the policy name in CVC [annotations](https://github.com/openebs/cstor-csi/blob/73dc59057ce481fc9e27a272eee0ee304e11ad3e/pkg/utils/maya.go#L85-L89)).
- Once the CVC is created then CVC controller will get an event and process the CVC by [reading policies](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/pkg/controllers/cstorvolumeconfig/controller.go#L292-L296) and create replicas based on various policies:
    -   [Rplica information specified in policies](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/pkg/controllers/cstorvolumeconfig/volume_operations.go#L375-L383) this is a case where users are migrating stack from SPC to CSPC based schema.
    - Users specified [replica-afinity](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/docs/tutorial/volumes/policy.md#replica-affinity) feature via policies then CVC-Controller will schedule [replicas on the node where the pod](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/pkg/controllers/cstorvolumeconfig/volume_operations.go#L398-L405) is scheduled and this is helpful to distribute STS application volumereplicas in a different CSPI.
    - During the creation of clone volumes from snapshot CVRs should be placed in pools where snapshots exist else replicas will never turn to a healthy state. Currently, CVC controller [placing](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/pkg/controllers/cstorvolumeconfig/volume_operations.go#L385-L390) clone volume replicas in pools where snapshots exist.


**Restore case**:
There are two types of backup & restore supported by cStor:
- [Local backup & restore](https://github.com/openebs/velero-plugin#local-backuprestore): Usecase where users need to create a clone in a different namespace.
- [Remore backup & restore](https://github.com/openebs/velero-plugin#remote-backuprestore): Usecase where data is backed up in some remote location.

Remote restore case:
- In a remote restore case [restore end-point](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/pkg/server/cstorvolumeconfig/restore_endpoint.go#L55) will create CStorRestore resources(which are responsible for pulling data from a remote location) in a [pool](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/pkg/server/cstorvolumeconfig/restore_endpoint.go#L238-L239) where CVRs exist. In case of delayed binding values may be empty and this will case restore to fail.
**Note**: CStorRestore controller will be running in cStor pool pod and it works exactly in a way CVR works i.e binding to the pool will happen via labels on CStorRestore resource.

Local Restore case:
- In the local restore case restore-endpoint will create a CVC that will look like a [clone CVC](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/pkg/server/cstorvolumeconfig/restore_endpoint.go#L385)(i.e CVC for a clone PV). Once the CVC is created it will be picked by the CVC controller and creates replicas in a place where snapshots exist.

#### Day-2-Operations

**Scaleup operations**:
- Currently, the user can specify a new pool name(pool name where volume replica of that volume doesn't exist) in CVC spec and the CVC controller will trigger [scaling operation](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/pkg/controllers/cstorvolumeconfig/controller.go#L824) and it intern create [new CVR](https://github.com/openebs/cstor-operators/blob/b6a52f8f1df60da7308997da8d2d798f6d0a40dd/pkg/controllers/cstorvolumeconfig/volume_operations.go#L819)(immediate binding) in the new pool.



AFAIK above all needs to take care by informing to  **CVR scheduler** saying skip scheduling for CVRs that are created in any of the above cases.

@prateekpandey14 @shubham14bajpai please add if I missed anything
cc: @sonasingh46

Issue #3492, opened 2021-12-20
[Security] Workflow stale.yml is using vulnerable action actions/stale


akulpillai:
The workflow stale.yml is referencing action actions/stale using references v1. However this reference is missing the commit [af4072615903a8b031f986d25b1ae3bf45ec44d4](https://github.com/actions/stale/commits/af4072615903a8b031f986d25b1ae3bf45ec44d4) which may contain fix to the some vulnerability. 
The vulnerability fix that is missing by actions version could be related to: 
(1) CVE fix 
(2) upgrade of vulnerable dependency 
(3) fix to secret leak and others. 
Please consider to update the reference to the action.

kmova:
Hi @akulpillai -- would you like to go ahead and contribute the fix?

vharsh:
@akulpillai Thanks for creating an issue regarding the same, the OpenEBS community takes security very seriously, do let us know how you keep yourself abreast with all things security(viz some scanners, or just a loose set of channels like mailing-lists, Reddit, HackerNews, etc)


👋🏼 Aside, we hang out in [#openebs-dev](https://kubernetes.slack.com/archives/CUAKPFU78/) in the K8s Slack, where we get deep and discuss, do drop in to say hi.

PR #3506, opened 2022-01-07
Bump up actions/stale

vharsh:
fixes: #3492

~TODO: Check if any breaking changes are there b/w v1 & v4~

# List of breaking changes

1. Doesn't impact us.
2. [new-feature] We can limit or modify the `operations-per-run` to limit the API-usage or increase the limit to ensure that everything is processed in one single run. I don't think this should impact us, but we'd know if it exceeds the default number via the Action runs log.

Ref: https://github.com/actions/stale/releases/tag/v4.0.0

Signed-off-by: Harsh Vardhan <harsh.vardhan@mayadata.io>

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/stale@v1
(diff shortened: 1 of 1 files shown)


shubham14bajpai (approved):
LGTM

PR #3699, opened 2024-02-16
Update README.md

sakirma:
It is just a type on the readme that I have noticed.


orville-wright (approved):
Thanks for catching this.

PR #3715, opened 2024-04-19
feat(charts): add chart for v4.0.1

niladrih:


charts/README.md
```diff
@@ -92,10 +92,10 @@ helm delete `<RELEASE NAME>` -n `<RELEASE NAMESPACE>`
 
 | Repository | Name | Version |
 |------------|------|---------|
-|  | openebs-crds | 4.0.0 |
```
avishnu, line 95: why does crds chart version need to change?

niladrih: We're mapping the chart version to the crds chart.

avishnu: and why do we do that?

avishnu: Hmm, not sure. Without a change to the crds chart, is it right to increment its version?

niladrih: If we add a future template/functionality which makes use of the chart's version then that template would have to work with version ranges which differ from the chart's version. Which is not too hard to do I suppose. Do you think I should revert to 4.0.0 for the crds chart?

tiagolobocastro: IIRC we also made the crd chart follow the mayastor chart version.
But I agree that we may not need to always bump the crd if there are no changes..

tiagolobocastro (approved):

charts/README.md
niladrih: Reverted openebs-crds to 4.0.0.

avishnu (approved):

Issue #3741, opened 2024-06-13
[ DOCS ] Installing Replicated Engine with non-default cluster domain


tiagolobocastro:
When using the non-default cluster.local domain, etcd needs to be configured as such.
Example:
helm install ..... --set mayastor.etcd.clusterDomain=xxxx

## Context

Otherwise etcd will not work properly, example: https://github.com/openebs/mayastor/issues/1421#issuecomment-1980269618

orville-wright:
@balaharish7 
I think @tiagolobocastro is asking... 
- can you please add this to the ``` Replicated PV Mayastor ``` Install DOCs . 

Tag @avishnu for awareness

balaharish7:
This issue is documented in the PR [#462](https://github.com/openebs/website/pull/462).

avishnu:
Closing this as per the above comment.

Issue #462, opened 2017-10-03
Create v0.4.0 Release Notes


mahebbar:
New Release Notes for v0.4

PR #3747, opened 2024-06-18
chore(charts): expose 'etcd.clusterDomain' from the mayastor chart

niladrih:
This exposes the etcd.clusterDomain value from the mayastor chart. This is commented by default, so that its presence is conveyed while still using the mayastor chart's value by default.

Ref: #3741

@balaharish7 -- we could refer to this value in our docs.

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -108,10 +108,9 @@ helm delete `<RELEASE NAME>` -n `<RELEASE NAMESPACE>`
 | localpv-provisioner.rbac.create | bool | `true` |  |
 | lvm-localpv.crds.csi.volumeSnapshots.enabled | bool | `false` |  |
 | lvm-localpv.crds.lvmLocalPv.enabled | bool | `true` |  |
-| lvm-localpv.enabled | bool | `true` |  |
 | mayastor.crds.csi.volumeSnapshots.enabled | bool | `false` |  |
 | mayastor.csi.node.initContainers.enabled | bool | `true` |  |
-| mayastor.enabled | bool | `true` |  |
+| mayastor.etcd.clusterDomain | string | `"cluster.local"` | Kubernetes Cluster Domain |
 | mayastor.localpv-provisioner.enabled | bool | `false` |  |
 | openebs-crds.csi.volumeSnapshots.enabled | bool | `true` |  |
 | openebs-crds.csi.volumeSnapshots.keep | bool | `true` |  |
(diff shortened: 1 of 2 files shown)


charts/README.md
```diff
@@ -123,4 +122,3 @@ helm delete `<RELEASE NAME>` -n `<RELEASE NAMESPACE>`
 | release.version | string | `"4.0.1"` |  |
 | zfs-localpv.crds.csi.volumeSnapshots.enabled | bool | `false` |  |
 | zfs-localpv.crds.zfsLocalPv.enabled | bool | `true` |  |
-| zfs-localpv.enabled | bool | `true` |  |
```
avishnu, line 126: why are these 3 removed? Are they redundant?

charts/values.yaml
```diff
@@ -39,7 +39,9 @@ mayastor:
     node:
       initContainers:
         enabled: true
-
+  # etcd:
+  #   # -- Kubernetes Cluster Domain
+  #   clusterDomain: cluster.local
```
avishnu, line 44: Why not uncomment?

charts/README.md
niladrih: These options are no longer valid. They were removed during v4.0.0 release. However, this README.md still documents them.

Not related to the change of this PR.

charts/values.yaml
niladrih: In case mayastor chart moves to a different value, etc. There is no CI/automation to make the corresponding change on this chart. The behaviour we'd want out of this is: convey that this is configurable, but always fall back to the defaults issued by the mayastor chart.

Keeping it commented give us this flexibility. In case this value is updated on the mayastor chart, and this one is missed, then this chart would force the use of the outdated value and lead to undesirable behaviour.

Abhinandan-Purkait (approved):

balaharish7:
This issue is documented in the PR [#462](https://github.com/openebs/website/pull/462).

charts/values.yaml
tiagolobocastro: Do we need to add it here commented out at all?

niladrih: We haven't mentioned mayastor and the dependent charts clearly in the openebs.io/docs documentation. This yaml and this README.md may be only way for a naive user to find out about this.

tiagolobocastro: Adding to the README.md ok for the major settings, adding commented out to the yaml not sure.

niladrih: Sure I'll uncomment the configuration.

niladrih:
These changes are included in #3754.

PR #3772, opened 2024-08-27
fix: change helm registry for publishing charts

r1jt:
Change helm registry for publishing charts

### DIFF
--- a/.github/workflows/helm-testing-chart.yml
+++ b/.github/workflows/helm-testing-chart.yml
@@ -19,5 +19,5 @@ jobs:
         with:
           token: ${{ secrets.ORG_CI_GITHUB }}
           charts_dir: .
-          repository: openebs-chart-donotuse
+          repository: mayastor-chart-donotuse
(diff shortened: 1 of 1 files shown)


blaisedias (approved):

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
@tiagolobocastro Should we rename the mayastor-do-not-use repo to openebs... ? But I guess there will be impact on the test side of things?

tiagolobocastro (approved):
> @tiagolobocastro Should we rename the mayastor-do-not-use repo to openebs... ? But I guess there will be impact on the test side of things?

Yes I think we should do that instead...

r1jt:
> > @tiagolobocastro Should we rename the mayastor-do-not-use repo to openebs... ? But I guess there will be impact on the test side of things?
> 
> Yes I think we should do that instead...
Sure. I will create a task for it later as it needs some changes on the E2E side as well

PR #3780, opened 2024-09-17
chore: update umbrella dependent versions

Abhinandan-Purkait:
### Changes in the PR

- LocalPV Hostpath [4.1.0] --> [4.1.1] 
- LocalPV ZFS [2.6.0] --> [2.6.2] 
- LocalPV LVM [1.6.0] --> [1.6.2]
- Mayastor [2.7.0] --> [2.7.1] 
- Remove stale values.

### DIFF
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -25,7 +25,7 @@ dependencies:
     repository: "https://openebs.github.io/zfs-localpv"
     condition: engines.local.zfs.enabled
   - name: lvm-localpv
-    version: 1.6.1
+    version: 1.6.2
     repository: "https://openebs.github.io/lvm-localpv"
     condition: engines.local.lvm.enabled
   - name: mayastor
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
-version: 4.1.0
+version: 4.1.1
 name: openebs
-appVersion: 4.1.0
+appVersion: 4.1.1
 description: Containerized Attached Storage for Kubernetes
 icon: https://raw.githubusercontent.com/cncf/artwork/HEAD/projects/openebs/icon/color/openebs-icon-color.png
 home: https://www.openebs.io/
@@ -16,19 +16,19 @@ sources:
   - https://github.com/openebs/openebs
 dependencies:
   - name: openebs-crds
-    version: 4.1.0
+    version: 4.1.1
   - name: localpv-provisioner
-    version: 4.1.0
+    version: 4.1.1
     repository: "https://openebs.github.io/dynamic-localpv-provisioner"
   - name: zfs-localpv
-    version: 2.6.0
+    version: 2.6.1
     repository: "https://openebs.github.io/zfs-localpv"
     condition: engines.local.zfs.enabled
   - name: lvm-localpv
-    version: 1.6.0
+    version: 1.6.1
     repository: "https://openebs.github.io/lvm-localpv"
     condition: engines.local.lvm.enabled
   - name: mayastor
-    version: 2.7.0
+    version: 2.7.1
     repository: "https://openebs.github.io/mayastor-extensions"
     condition: engines.replicated.mayastor.enabled
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -21,7 +21,7 @@ dependencies:
     version: 4.1.1
     repository: "https://openebs.github.io/dynamic-localpv-provisioner"
   - name: zfs-localpv
-    version: 2.6.1
+    version: 2.6.2
     repository: "https://openebs.github.io/zfs-localpv"
     condition: engines.local.zfs.enabled
   - name: lvm-localpv
--- a/charts/charts/openebs-crds/Chart.yaml
+++ b/charts/charts/openebs-crds/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v2
 name: openebs-crds
-version: 4.1.0
+version: 4.1.1
 description: A Helm chart that collects CustomResourceDefinitions (CRDs) from OpenEBS.
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -9,7 +9,7 @@ localpv-provisioner:
   rbac:
     create: true
 
-# Refer to https://github.com/openebs/zfs-localpv/blob/v2.6.1/deploy/helm/charts/values.yaml for complete set of values.
+# Refer to https://github.com/openebs/zfs-localpv/blob/v2.6.2/deploy/helm/charts/values.yaml for complete set of values.
 zfs-localpv:
   crds:
     zfsLocalPv:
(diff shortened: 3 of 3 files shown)


charts/charts/openebs-crds/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v2
 name: openebs-crds
-version: 4.1.0
+version: 4.1.1
```
avishnu, line 3: Is this version bump necessary?

Abhinandan-Purkait: Not really, but iirc we decided that we would follow the chart version for this.

charts/Chart.yaml
```diff
@@ -16,19 +16,19 @@ sources:
...
+    version: 4.1.1
     repository: "https://openebs.github.io/dynamic-localpv-provisioner"
   - name: zfs-localpv
-    version: 2.6.0
+    version: 2.6.1
     repository: "https://openebs.github.io/zfs-localpv"
     condition: engines.local.zfs.enabled
   - name: lvm-localpv
-    version: 1.6.0
+    version: 1.6.1
```
niladrih, line 28:
```suggestion
    version: 1.6.2
```

charts/values.yaml
```diff
@@ -34,7 +32,6 @@ lvm-localpv:
 
 # Refer to https://github.com/openebs/mayastor-extensions/blob/v2.7.0/chart/values.yaml for complete set of values.
```
niladrih, line 33:
```suggestion
# Refer to https://github.com/openebs/mayastor-extensions/blob/v2.7.1/chart/values.yaml for complete set of values.
```

charts/values.yaml
```diff
@@ -14,7 +14,6 @@ localpv-provisioner:
 
 # Refer to https://github.com/openebs/zfs-localpv/blob/HEAD/deploy/helm/charts/values.yaml for complete set of values.
```
niladrih, line 15:
```suggestion
# Refer to https://github.com/openebs/zfs-localpv/blob/v2.6.1/deploy/helm/charts/values.yaml for complete set of values.
```

charts/values.yaml
```diff
@@ -24,7 +23,6 @@ zfs-localpv:
   
 # Refer to https://github.com/openebs/lvm-localpv/blob/HEAD/deploy/helm/charts/values.yaml for complete set of values.
```
niladrih, line 24:
```suggestion
# Refer to https://github.com/openebs/lvm-localpv/blob/lvm-localpv-1.6.2/deploy/helm/charts/values.yaml for complete set of values.
```

niladrih (commented):
[This](https://github.com/openebs/openebs/blob/main/charts/README.md#requirements) bit needs to be updated.
Also, we can drop the first two lines of the values.yaml
```yaml
release:
  version: "4.1.0"
```

charts/values.yaml
```diff
@@ -12,29 +9,26 @@ localpv-provisioner:
...
-  enabled: true
   crds:
     zfsLocalPv:
       enabled: true
     csi:
       volumeSnapshots:
         enabled: false
   
-# Refer to https://github.com/openebs/lvm-localpv/blob/HEAD/deploy/helm/charts/values.yaml for complete set of values.
+# Refer to https://github.com/openebs/lvm-localpv/blob/lvm-localpv-1.6.2/deploy/helm/charts/values.yaml for complete set of values.
```
kranurag7, line 21: hey there, I think this is not going to go well for the users of helm chart. The helm chart of version `v1.6.2` is using the env `LVM_NAMESPACE` via this commit https://github.com/openebs/lvm-localpv/commit/45e0fbdd6af38f3025132fff5b6d10c3c2eec1fb 

but the application is hardcoded to use `OPENEBS_NAMESPACE` and the application won't start because of this mismatch. (ref: https://github.com/openebs/lvm-localpv/blob/lvm-localpv-1.6.2/pkg/lvm/volume.go#L38)

I've created an issue https://github.com/openebs/lvm-localpv/issues/336 to seek clarity on the name we are going to use going forward and based on agreement we either need to change the helm chart of application. And aftewards, we either need to release the new version of application or helm chart and then this is good to be merged. Please let me know what you think?

Abhinandan-Purkait: Hi @kranurag7, v1.6.2 published chart has LVM_NAMESPACE in both chart and code. Things on develop appear a little bit different because of some CI work that is to be done. I believe @niladrih has already explained at channels. 
You can try installing v1.6.2 lvm standalone chart and see the behaviour.

Btw thanks for spending time on figuring this out and providing feedback on the same. The newer CI should be in place in couple of weeks hopefully. Hope this answers your question?

kranurag7: @Abhinandan-Purkait thanks for the update and acknowledgement. 

> Things on develop appear a little bit different because of some CI work that is to be done.

I think v1.6.2 doesn't have LVM_NAMESPACE from code side. Please check the following block: https://github.com/openebs/lvm-localpv/blob/lvm-localpv-1.6.2/pkg/lvm/volume.go#L74-L79 

It's coming from [1.6.2](https://github.com/openebs/lvm-localpv/releases/tag/lvm-localpv-1.6.2) tag which was released 4 days back and during the `init` call, we are checking for `OPENEBS_NAMESPACE` and I'm getting the same error message while testing _(not standalone but full openebs/openebs)_ controller as well. I still think that we need a fix for this and a new release. Please let me know if I'm missing anything. Happy to help with the fix.

kranurag7: Thanks again, I acknowledge that the 1.6.1 version of chart will work because revert was created in 1.6.2 version of the chart.

Side Note from packagers point of view: We rely heavily on tags to build the package and given the develop helm chart and application have a mismatch on the latest tag which 1.6.2 I ran into problems. 

Having said that, I understand that this issue is temporary and will be resolved in coming days/weeks.
I'll for now go ahead with 1.6.1 version of helm chart during testing and that should resolve it.

With that, I'll mark this thread as resolved.
Thanks again for the project and all the awesome work you all do.

charts/Chart.yaml
```diff
@@ -16,19 +16,19 @@ sources:
...
   - name: openebs-crds
-    version: 4.1.0
+    version: 4.1.1
   - name: localpv-provisioner
-    version: 4.1.0
+    version: 4.1.1
     repository: "https://openebs.github.io/dynamic-localpv-provisioner"
   - name: zfs-localpv
-    version: 2.6.0
+    version: 2.6.1
```
niladrih, line 24:
```suggestion
    version: 2.6.2
```

niladrih (changes requested):

charts/values.yaml
niladrih: You'd want to use v1.6.2 version of the helm chart. v1.6.1 has been retracted. Note that I am talking about the chart version in this case. The appVersion is v1.6.1, this is same for chart v1.6.1 and v1.6.2.

niladrih (approved):
LGTM

Issue #3792, opened 2024-10-14
Bind pv and pod to same node


fara-tode:
## Description
When setting pod over mayastor storage class with 1 replica, I want ensure that pod is scheduled and run on same node as the pv is.

in theory i should be able to achieve this by defining storagecass: volumeBindingMode: WaitForFirstConsumer which im using already.
But that seems to not really work for mayastor. I have multiple pod that are running on different nodes than their mayastor PVs.

There was parameter local in past which seems that did what is expected now, but the local parameter is deprecated now:

```
"local"
A flag of the type Boolean, with the default value of true. The flag controls the scheduling behaviour of nexus and replicas to storage nodes in the cluster.

All the following values are interpreted as "true" and anything else as "false": 'y', 'Y', 'yes', 'Yes', 'YES', 'true', 'True', 'TRUE', 'on', 'On', 'ON'.

This value must be set to true for correct operation of Mayastor provisioning and publishing. It is recommended that the volumeBindingMode in storage class be set to WaitForFirstConsumer. This limitation will be removed in a future release
A consequence of the above limitation is that applications pods which use Mayastor provisioned PVs may only be scheduled on nodes which are running Mayastor pods (i.e. data engine container). That is to say, only on MSNs. This limitaion will be removed in a future release.
```
I don't see that as limitation is that application pod will be scheduled only on the node that is hosting the actual volume.
I see it rather as expected as an optimal way running applications when you dont have dedicated storage k8s nodes, but want to just run an app on same node as data is.

## Context
In situation that with having 2 k8s nodes, and 2 pods that running replicated application. Both pods are sts but not in same stsgroup.
sts1-0 was running on node1 with mayastor pv on node2
sts2-0 was running on node2 with mayastor pv on node1
when any node die both 2 pods/sts are not working.

## Possible Solution
provide back the local parameter?

tiagolobocastro:
Hi @fara-tode, I think indeed pinning of the volume is a good idea for single replica volumes.

One thing to consider though, is that these single-replica volumes may be moved to another node, and in this case the pinning is not desirable as would prevent the application from starting on another node.

A custom scheduler perhaps is what's needed here, to ensure application is placed together with the storage? 

Probably a very bad idea, but just throwing it out there: adding a label to the nodes for every volume with a replica living there, and then we'd simply move the label to another node, thus achieving a flexible pinning. Obvious downside is lots of labels on the nodes...

[...]

PR #3801, opened 2024-11-07
docs(oep): proposal for local bind volumes (#3792)

todeb:


designs/helm-charts/local-bind-volumes.md
```diff
@@ -0,0 +1,119 @@
...
+## 3. Goals
+
+- Implement the local feature for single-replica volumes in Mayastor, ensuring that PVs are created on the same node as their consuming pods when `volumeBindingMode: WaitForFirstConsumer` is used.
+- Prevent pods from being rescheduled to nodes where the associated volume does not exist.
+- Allow users to set or unset the `local` flag for a volume through `kubectl-mayastor`, providing flexibility and administrative control.
+
+## 4. Non-Goals
+
+- This proposal does not implement the local feature for multi-replica volumes, as cross-node replication complicates strict local binding.
+- The local feature is specific to Mayastor and does not apply to other OpenEBS engines, such as Jiva or cStor.
```
tiagolobocastro, line 51: I think you can leave Jiva and cStor out, as they're no longer OpenEBS engines (they're archived)

designs/helm-charts/local-bind-volumes.md
```diff
@@ -0,0 +1,119 @@
...
+
+### OEP Number: <To be assigned>
+### Title: Reintroduction of Local Feature in Mayastor for Single-Replica Volumes
+### Authors: @todeb
+### Status: Proposed
+### Created: 2024-11-07
+### Last Updated: 2024-11-07
+### Target Release: ?
+
+---
```
tiagolobocastro, line 11: We can set as provisional since it's in PR.
It can also be merged as many times as you wish whilst provisional

```suggestion
---
oep-number: Issue also needs to be created
title: OpenEBS Enhancement Proposal for Mayastor Local Feature
authors:
  - "@todeb"
owners:
  - "@todeb"
  - "@tiagolobocastro"
editor: TBD
creation-date: 2024-11-07
last-updated: 2024-11-07
status: provisional
---

# OpenEBS Enhancement Proposal for Mayastor Local Feature
```

designs/helm-charts/local-bind-volumes.md
```diff
@@ -0,0 +1,119 @@
...
+
+## 2. Motivation
+
+In clusters that do not have dedicated storage nodes, keeping application pods on the same node as their Mayastor volumes is essential. While `volumeBindingMode: WaitForFirstConsumer` is intended to ensure that volumes are created on the same node as their consuming pods, it does not always work as expected with Mayastor volumes. Reintroducing the `local` feature with this proposal enforces both local provisioning and node-level affinity for single-replica volumes, thereby avoiding cross-node data access and potential latency or availability issues.
+
+## 3. Goals
+
+- Implement the local feature for single-replica volumes in Mayastor, ensuring that PVs are created on the same node as their consuming pods when `volumeBindingMode: WaitForFirstConsumer` is used.
+- Prevent pods from being rescheduled to nodes where the associated volume does not exist.
+- Allow users to set or unset the `local` flag for a volume through `kubectl-mayastor`, providing flexibility and administrative control.
```
tiagolobocastro, line 46: What if you set `local` on an existing volume, residing on node_x and existing pod residing on node_y?
What would be the expected behaviour? The pod consuming the volume would be moved only after existing pod is deleted? (assuming it's some kind of replica set and not a plain Pod without a controller)

designs/helm-charts/local-bind-volumes.md
```diff
@@ -0,0 +1,119 @@
...
+  - Add a `local` flag to the volume specification. When `local: true` is set, Mayastor enforces both local provisioning and anti-rescheduling of single-replica volumes.
+  - `kubectl-mayastor` commands can be used to set or unset the `local` flag for active volumes, dynamically controlling local binding.
+
+- **Single-Replica Volume Constraint**: 
+  - This feature applies only to volumes configured with one replica, as multi-replica configurations may require cross-node replication.
+  - If additional replicas are added, the `local` flag is automatically disabled, and cross-node rescheduling may occur.
+
+- **Affinity Management**:
+  - Node affinity rules are applied to the volume based on the `local` flag setting, ensuring that the pod and volume remain co-located.
+  - Kubernetes will use these affinity rules to restrict rescheduling to the initial node only.
```
tiagolobocastro, line 103: Do these rules help with `Immediate` binding mode? Also can they be modified after initial creation? (since local flag is intended to be modifiable via plugin)

tiagolobocastro (commented):
Great proposal @todeb !
@openebs/org-maintainers please take the time to review this.
Other folks please also chip in.
@todeb you may merge this PR as required as we iterate, we don't need to agree all details until the final decision to take this in.

tiagolobocastro:
@abhilashshetty04 shall we disable rust lint for PR's which don't modify rust code?

designs/helm-charts/local-bind-volumes.md
todeb: Do nothing, is up to user to rollout restart or delete pod and kubernetes scheduler will reshedule it then with affinity to local volume.

designs/helm-charts/local-bind-volumes.md
todeb: local flag is intended to be set in storageclass, and also be modified via plugin.
If volume is created with Immediate and local, then volume can be created on random node and set with proper affinity. K8S scheduler will then schedule pod on specific node.
```
Node Affinity:
  Required Terms:
    Term 0:        openebs.io/csi-node in [mayastor]
```
```
Node Affinity:
  Required Terms:
    Term 0:        openebs.io/nodename=node1
```

todeb (commented):
did some clarifications in commit, let me know if sth more is missing.

designs/helm-charts/local-bind-volumes.md
Abhinandan-Purkait: How do we set PV affinity to existing volumes? The affinity is immutable in the PV. Also for a local:true volume, even if we set local to false, once PV affinity is set the application cannot be rescheduled to a new node. So how does the unset flow work?

todeb: right that might be the issue, possible options I see:
- get rid of the unset, once volume is created cannot be changed.
- delete pv definition retaining data and recreate the new pv
- make a volume replica with openebs.io/csi-node in [mayastor] affinity, and delete original volume

abhilashshetty04:
> @abhilashshetty04 shall we disable rust lint for PR's which don't modify rust code?

I think we should.

designs/helm-charts/local-bind-volumes.md
tiagolobocastro: I see, it kinda reverses the initial step, so basically:
`WaitForFirsConsumer`: brings storage to node where pod is first scheduled on
`Immediate`: brings pod to the node where mayastor chooses to place the replica

todeb: yes, and what is the confusion or reverse?

designs/helm-charts/local-bind-volumes.md
todeb: unset is not mandatory to be implemented for this, but could be nice to have add on.
3rd point right, it doesn't make any sense as the pv definition has the nodeaffinity and not the mayastor itself, so making more replicas of volume won't change anything in that case.

designs/helm-charts/local-bind-volumes.md
tiagolobocastro: It's clear for me now, just wanted to highlight this difference, which I think is important.

Abhinandan-Purkait:
@todeb You would need to sign your commits for DCO. You can refer the below for steps.
https://github.com/openebs/openebs/pull/3801/checks?check_run_id=32713394687

designs/helm-charts/local-bind-volumes.md
tiagolobocastro: > 3rd point right, it doesn't make any sense 

ok...
Not sure where "making more replicas" is coming from?
My point here is that with a custom scheduler you don't need to use the CSI topology feature, at all. 
At the cost of having to maintain a custom scheduler, which we might want to avoid..

tiagolobocastro: oh I see, your 3rd point, right yes I think replicas won't help here.
Anyways my suggestion is to keep this simple, and remove the ability to modify the local for now, that perhaps can be tackled on separate oep?

tiagolobocastro (approved):
IMHO we should remove local modification to keep this simpler, or add it as a milestone before we make a case to bring this into the roadmap

tiagolobocastro:
I think we can merge first cut of this, @todeb would you be able to move this file from helm-charts\ into replicated-pv\mayastor ?

todeb:
> I think we can merge first cut of this, @todeb would you be able to move this file from helm-charts\ into replicated-pv\mayastor ?

moved, I'm not authorized, so you can merge this.
You’re not [authorized](https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches) to merge this pull request.

avishnu:
@todeb would you be interested in taking this to the next level? This means, requesting the maintainers for consideration of the proposed OEP. Also, if you have further thoughts on this, please feel free to modify the proposal.

fara-tode:
@avishnu I don't really know how to handle this, I'm not developer.
But I'm actually seeing another problem were bind is not used. 

When doing restore by velero, mayastor will create volume on "random" node, and velero will create pod handling restore process on "random" node too. Usually these "random" nodes are different, and this slows significally restore time for volume.

todeb:
@avishnu I'm not a developer, so I'm not entirely sure how to proceed with requesting consideration from the maintainers.

I’ve noticed another issue related to how bind is not being used. During a restore with Velero, Mayastor often creates the volume on a "random" node, while Velero schedules the restore pod on a different "random" node. When these nodes don’t match, restore performance is significantly degraded because IOPS are much lower when accessing the volume remotely.

PR #3822, opened 2025-01-03
refactor openebs plugin

abhilashshetty04:
This PR contains following enhancements:

1. All storage units will be represented in human readable format like 4.5GB, 2.3KB etc.
2. All localpv engines need 2 cliargs. i.e namespace and outout_format. Since these were available already in mayastor plugin's Cliarg we were using them for localpv modules also. However, Mayastor args has some of the args which may not be needed for localpv cmds. Hence, We have separated them now. All localpv modules has their own cliargs.


tiagolobocastro:
There's no content in the PR body and PR title is also a bit vague

plugin/src/cli_utils/localpv/lvm/volume/types.rs
```diff
@@ -213,3 +223,10 @@ impl TryFrom<(&LvmVolume, PersistentVolume)> for LvmVolumeObject {
         })
     }
 }
+
+/// Converts bytes into human-readable capacity unit.
+fn adjust_bytes(bytes: u128) -> String {
```
dsharma-dc, line 228: Please move this function to some common place instead of copying implementation in three places.

abhilashshetty04: Done

dsharma-dc (approved):

plugin/src/cli_utils/localpv/zfs/mod.rs
```diff
@@ -1,6 +1,6 @@
-use crate::cli_utils::CliArgs;
 use clap::Parser;
 use kube::Client;
+use plugin::resources::utils::OutputFormat;
```
Abhinandan-Purkait, line 3: Same here

plugin/src/cli_utils/localpv/zfs/node/types.rs
```diff
@@ -1,3 +1,6 @@
+use super::Error;
+use crate::cli_utils::localpv::adjust_bytes;
+use anyhow::anyhow;
```
Abhinandan-Purkait, line 3: Add a blank line below

plugin/src/cli_utils/localpv/zfs/volume/types.rs
```diff
@@ -1,4 +1,6 @@
 use super::Error;
+use crate::cli_utils::localpv::adjust_bytes;
```
Abhinandan-Purkait, line 2: blank line here

plugin/src/cli_utils/mod.rs
```diff
@@ -1,63 +1,37 @@
 use clap::Parser;
 use kubectl_plugin::resources;
 use plugin::ExecuteOperation;
-use std::ops::Deref;
 pub(crate) mod localpv;
+pub(crate) mod mayastor;
```
Abhinandan-Purkait, line 5: Same here

plugin/src/main.rs
```diff
@@ -1,9 +1,7 @@
 use clap::Parser;
 use kube::Client;
 use kubectl_plugin::resources;
-use plugin::ExecuteOperation;
-use std::{env, ops::Deref};
-
+use std::env;
```
Abhinandan-Purkait, line 4: Same here

plugin/src/main.rs
```diff
@@ -17,37 +15,40 @@ struct CliArgs {
...
-    fn deref(&self) -> &Self::Target {
-        &self.args
-    }
-}
-
 #[tokio::main]
 async fn main() {
     let cli_args = CliArgs::args().await;
-    let _tracer_flusher = cli_args.init_tracing();
+    //let _tracer_flusher = cli_args.init_tracing();
```
Abhinandan-Purkait, line 51: Why is this commented?

abhilashshetty04: I will remove it. We dont need tracing for plugin.

plugin/src/cli_utils/localpv/lvm/node/types.rs
```diff
@@ -165,7 +165,7 @@ impl From<Vec<LvmNode>> for VolumeGroupRecord {
         let mut volume_group: Vec<LvmVolumeGroup> = Vec::new();
         for node in nodes {
             for vg in &node.volume_groups {
-                volume_group.push(LvmVolumeGroup::from((vg, &node.name_any())));
+                volume_group.push(LvmVolumeGroup::from((vg, &node.name_unchecked())));
```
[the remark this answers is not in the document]
abhilashshetty04: We can use name_any() also. name_unchecked() panics if its not set. We are sure in all cases that name would exist for For our Cr and Persistent volume.

It was suggested by @niladrih. I feel its okay to use both.

plugin/src/main.rs
tiagolobocastro: We don't we need it?
What if we need to debug it?

abhilashshetty04: I tried,  println!() works. That should be enough right?

tiagolobocastro: humm, you can't compare tracing/opentelemetry to a println :)

Abhinandan-Purkait (approved):

plugin/src/main.rs
abhilashshetty04: Done, we init tracing for Mayastor commands.

PR #3840, opened 2025-01-28
[OEP 3843]: Mayastor: at-rest encryption of data on diskpools.

dsharma-dc:
Add OEP for mayastor at-rest encryption of data 

## Motivation and Context

See the OEP

## Outcome

Please review the proposal

### DIFF
--- a/designs/replicated-pv/mayastor/encryption-at-rest.md
+++ b/designs/replicated-pv/mayastor/encryption-at-rest.md
@@ -1,0 +1,146 @@
+---
+oep-number: OEP 3843
+title: OpenEBS Enhancement Proposal for Mayastor At-Rest Encryption of data
+authors:
+  - "@dsharma-dc"
+owners:
+  - "@dsharma-dc"
+  - "@Abhinandan-Purkait"
+editor: TBD
+creation-date: 2025-01-24
+last-updated: 2025-01-24
+status: provisional
+---
+
+# OpenEBS Enhancement Proposal for Mayastor At-Rest Encryption of data
+
+## Table of Contents
+
+1. [Overview](#overview)
+2. [Motivation](#motivation)
+3. [Goals](#goals)
+4. [Non-Goals](#non-goals)
+5. [Proposal](#proposal)
+6. [User Stories](#user-stories)
+7. [Implementation Details](#implementation-details)
+8. [Testing](#testing)
+
+---
+
+## Overview
+
+This proposal introduces the encryption of data at-rest in the Replicated PV Mayastor (hereafter referred as Mayastor) diskpools. This feature ensures that:
+
+1. All the data on a Maystor diskpool is stored encrypted if the encryption has been requested on the pool during pool creation.
+2. It is not possible to disable encryption on a pool that has been created with encryption enabled initially.
+3. The encryption can not be enabled on a pool that is already created un-encrypted.
+4. A volume that is created with a topology for encrypted pools gets all the replicas placed on encrypted pools.
+
+This feature ensures data protection and compliance for various use-cases of data management.
+
+## Motivation
+
+Implementing data encryption at rest ensures that sensitive information remains secure even if storage media is lost, stolen, or compromised. It protects against unauthorized access by ensuring data is only readable by authorized users or systems. Encryption also helps meet regulatory compliance requirements, demonstrating a commitment to privacy and data protection. Additionally, it mitigates the risk of data breaches, maintaining trust with stakeholders.
+
+## Goals
+
+- Implement the at-rest encryption of data in Mayastor, at a diskpool level so that all user data on the diskpool is stored encrypted.
+- Support FIPS compliant encryption methods.
+
+## Non-Goals
+
+- This proposal does not implement (or integrate with) any Key management service for the encryption keys.
+- This proposal does not implement in-flight data encryption.
+- This proposal does not implement support for disabling encryption on a pool that has been created with encryption.
+- This proposal does not implement support for enabling encryption on a pool that is already created as un-encrypted.
+- This proposal does not implement or strengthen any existing communication protocols between various services.
+
+## Proposal
+
+### Key Concepts
+
+1. **Diskpool Level Encryption**: Ability to encrypt data at a pool level without differentiating which replica the data belongs to. All replicas' data on such a pool will get encrypted.
+2. **Encryption Key Management**: This is a prerequisite.
+   - A key is provisioned by admin or a Key Management Service (KMS) with a supported cipher, for the use of storage system, and stored as a Kubernetes Secret.
+   - The admin can additionally safeguard the key via Resource-Encryption-at-Rest facility provided by Kubernetes.
+
+### Workflow
+
+1. **Encrypted Diskpool Creation**:
+   - The user/admin creates a diskpool yaml spec that contains the name of the Secret which holds the key.
+   - When the spec is applied, the diskpool operator picks up this request and dispatches the create pool request containing Secret name to the Mayastor agents that complete the diskpool provisioning.
+   - The spec for creating a pool with encryption will look like below:
+
+```
+apiVersion: "openebs.io/v1beta3"
+kind: DiskPool
+metadata:
+  name: <pool-name>
+  namespace: <namespace>
+spec:
+  node: <node-name>
+  disks: ["/dev/disk/by-id/<id>"]
+  topology:
+    labelled:
+      topology-key: topology-value
+  encryptionKeyConfig:
+    type: Secret
+    config:
+      name: <myKeySecretName>
+```
+2. **Diskpool CRD migration**:
+   - After an upgrade to Diskpool CRD version v1beta3 happens in the cluster, any existing CRs that are on version v1beta2 will have to be migrated to version v1beta3 by the implementation. This would also require that the new field `encryptionKeyConfig` be defined as optional in the CRD.
+
+3. **Diskpool import upon node or io-engine restart**:
+   - In the event of a node restart, io-engine restart or a node going offline and coming up again - the diskpool
+   is imported. The import needs to be done using the same DEK that was used during diskpool creation.
+
+4. **Volume Provisioning**:
+   - Pool topology rules ensure that for a volume requesting encryption, the replicas are only placed on the diskpools that have
+   encryption enabled on them. This will be handled by storageclass via the poolHasTopologyKey setting to let the volume replica placement happen on pools that are labelled with a specific key identifying encryption.
+   - The storageclass definition required for volumes to be encrypted will have an additional field named `secure`, which needs to be set to `true` if encryption is required.
+```
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: mayastor-3
+parameters:
+  secure: "true"
+  repl: "3"
+provisioner: io.openebs.csi-mayastor
+```
+
+## User Stories
+
+1. **Story 1**: As an organisation's security lead, I want all our data getting stored on the storage systems be encrypted with a key that our admins provide.
+2. **Story 2**: As a system administrator, I want minimal configuration control to easily let storage use the encryption keys for data at-rest encryption.
+
+## Implementation Details
+
+### Design
+
+- **Secret Name Parameter**: 
+   - Add a new field `encryptionKeyConfig` to the diskpool CR. This fields holds the name of Secret object that contains actual DEK.
+   - The control plane agent-core refers the metadata info from `encryptionKeyConfig` and then parses the actual Data Encryption Key(DEK) from the Secret, via Kubernetes client APIs.
+   - The data plane receives the request to create the pool using the provided DEK.
+   - Once the pool is created using a Secret, the key for that pool can't be transparently changed via a different Secret. Doing so will require a full rebuild of pool onto a different pool.
+
+- **Secret Name during Pool Import**: 
+  - Set the `encryptionKeyConfig` in the PoolSpec.
+  - When a pool import is required, use the `encryptionKeyConfig` from PoolSpec to fetch the DEK again.
+  - Dispatch the import operation to data plane.
+
+### Components to Update
+
+- **Diskpool Custom Resource Definition**: The Custom Resource need to identify the `encryptionKeyConfig` field.
+- **Control-Plane agent-core**: The agent needs to be modified to have an ability to read Kubernetes API objects.
+- **Data-Plane io-engine**: io-engine need to be able to create and place a crypto block device on top of base block device of the diskpool.
+
+## Testing
+
+ - Create a diskpool with a Secret of AES_CBC cipher and 128-bit key. The pool creation must succeed.
+ - Create a diskpool with a Secret of AES_CBC cipher and 256-bit key. The pool creation must succeed.
+ - Create a diskpool with a Secret of AES_XTS cipher and 128-bit keys. The pool creation must succeed.
+ - Upon a node restart, the import of the encrypted pool on that node must successfully complete.
+ - Provision a volume via encryption storage class. The volume replicas must get placed only on encrypted pools.
+ - Scale up an encrypted volume. The new replicas must get placed only on encrypted pools.
--- a/designs/oep-template.md
+++ b/designs/oep-template.md
@@ -71,6 +71,7 @@ A table of contents is helpful for quickly jumping to sections of a OEP and for 
 * [Drawbacks [optional]](#drawbacks-optional)
 * [Alternatives [optional]](#alternatives-optional)
 * [Infrastructure Needed [optional]](#infrastructure-needed-optional)
+* [Testing](#testing)
 
 [Tools for generating]: https://github.com/ekalinin/github-markdown-toc
 
@@ -157,3 +158,8 @@ Similar to the `Drawbacks` section the `Alternatives` section is used to highlig
 Use this section if you need things from the project/owner.
 Examples include a new subproject, repos requested, github details.
 Listing these here allows a owner to get the process for these resources started right away.
+
+## Testing
+
+Use this section to list down the methods of testing the proposed enhancement.
+This can include various high level scenarios and the behaviour to be tested.


designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+6. [User Stories](#user-stories)
+7. [Implementation Details](#implementation-details)
+
+---
+
+## 1. Overview
+
+This proposal introduces the encryption of data at-rest in the Mayastor diskpools. This feature ensures that:
+
+1. All the data on a Maystor diskpool is stored encrypted if the encryption has been requested on the pool during pool creation.
```
Abhinandan-Purkait, line 33: Maystor --> Replicated PV Mayastor

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+2. It is not possible to disable encryption a pool that has been created with encryption.
```
Abhinandan-Purkait, line 34: disable encryption a pool --> disable encryption on a pool

created with encryption --> provisioned with encryption initially?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+3. The encryption can not be enabled on a pool that is already created un-encrypted.
+4. It would be possible upgrading Mayastor to a release where pool encryption is supported, and the pools can migrate from un-encrypted to
+encrypted if required, non-disruptively.
```
Abhinandan-Purkait, line 37: Are we sure we want to tackle this now? Isn't this dependent on Pool Drain?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+## 3. Goals
+
+- Implement the at-rest encryption of data in Mayastor, at a diskpool level so that all user data on the diskpool is stored encrypted.
+- Support FIPS compliant encryption methods.
+
+## 4. Non-Goals
+
+- This proposal does not implement (or integrate with) any Key management service for the encryption keys.
+- This proposal does not implement the encryption support per-volume.
+- This proposal does not implement in-flight data encryption.
```
Abhinandan-Purkait, line 54: Do you think we should also mention not implementing additional security protocols for communication between various services?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+
+## 5. Proposal
+
+### Key Concepts
+
+1. **Diskpool Level Encryption**: Ability to encrypt data at a pool level without differentiating which replica the data belongs to. All replicas' data on such a pool will get encrypted
```
Abhinandan-Purkait, line 60: `.` missing at the end.

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+2. **Encryptio Key Management**: A key provisioned by admin or a Key Management Service(KMS) for a supported cipher, for the use of storage system, and stored as a Kubernetes Secret. The admin can additionally safeguarded the key via Resource-Encryption-at-Rest facility provided by Kubernetes.
+
+### Workflow
+
+1. **Encryption Key Provisioning**
+   - The user/admin creates a Kubernetes Secret containing a key with supported cipher and length.
+   - The user/admin optionally secures the created Secret via Resource-Encryption-at-Rest facility provided by Kubernetes.
```
Abhinandan-Purkait, line 67: Should we mention this in the workflow as this isn't actually a part of implementable workflow and rather a good to have thing?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+
+2. **Encrypted Diskpool Creation**:
+   - The user/admin creates a diskpool yaml spec that contains the name of the Secret which holds the key.
+   - When the spec is applied, the diskpool operator picks up this request and dispatches the create pool request containing Secret name
```
Abhinandan-Purkait, line 71: missing `.`

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+4. **Volume Provisioning**:
+   - Pool topology rules ensure that for a volume requesting encryption, the replicas are only placed on the diskpools that have
+   encryption enabled on them.
+
+## 6. User Stories
+
+1. **Story 1**: As an organisation's security lead, I want all our data getting stored on the storage systems be encrypted with a key that our admins provide.
+2. **Story 2**: As a system administrator, I want minimal configuration control to easily let storage use the encryption keys for data at-rest encryption.
+3. **Story 3**: As a system administrator, I want a non-disruptive way to migrate the required diskpools from un-encrypted to encrypted during
+a system upgrade.
```
Abhinandan-Purkait, line 87: Again, this probably is dependent on a different feature.

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+
+## 7. Implementation Details
+
+### Design
+
+- **Secret Name Parameter**: 
+   - Add a new field `SecretName` to the diskpool CR. This fields holds the name of Secret object that contains actual DEK.
+   - The control plane agent parse the actual Data Encryption Key(DEK) using the `SecretName`.
```
Abhinandan-Purkait, line 95: I think doing it via the io-engine might be more safer as there would be less hops to sniff? Although is the k8s-api call secure in some way?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+  - When a pool import is required, use the `SecretName` from PoolSpec to fetch the DEK again.
+  - Dispatch the import operation to data plane.
+
+### Components to Update
+
+- **Diskpool Custom Resource Definition**: The Cutom Resource need to identify a `SecretName` field.
+- **Control-Plane core-agent**: The agent needs to be modified to have an ability to read Kubernetes API objects.
+- **Data-Plane io-engine**: io-engine need to be able to create and place a crypto block device on top of base block device of the diskpool.
+
+
```
Abhinandan-Purkait, line 109: We probably need to include a test plan as well. Also I feel we need to interface it a way that the testing can be done out of k8s ex. deployer. Maybe read via a file in non-k8s env for now, just for testing?

designs/replicated-pv/mayastor/encryption-at-rest.md
Alex130469: Not sure the above is correct, because encryption is possible per PV, it is not possible to have unencrypted and encrypted PVs share the same pool, but with both type of pools present it is possible to have PVs turning encrypted/unencrypted depending on where the replica is placed in.

designs/replicated-pv/mayastor/encryption-at-rest.md
Alex130469: The above is correct for the PV but not for the pool, because we have not scoped to encrypt/decrypt pools after creation. The encryption state is determined at pool creation and the only way to change it is by deleting the pool and re-create it with a different encryption setting.

tiagolobocastro:
1809 is the number from the mayastor repo, IIRC it has to be an issue on openebs created with the OEP template

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
+---
+oep-number: OEP 1809
+title: OpenEBS Enhancement Proposal for Mayastor At-Rest Encryption of data
```
tiagolobocastro, line 3: Let's include Mayastor on the PR title as well to make it clear this is targeting mayastor only

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
+---
+oep-number: OEP 1809
```
tiagolobocastro, line 2: Needs fixing as mentioned

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+title: OpenEBS Enhancement Proposal for Mayastor At-Rest Encryption of data
+authors:
+  - "@dsharma-dc"
+owners:
+  - "@dsharma-dc"
+  - "@Abhinandan-Purkait"
+editor: TBD
+creation-date: 2025-01-24
+last-updated: 2025-01-24
+status: implementable
```
tiagolobocastro, line 12: Proposal is provisional until accepted as implementable by the maintainers

designs/replicated-pv/mayastor/encryption-at-rest.md
tiagolobocastro: I wondered about this on another OEP...
I'd suggest just mayastor is fine, otherwise it's quite a mouthful to write every time?

designs/replicated-pv/mayastor/encryption-at-rest.md
tiagolobocastro: > Do you think we should also mention not implementing additional security protocols for communication between various services?

Yeah we might want to mention here the inter-service communication and that we rely on RBAC?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+2. [Motivation](#motivation)
+3. [Goals](#goals)
+4. [Non-Goals](#non-goals)
+5. [Proposal](#proposal)
+6. [User Stories](#user-stories)
+7. [Implementation Details](#implementation-details)
+
+---
+
+## 1. Overview
```
tiagolobocastro, line 29: IIRC numbering is not part of the template

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+## 2. Motivation
+
+Implementing data encryption at rest ensures that sensitive information remains secure even if storage media is lost, stolen, or compromised. It protects against unauthorized access by ensuring data is only readable by authorized users or systems. Encryption also helps meet regulatory compliance requirements, demonstrating a commitment to privacy and data protection. Additionally, it mitigates the risk of data breaches, maintaining trust with stakeholders.
+
+## 3. Goals
+
+- Implement the at-rest encryption of data in Mayastor, at a diskpool level so that all user data on the diskpool is stored encrypted.
+- Support FIPS compliant encryption methods.
+
+## 4. Non-Goals
```
tiagolobocastro, line 50: I think some of what was mentioned in the overview probably should be listed here:
- It is not possible to disable encryption a pool that has been created with encryption.
- The encryption can not be enabled on a pool that is already created un-encrypted.

Also worth mentioning:
- transparently changing the key for an encrypted pool is not supported as this would require a full pool rebuild

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+- This proposal does not implement (or integrate with) any Key management service for the encryption keys.
+- This proposal does not implement the encryption support per-volume.
+- This proposal does not implement in-flight data encryption.
+
+## 5. Proposal
+
+### Key Concepts
+
+1. **Diskpool Level Encryption**: Ability to encrypt data at a pool level without differentiating which replica the data belongs to. All replicas' data on such a pool will get encrypted
+2. **Encryptio Key Management**: A key provisioned by admin or a Key Management Service(KMS) for a supported cipher, for the use of storage system, and stored as a Kubernetes Secret. The admin can additionally safeguarded the key via Resource-Encryption-at-Rest facility provided by Kubernetes.
```
tiagolobocastro, line 61:
```suggestion
2. **Encryption Key Management**: A key provisioned by admin or a Key Management Service(KMS) for a supported cipher, for the use of storage system, and stored as a Kubernetes Secret. The admin can additionally safeguarded the key via Resource-Encryption-at-Rest facility provided by Kubernetes.
```

designs/replicated-pv/mayastor/encryption-at-rest.md
tiagolobocastro: Maybe this would be good as a pre-requisite?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Workflow
+
+1. **Encryption Key Provisioning**
+   - The user/admin creates a Kubernetes Secret containing a key with supported cipher and length.
+   - The user/admin optionally secures the created Secret via Resource-Encryption-at-Rest facility provided by Kubernetes.
+
+2. **Encrypted Diskpool Creation**:
+   - The user/admin creates a diskpool yaml spec that contains the name of the Secret which holds the key.
```
tiagolobocastro, line 70: A yaml example here would be nice

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+4. [Non-Goals](#non-goals)
+5. [Proposal](#proposal)
+6. [User Stories](#user-stories)
+7. [Implementation Details](#implementation-details)
+
+---
+
+## 1. Overview
+
+This proposal introduces the encryption of data at-rest in the Mayastor diskpools. This feature ensures that:
```
tiagolobocastro, line 31: Missing here a bit of information about volume scheduling on encrypted pools.

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+   - The user/admin creates a diskpool yaml spec that contains the name of the Secret which holds the key.
+   - When the spec is applied, the diskpool operator picks up this request and dispatches the create pool request containing Secret name
+   to the Mayastor agents that complete the diskpool provisioning.
+
+3. **Diskpool import upon node or io-engine restart**:
+   - In the event of a node restart, io-engine restart or a node going offline and coming up again - the diskpool
+   is imported. The import needs to be done using the same DEK that was used during diskpool creation.
+
+4. **Volume Provisioning**:
+   - Pool topology rules ensure that for a volume requesting encryption, the replicas are only placed on the diskpools that have
```
tiagolobocastro, line 79: How does volume provisioning work here.
Can all volumes be placed on encrypted pools by default? Or must their SC explicitly enable encrypted pools for example?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,109 @@
...
+   encryption enabled on them.
```
tiagolobocastro, line 80: What about upgrades, can you move existing volumes to encrypted pools?

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: > I wondered about this on another OEP... I'd suggest just mayastor is fine, otherwise it's quite a mouthful to write every time?

Yeah. I'll mention the first occurence of it in this doc as `Replicated PV Mayastor` and all subsequent ones as `Mayastor` only.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: @Alex130469  Yes that understanding is correct. `Migrate` word is seemingly not appropriate here as we are indeed going to create a different pool altogether. `Migrate` would be apt from context of volume here.
@Abhinandan-Purkait  I'll remove the mentions of this scenario for now until we decide how we want to deal migrating replicas across pools.

dsharma-dc:
> 1809 is the number from the mayastor repo, IIRC it has to be an issue on openebs created with the OEP template

Transferred the issue to openebs repo now.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: referred the local-bind-volumes OEP where it's numbered.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: Elaborated a bit more to mention poolHasTopologyKey.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: added a sample yaml.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: added these details.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: OEP number changed to openebs repo issue now.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: Removed this story for now until we finalise on migration strategy to new pool during upgrade.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: moved above, mentioned as pre-requisite.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: removed the per-volume statement, and mentioned about services' communication aspect.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: Not in MVP but we have to tackle that sooner or later.

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,119 @@
...
+```
+apiVersion: "openebs.io/v1beta2"
+kind: DiskPool
+metadata:
+  name: <pool-name>
+  namespace: <namespace>
+spec:
+  node: <node-name>
+  disks: ["/dev/disk/by-id/<id>"]
+  secretName: <myKeySecretName>
```
niladrih: This should be a bit more extendable? E.g.
```yaml
apiVersion: "openebs.io/v1beta2"
kind: DiskPool
metadata:
  name: <pool-name>
  namespace: <namespace>
spec:
  node: <node-name>
  disks: ["/dev/disk/by-id/<id>"]
  encryptionKeyConfig:
    type: Secret
    config:
      name: <myKeySecretName>
```
... this would let us plug in more config for a kms, e.g.
```yaml
apiVersion: "openebs.io/v1beta2"
kind: DiskPool
metadata:
  name: <pool-name>
  namespace: <namespace>
spec:
  node: <node-name>
  disks: ["/dev/disk/by-id/<id>"]
  encryptionKeyConfig:
    type: Vault
    config:
      url: <vaultServiceUrl>
      username: <vaultAuthUser>
      password: <vaultPasswd>
```

Abhinandan-Purkait: @niladrih Should that also come from a secret rather that adding them in plain text in the DSP CR itself.

```yaml
apiVersion: "openebs.io/v1beta2"
kind: DiskPool
metadata:
  name: <pool-name>
  namespace: <namespace>
spec:
  node: <node-name>
  disks: ["/dev/disk/by-id/<id>"]
  encryptionKeyConfig:
    type: Vault
    config:
      name: <myVaultCredsSecretName>
```

niladrih: Yeah, that bit is left to the best practices associated with the specific kms. I mean to suggest that the stucture should be extensible.

designs/replicated-pv/mayastor/encryption-at-rest.md
tiagolobocastro: oh, that's wrong as well then

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: expanded this.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: As agreed we will design it to do via control-paths. k8s API call will have standard RBAC security, as well as tls is there if we wish to leverage at some point.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: > We probably need to include a test plan as well. Also I feel we need to interface it a way that the testing can be done out of k8s ex. deployer. Maybe read via a file in non-k8s env for now, just for testing?

Yes, we'll have to take care of handling the testing part in non-k8s setup. But do we really need test plan in OEP?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,122 @@
...
+      name: <myKeySecretName>
+```
+
+2. **Diskpool import upon node or io-engine restart**:
+   - In the event of a node restart, io-engine restart or a node going offline and coming up again - the diskpool
+   is imported. The import needs to be done using the same DEK that was used during diskpool creation.
+
+3. **Volume Provisioning**:
+   - Pool topology rules ensure that for a volume requesting encryption, the replicas are only placed on the diskpools that have
+   encryption enabled on them. This will be handled by storageclass via the poolHasTopologyKey setting to let the volume replica placement happen on pools that are labelled with a specific key identifying encryption.
```
tiagolobocastro, line 100: So this would be a `reserved` topology key that cannot be modified by user and is automatically set on pools which are encrypted (by the control-plane)?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,122 @@
...
+   - Once the pool is created using a Secret, the key for that pool can't be transparently changed via a different Secret. Doing so will require a full rebuild of pool onto a different pool.
+
+- **Secret Name during Pool Import**: 
+  - Set the `secretName` in the PoolSpec.
+  - When a pool import is required, use the `secretName` from PoolSpec to fetch the DEK again.
+  - Dispatch the import operation to data plane.
+
+### Components to Update
+
+- **Diskpool Custom Resource Definition**: The Cutom Resource need to identify a `secretName` field.
```
tiagolobocastro:
```suggestion
- **Diskpool Custom Resource Definition**: The Custom Resource needs to identify a `secretName` field.
```

designs/replicated-pv/mayastor/encryption-at-rest.md
tiagolobocastro: Yeah I think we should highlight a general test plan here
And then later on we could add some BDD's here as well?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,122 @@
...
+    type: Secret
+    config:
+      name: <myKeySecretName>
+```
+
+2. **Diskpool import upon node or io-engine restart**:
+   - In the event of a node restart, io-engine restart or a node going offline and coming up again - the diskpool
+   is imported. The import needs to be done using the same DEK that was used during diskpool creation.
+
+3. **Volume Provisioning**:
```
tiagolobocastro: Would be good to see what the `StorageClass` for an encrypted volume would look like? And for a non-encrypted too.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: > Yeah I think we should highlight a general test plan here And then later on we could add some BDD's here as well?

I can add a Test related section in the template as well if that's helpful for future OEPs too. wdyt?

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: > So this would be a `reserved` topology key that cannot be modified by user and is automatically set on pools which are encrypted (by the control-plane)?

Yes, not giving an **additional** knob to users for label key would be better. It'd be implied that if a pool is encrypted, then we should be ale to use it for volumes requesting encryption.

designs/replicated-pv/mayastor/encryption-at-rest.md
tiagolobocastro: That would be good. (though one might already exist IIRC)

tiagolobocastro: No, it doesn't exist! Would be great if you could add it, thanks!

tiagolobocastro (approved):
Whilst provisional we can merge as often as needed

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,130 @@
...
+5. [Proposal](#proposal)
+6. [User Stories](#user-stories)
+7. [Implementation Details](#implementation-details)
+8. [Testing](#testing)
+
+---
+
+## Overview
+
+This proposal introduces the encryption of data at-rest in the Replicated PV Mayastor(hereafter referred as Mayastor) diskpools. This feature ensures that:
```
niladrih:
```suggestion
This proposal introduces the encryption of data at-rest in the Replicated PV Mayastor (hereafter referred as Mayastor) diskpools. This feature ensures that:
```

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,130 @@
...
+- This proposal does not implement support for disabling encryption on a pool that has been created with encryption.
+- This proposal does not implement support for enabling encryption on a pool that is already created as un-encrypted.
+- This proposal does not implement or strengthen any existing communication protocols between various services.
+
+## Proposal
+
+### Key Concepts
+
+1. **Diskpool Level Encryption**: Ability to encrypt data at a pool level without differentiating which replica the data belongs to. All replicas' data on such a pool will get encrypted.
+2. **Encryption Key Management**: This is a pre-requisite.
```
niladrih:
```suggestion
2. **Encryption Key Management**: This is a prerequisite.
```

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,130 @@
...
+   - A key is provisioned by admin or a Key Management Service(KMS) with a supported cipher, for the use of storage system, and stored as a Kubernetes Secret.
```
niladrih:
```suggestion
   - A key is provisioned by admin or a Key Management Service (KMS) with a supported cipher, for the use of storage system, and stored as a Kubernetes Secret.
```

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Workflow
+
+1. **Encrypted Diskpool Creation**:
+   - The user/admin creates a diskpool yaml spec that contains the name of the Secret which holds the key.
+   - When the spec is applied, the diskpool operator picks up this request and dispatches the create pool request containing Secret name to the Mayastor agents that complete the diskpool provisioning.
+   - The spec for creating a pool with encryption will look like below:
+
+```
+apiVersion: "openebs.io/v1beta2"
```
niladrih: The apiVersion will need a bump?
```suggestion
apiVersion: "openebs.io/v1beta3"
```

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,130 @@
...
+
+- **Secret Name Parameter**: 
+   - Add a new field `secretName` to the diskpool CR. This fields holds the name of Secret object that contains actual DEK.
+   - The control plane agent parse the actual Data Encryption Key(DEK) using the `secretName`.
+   - The data plane receives the request to create the pool using the provided DEK.
+   - Once the pool is created using a Secret, the key for that pool can't be transparently changed via a different Secret. Doing so will require a full rebuild of pool onto a different pool.
+
+- **Secret Name during Pool Import**: 
+  - Set the `secretName` in the PoolSpec.
+  - When a pool import is required, use the `secretName` from PoolSpec to fetch the DEK again.
```
niladrih: This needs to reflect the revised spec which uses encryptionKeyConfig.

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,130 @@
...
+  - Dispatch the import operation to data plane.
+
+### Components to Update
+
+- **Diskpool Custom Resource Definition**: The Custom Resource need to identify a `secretName` field.
+- **Control-Plane core-agent**: The agent needs to be modified to have an ability to read Kubernetes API objects.
```
niladrih:
```suggestion
- **Control-Plane agent-core**: The agent needs to be modified to have an ability to read Kubernetes API objects.
```

niladrih (approved):

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Workflow
+
+1. **Encrypted Diskpool Creation**:
+   - The user/admin creates a diskpool yaml spec that contains the name of the Secret which holds the key.
+   - When the spec is applied, the diskpool operator picks up this request and dispatches the create pool request containing Secret name to the Mayastor agents that complete the diskpool provisioning.
+   - The spec for creating a pool with encryption will look like below:
+
+```
+apiVersion: "openebs.io/v1beta3"
```
abhilashshetty04, line 75: Do we need migration machanism here from older CR to v1beta3 CR?

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,141 @@
...
+   - Pool topology rules ensure that for a volume requesting encryption, the replicas are only placed on the diskpools that have
+   encryption enabled on them. This will be handled by storageclass via the poolHasTopologyKey setting to let the volume replica placement happen on pools that are labelled with a specific key identifying encryption.
+   - The storageclass definition required for volumes to be encrypted will have an additional field named `secure`, which needs to be set to `true` if encryption is required.
+```
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: mayastor-3
+parameters:
+  secure: "true"
```
abhilashshetty04, line 108: Can we call it `encrypt` ?

dsharma-dc: > Can we call it `encrypt` ?

Thinking a bit more on it - `secure` seems generic and user friendly enough. Let's say we decide to **secure** volumes in a different way than encryption e.g ACLs, RBAC or whatever, then this won't need to change.

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: > Do we need migration machanism here from older CR to v1beta3 CR?

Is that something w.r.t implementation or a documented procedure to be followed?

abhilashshetty04: We manage CRDs through code. Need to support for upgrade from older version.

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,141 @@
...
+
+```
+apiVersion: "openebs.io/v1beta3"
+kind: DiskPool
+metadata:
+  name: <pool-name>
+  namespace: <namespace>
+spec:
+  node: <node-name>
+  disks: ["/dev/disk/by-id/<id>"]
```
abhilashshetty04, line 82: Also add topology labels. Its part of v1beta2

designs/replicated-pv/mayastor/encryption-at-rest.md
abhilashshetty04: In that case wouldnt it be good if user can selectively enforce a set of security measures?

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: done

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: > In that case wouldnt it be good if user can selectively enforce a set of security measures?

possibly, but until we have other options to secure, there is no choice. The choices will be anyway new subfields to this, whenever that happens.

designs/replicated-pv/mayastor/encryption-at-rest.md
```diff
@@ -0,0 +1,146 @@
...
+
+- **Secret Name Parameter**: 
+   - Add a new field `encryptionKeyConfig` to the diskpool CR. This fields holds the name of Secret object that contains actual DEK.
+   - The control plane agent parse the actual Data Encryption Key(DEK) using the `encryptionKeyConfig`.
+   - The data plane receives the request to create the pool using the provided DEK.
+   - Once the pool is created using a Secret, the key for that pool can't be transparently changed via a different Secret. Doing so will require a full rebuild of pool onto a different pool.
+
+- **Secret Name during Pool Import**: 
+  - Set the `encryptionKeyConfig` in the PoolSpec.
+  - When a pool import is required, use the `encryptionKeyConfig` from PoolSpec to fetch the DEK again.
```
abhilashshetty04, line 130: Will control plane agent call k8s api to get secret here?

abhilashshetty04 (approved):

designs/replicated-pv/mayastor/encryption-at-rest.md
dsharma-dc: yes, elaborated that.

PR #3854, opened 2025-02-05
add kubeconfig path arg to localpv commands

abhilashshetty04:


plugin/src/cli_utils/localpv/hostpath/mod.rs
```diff
@@ -32,6 +32,10 @@ pub struct CliArgs {
     /// The Output, viz yaml, json.
     #[clap(global = true, default_value = OutputFormat::None.as_ref(), short, long)]
     pub output: OutputFormat,
+
+    /// Path to kubeconfig file.
+    #[clap(skip)]
```
tiagolobocastro, line 37: This is because it's global?

tiagolobocastro (approved):

plugin/src/cli_utils/localpv/hostpath/mod.rs
abhilashshetty04: Yes, We just copy from global arg to this. SImilar to how we handle namespace.

Abhinandan-Purkait (approved):

PR #3856, opened 2025-02-06
[OEP 3855] Enable TLS

Johnaius:
## Description

Add OEP for Enabling TLS. See OEP for details. 

## Outcome

Please review proposal


Abhinandan-Purkait (approved):
Thanks. Looks reasonable to me.

designs/tls/install-cert-manager.md
```diff
@@ -0,0 +1,187 @@
...
+
+As a user, I want to not have to worry about certificate creation or management.
+
+### Implementation Details/Notes/Constraints
+
+#### Cert Manager
+
+Cert-manager installation shall be done using [helm charts][certManagerHelmCharts]
+
+* Should there be discussion about drawbacks of depending on cert-manager or other strategies?
```
tiagolobocastro, line 81: Any suggestions on what other strategies would be?
We can highlight them on an Alternatives section: https://github.com/openebs/openebs/blob/develop/designs/oep-template.md#alternatives-optional

designs/tls/install-cert-manager.md
```diff
@@ -0,0 +1,187 @@
...
+
+> **_NOTE:_** We will utilize the secrets created to volumeMount certs on appropriate deployment or daemonset.
+
+#### HTTP as default
+
+Components should be implemented in such a way that they will communicate using HTTP if tls is not enabled and HTTPS if tls is enabled.
+
+#### etcd
+
+Will need direction on how to implement tls between etcd and it's communication partners
```
tiagolobocastro, line 178: Yes this would also require enabling TLS on the etcd chart
I think the etcd chart already supports this

designs/tls/install-cert-manager.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Test Plan
+
+We should validate the communications are secure, ensuring they are and that certicates can be rotated safely.  Would like to discuss actual implementation of tests for these changes.
+
+### Risks and Mitigations
+
+Mayastor being upgraded to new version with TLS enabled may not be straightforward.
```
tiagolobocastro, line 186: hmm indeed
Perhaps on upgrade, we shouldn't enable TLS, and let user explicitly enable it if they want to?

designs/tls/install-cert-manager.md
```diff
@@ -0,0 +1,187 @@
...
+{{- end }}
+```
+
+In this example:
+
+* The `rest-api-server` certificate is created and signed by the `ca-issuer`, which uses the root certificate.
+
+* The `isCA` field is set to `false`, indicating that this is not a CA certificate.
+
+> **_NOTE:_** We will utilize the secrets created to volumeMount certs on appropriate deployment or daemonset.
```
tiagolobocastro, line 162: How does this work with certificate rotation? IIRC the mounted data wouldn't change unless the pods are restarted?

designs/tls/install-cert-manager.md
Johnaius: Manually created is the only other thing I thought of. I wasn't sure if you wanted to depend on cert-manager. Does more thought need to go into that?

designs/tls/install-cert-manager.md
Johnaius: https://github.com/Johnaius/mayastor-extensions/blob/47b5a178d03491e67f61a7abc1711e7749fc7b9a/chart/values.yaml#L512-L517

designs/tls/install-cert-manager.md
Johnaius: yea thats the idea for sure.  I implemented a switch in the extensions repo.  I'm not sure of the upgrade process here, but we will certainly need to test switching from non-tls to tls.

designs/tls/install-cert-manager.md
Johnaius: That sounds about right

designs/tls/install-cert-manager.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Motivation
+
+Security
+
+### Goals
+
+1. Create TLS switch
+
+1. Install Cert-Manager
```
tiagolobocastro, line 51: I wonder, should install cert-manager be a goal, or should the goal be to use cert-manager?
With this a few questions:
1. Should we support using existing cert-manager?
2. Should we make cert-manager a pre-requisite rather than installing it as a dependency? This would also get around the CRD issue?

designs/tls/install-cert-manager.md
tiagolobocastro: I guess this is where we have options
1. no direct dependency on cert-manager by using `volumesMounts`
2. use cert-manager API, so we don't need pod restarts to refresh the certificate for example, but we would depend on the cert-manager API

Is my understanding correct?

designs/tls/install-cert-manager.md
tiagolobocastro: Another place for `TLS` is the agent/agent `gRPC` and agent/io-engine `gRPC`

designs/tls/install-cert-manager.md
Johnaius: > * Should we support using existing cert-manager?

could you clarify what you mean here? I'm not sure I follow

designs/tls/install-cert-manager.md
Johnaius: I havent looked into option 2 but have been working on implementing option 1

designs/tls/install-cert-manager.md
tiagolobocastro: Let's say someone already installed the cert-manager for their other projects.
Can we make use of it?

tiagolobocastro: Ah, found it. We have jaeger tracing for debugging, and for it to work, a pre-req is the cert-manager.
<https://github.com/jaegertracing/helm-charts/tree/main/charts/jaeger-operator#prerequisites>
So we could have cert-manager as a pre-requisite?
This brings other questions, can multiple instances of cert-manager be installed in a cluster?

Johnaius: I believe not. I remember reading in the [docs](https://cert-manager.io/v1.12-docs/installation/helm/#:~:text=care%20must%20be%20taken%20to%20ensure%20that%20it%20is%20installed%20exactly%20once) that cert-manager should only be install once in cluster. I'd be interested to test this scenario.  Could this be checked pre-install?  any thoughts on how to implement that? At the very least we'd need to outline this scenario in the docs and offer solutions on how to proceed.  we could enable tls but disable cert-manager in helm chart perhaps?

designs/tls/install-cert-manager.md
Johnaius: I think depending on cert-manager is the way.

Johnaius: I think cert-manager is the way. the only drawback I can think if is a slightly longer install time.  It appears cert-manager-webhook container takes roughly 90s to be ready to go.

designs/tls/install-cert-manager.md
tiagolobocastro: > we could enable tls but disable cert-manager in helm chart perhaps?

Yeah seems like we'd have to do that since cert-manager should only be installed once.
We do similar thing for etcd: <https://github.com/openebs/mayastor-extensions/blob/e620241d89d9d92066da4b2e35e593008fe3c27b/chart/values.yaml#L487>

Btw seems like we can get the crd installed through the cert-manager chart using: `cert-manager.crds.enabled` and `cert-manager.crds.keep`

niladrih (approved):

tiagolobocastro (approved):
We'll have to modify the owner as @Johnaius would not be working on this

PR #3860, opened 2025-02-11
ci: disable usage metrics for CI test runs

niladrih:
Changes:
1. Disabled analytics for the localpvs, disables call-home for mayastor
2. Set etcd probes to a lower initial delay
3. set etcd and nats to use a single replica.

### DIFF
--- a/.github/workflows/chart-lint.yml
+++ b/.github/workflows/chart-lint.yml
@@ -28,17 +28,17 @@ jobs:
         if: steps.changes.outputs.chart == 'true'
         uses: azure/setup-helm@v4
         with:
-          version: v3.14.3
+          version: v3.17.1
 
       - uses: actions/setup-python@v5
         if: steps.changes.outputs.chart == 'true'
         with:
-          python-version: '3.10'
+          python-version: '3.x'
           check-latest: true
 
       - name: Set up chart-testing
         if: steps.changes.outputs.chart == 'true'
-        uses: helm/chart-testing-action@v2.6.1
+        uses: helm/chart-testing-action@v2.7.0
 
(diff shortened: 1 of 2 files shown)


.github/workflows/chart-lint.yml
```diff
@@ -68,4 +68,4 @@ jobs:
 
       - name: Run chart-testing (install)
         if: steps.list-changed.outputs.changed == 'true'
-        run: ct install --config ct.yaml
+        run: ct install --config ct.yaml --helm-extra-set-args "--set localpv-provisioner.analytics.enabled=false,lvm-localpv.analytics.enabled=false,mayastor.etcd.livenessProbe.initialDelaySeconds=5,mayastor.etcd.readinessProbe.initialDelaySeconds=5,mayastor.etcd.replicaCount=1,mayastor.obs.callhome.sendReport=false,mayastor.nats.cluster.enabled=false,mayastor.nats.cluster.replicas=1,zfs-localpv.analytics.enabled=false"
```
tiagolobocastro: Can't these be set of the ct.yaml?

pchandra19: I would also suggest if this can be a part of ct.yaml. Also, I think "Run chart-testing (lint)" step should be run with the same parameters.

tiagolobocastro (commented):
Also, ct is currently not running at all
I think the version we're using is perhaps broken?

.github/workflows/chart-lint.yml
niladrih: Done.

tiagolobocastro (approved):

PR #3863, opened 2025-02-13
CI: ensure path filters work with protection settings

tiagolobocastro:
GitHub doesn't handle path filters for a workflow properly. If it doesn't run, then the branch protection setting prevents the PR from passing!
Instead, let's run the workflow but filter from the job steps.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -10,6 +10,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
+  CI: 1
 
 jobs:
   rust-linter-build:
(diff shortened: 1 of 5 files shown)


Abhinandan-Purkait (approved):

.github/workflows/rust.yaml
```diff
@@ -21,18 +18,33 @@ jobs:
...
         with:
           submodules: 'recursive'
+      - uses: dorny/paths-filter@v3
+        id: changes
+        with:
+          filters: |
+            rust:
+              - 'plugin/**'
+              - '**.nix'
+              - 'Cargo.toml'
```
abhilashshetty04: Will we add Cargo.lock to root? :)

abhilashshetty04 (approved):

pchandra19 (approved):

.github/workflows/rust.yaml
tiagolobocastro: Not sure, but this will work regardless :)

niladrih (approved):

PR #3868, opened 2025-02-14
build: add cargo workspace

abhilashshetty04:


.github/workflows/rust.yaml
```diff
@@ -50,11 +50,4 @@ jobs:
...
       - name: Build rust binaries
-        if: steps.changes.outputs.rust == 'true'
-        run: nix-shell --run "./scripts/rust/builder.sh"
-      - name: Ensure Cargo.lock is up to date
-        run: |
-          if git diff --name-only | grep -q 'Cargo.lock'; then
-            echo "Cargo.lock is not up to date!"
-            exit 1
-          fi
+        run: nix-shell --run "cargo build --bins"
```
tiagolobocastro, line 54: not sure why you're modifying this?

scripts/rust/builder.sh
tiagolobocastro, line 1: And this

tiagolobocastro (changes requested):

.github/workflows/rust.yaml
abhilashshetty04: Ohh, ill `keep Ensure Cargo.lock up to date`. We wont run the builder.sh as we have Cargo.toml in root now. `nix-shell --run "cargo build --bins"` was there previously also.

scripts/rust/builder.sh
abhilashshetty04: We dont need this right? Since it just does cargo build on plugin pkg.

Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[workspace]
+members = [
+    "plugin",
+]
+
+exclude = ["./mayastor"]
+resolver = "1"
```
Abhinandan-Purkait: Add blank line

Abhinandan-Purkait (approved):

.github/workflows/rust.yaml
niladrih: Does this work without the cd?

tiagolobocastro (approved):

niladrih (approved):

.github/workflows/rust.yaml
abhilashshetty04: Yes it runs on root. We can Cargo.toml there.

PR #3876, opened 2025-02-27
rename flag kube-config-path to kubeconfig

abhilashshetty04:


plugin/src/main.rs
```diff
@@ -27,29 +27,29 @@ impl CliArgs {
...
             None => {
-                let client = kube_proxy::client_from_kubeconfig(args.kube_config_path.clone())
+                let client = kube_proxy::client_from_kubeconfig(args.kubeconfig.clone())
                     .await
                     .map_err(|err| anyhow::anyhow!("{err}"))?;
                 client.default_namespace().to_string()
             }
         };
-        let path = args.kube_config_path.clone();
+        let path = args.kubeconfig.clone();
```
tiagolobocastro, line 36: hmm I think we also need to modify mayastor kube_config_path to be defined in the binary cli args and copied in. Otherwise this will clash.

abhilashshetty04: Now we have same flag in mayastor arg also.

tiagolobocastro (approved):

niladrih (approved):

PR #3886, opened 2025-03-21
add checkout submodule to release script

abhilashshetty04:
We need to checkout submodule as we use script from submodule in later steps.

### DIFF
--- a/.github/workflows/publish-release-artifacts.yml
+++ b/.github/workflows/publish-release-artifacts.yml
@@ -17,6 +17,8 @@ jobs:
           echo "ShouldUpdateKrew:"${{ !github.event.release.prerelease && (steps.latest-release.outputs.latestTag == github.event.release.tag_name) }}
           echo "latestTag=$LATEST_VERSION" >> "$GITHUB_OUTPUT"
       - uses: actions/checkout@v4
+        with:
+          submodules: recursive
       - name: Download Artifacts
         run: |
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (approved):

.github/workflows/publish-release-artifacts.yml
```diff
@@ -17,6 +17,9 @@ jobs:
           echo "ShouldUpdateKrew:"${{ !github.event.release.prerelease && (steps.latest-release.outputs.latestTag == github.event.release.tag_name) }}
           echo "latestTag=$LATEST_VERSION" >> "$GITHUB_OUTPUT"
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
tiagolobocastro: history is not needed?

tiagolobocastro (approved):

.github/workflows/publish-release-artifacts.yml
abhilashshetty04: Wont we need all commits? Without this i think it pulls only latest commit?

tiagolobocastro: Why do we need all commits here? We're not interacting with the git history right?

abhilashshetty04: Ohh, Does it merge all commits into one commit. Just want to make sure all changes will be pulled.

Issue #3835, opened 2025-01-17
Support branches and version independence


niladrih:
On the openebs/openebs repo, it's not possible to release patch versions of an older minor release, after the release of an newer minor version.
This was the case with the LocalPV's as well. The newer CI on the LocalPVs support release of patches for older release. Would be good to have this on the openebs/openebs repo also.

pchandra19:
PR : https://github.com/openebs/openebs/pull/3888 is up for review

pchandra19:
Based on the discussion , we decided to change the stragey for this and use workflow_dispatch entirely. PR : https://github.com/openebs/openebs/pull/3906

PR #3888, opened 2025-03-25
chore(ci): adding branch preparation workflow and scripts

pchandra19:
This PR addresses https://github.com/openebs/openebs/issues/3835

The workflow creates release branch based on user input and updates the dependencies versions accordingly.


.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,284 @@
...
+          labels: |
+            prepare-release-branch
+            automated-pr
+          draft: false
+          signoff: true
+          branch: "create-pull-request/patch-${{ github.event.inputs.release-branch }}"
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  update_develop_branch_on_release_branch_creation:
+    runs-on: ubuntu-22.04
```
tiagolobocastro: Why is previous on ubuntu-latest and this on 22.04?

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,284 @@
...
+          branch_name=${{ github.event.inputs.release-branch }}
+          nix-shell --pure --run "./scripts/helm/update-chart-version.sh --branch $branch_name --type release --chart-version ${VERSION}-develop  --localpv-provisioner-version ${LOCALPV_PROVISIONER_VERSION} --zfs-localpv-version ${ZFS_LOCALPV_VERSION} --lvm-localpv-version ${LVM_LOCALPV_VERSION} --mayastor-version ${MAYASTOR_VERSION}" ./shell.nix
+        env:
+          VERSION: ${{ github.event.inputs.helm-version }}
+          LOCALPV_PROVISIONER_VERSION: ${{ github.event.inputs.localpv-provisioner-version }}
+          ZFS_LOCALPV_VERSION: ${{ github.event.inputs.zfs-localpv-version }}
+          LVM_LOCALPV_VERSION: ${{ github.event.inputs.lvm-localpv-version }}
+          MAYASTOR_VERSION: ${{ github.event.inputs.mayastor-version }}
+
+        #   yq -i ".version = \"${VERSION}-prerelease\" | .appVersion = \"${APP_VERSION}-prerelease\"" charts/Chart.yaml
```
tiagolobocastro: Remove if unused

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,284 @@
...
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
+
+      - name: Get dependencies version from develop
+        id: extract-versions
+        run: |
+          GITHUB_API_URL="https://raw.githubusercontent.com"
+
+          # Define output file for GitHub Actions
+          OUTPUT_FILE="$GITHUB_OUTPUT"
```
tiagolobocastro, line 155: nit: why not just use the github variable directly as it's fairly well known?

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,284 @@
...
+      - name: Debug extracted versions
+        run: |
+          echo "Extracted LocalPV Provisioner Version: ${{ steps.extract-versions.outputs.localpv_provisioner_version }}"
+          echo "Extracted ZFS LocalPV Version: ${{ steps.extract-versions.outputs.zfs_localpv_version }}"
+          echo "Extracted LVM LocalPV Version: ${{ steps.extract-versions.outputs.lvm_localpv_version }}"
+          echo "Extracted Mayastor Version: ${{ steps.extract-versions.outputs.mayastor_version }}"
+
+      - name: Test the chart version updater script
+        run: |
+          nix-shell --pure --run "./scripts/helm/test-update-chart-version.sh" ./shell.nix
```
tiagolobocastro, line 174: If this is testing the script itself, shouldn't it run on develop CI?

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,284 @@
...
+        uses: peter-evans/create-pull-request@v5
+        with:
+          base: develop
+          commit-message: "chore(ci): update helm chart versions and/or git submodules"
+          committer: GitHub <noreply@github.com>
+          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          title: "Prepare develop branch on ${{ github.ref_name }} creation"
+          labels: |
+            update-develop-branch
+            automated-pr
```
tiagolobocastro: let's pick some labels which match our new labels now, we use k8s like labels

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,284 @@
...
+      localpv_provisioner_version: ${{ steps.extract-versions-from-tag.outputs.localpv_provisioner_version }}
+      zfs_localpv_version: ${{ steps.extract-versions-from-tag.outputs.zfs_localpv_version }}
+      lvm_localpv_version: ${{ steps.extract-versions-from-tag.outputs.lvm_localpv_version }}
+      mayastor_version: ${{ steps.extract-versions-from-tag.outputs.mayastor_version }}
+    runs-on: ubuntu-22.04
+    if: github.ref_type == 'tag' && github.event_name == 'push' && !contains(github.ref_name, 'rc')
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          #ref: ${{ github.ref_name }}
```
tiagolobocastro: remove if not needed

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,284 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  update_release_branch_after_release:
+    outputs:
+      localpv_provisioner_version: ${{ steps.extract-versions-from-tag.outputs.localpv_provisioner_version }}
+      zfs_localpv_version: ${{ steps.extract-versions-from-tag.outputs.zfs_localpv_version }}
+      lvm_localpv_version: ${{ steps.extract-versions-from-tag.outputs.lvm_localpv_version }}
+      mayastor_version: ${{ steps.extract-versions-from-tag.outputs.mayastor_version }}
+    runs-on: ubuntu-22.04
+    if: github.ref_type == 'tag' && github.event_name == 'push' && !contains(github.ref_name, 'rc')
```
tiagolobocastro: checking just for rc does not seem very robust, probably you want to check properly with the semvertool in a step, something like:

`[ $(semver get release $tag) != $tag ]`

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,284 @@
...
+
+      - name: Get dependencies version from release tag
+        id: extract-versions-from-tag
+        run: |
+          GITHUB_API_URL="https://api.github.com/repos"
+
+          # Define output file for GitHub Actions
+          OUTPUT_FILE="$GITHUB_OUTPUT"
+
+          # Fetch and extract versions
```
tiagolobocastro, line 253: Shouldn't we simply remove the -prerelease?
Otherwise this could point the release to an incorrect version. Ex: let's say we release a patch on a previous release, we can't just update the dependencies to their latest.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+            yq -i '(.dependencies[] | select(.name == "mayastor") | .version) |= sub("-prerelease$", "")' charts/Chart.yaml
+
+      - name: Check if the chart is publishable
+        run: |
+          TAG=${{ github.event.release.tag_name }}
+          nix-shell --pure --run "./scripts/helm/update-chart-version.sh --tag $TAG --publish-release" ./shell.nix
+
+      - name: Install Helm
+        run: |
+            curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
```
tiagolobocastro: what's this for?

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+            yq -i '(.dependencies[] | select(.name == "localpv-provisioner") | .version) |= sub("-prerelease$", "")' charts/Chart.yaml
+            yq -i '(.dependencies[] | select(.name == "zfs-localpv") | .version) |= sub("-prerelease$", "")' charts/Chart.yaml
+            yq -i '(.dependencies[] | select(.name == "lvm-localpv") | .version) |= sub("-prerelease$", "")' charts/Chart.yaml
+            yq -i '(.dependencies[] | select(.name == "mayastor") | .version) |= sub("-prerelease$", "")' charts/Chart.yaml
+
+      - name: Check if the chart is publishable
+        run: |
+          TAG=${{ github.event.release.tag_name }}
+          nix-shell --pure --run "./scripts/helm/update-chart-version.sh --tag $TAG --publish-release" ./shell.nix
```
tiagolobocastro, line 32: this is not enough, since we've updated the versions we have to re-generate the Chart.yaml metadata

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: cachix/install-nix-action@v22
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
+
+      - name: Remove prerelease label from dependencies
+        run: |
+            yq -i '(.dependencies[] | select(.name == "localpv-provisioner") | .version) |= sub("-prerelease$", "")' charts/Chart.yaml
```
tiagolobocastro: you might want to script this as IIRC yq messes with the blank lines or something like that

.github/workflows/release-workflow-dispatcher.yml
pchandra19: updated all to ubuntu-latest.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Done

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Done.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: This is no longer required as am using the script. I have removed it.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: AFAICT , thats the preferred way of using output variables in steps as opposed to env variable hence the approach.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: It is running on all 3 jobs before helm chart update script.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Not sure, we remove the '-prerelease' when we create a release. Patch releases won't be marked as latest. Maybe we can discuss this ?

.github/workflows/release.yml
pchandra19: Ah, I was under the impression that the chart-images.yaml (Helm Chart Images CI) would do that. Should I extend the workflow to include 
on:
  release:
    types:
      - "created"
OR 
Do the steps explicitly in this workflow.

.github/workflows/release.yml
pchandra19: sure.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: I have chosen area/helm and kind/ci as the labels

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Also , we can update the versions when the PR is created. It is not an automated merge considering. So once the PR is created it can always be manually edited.

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,266 @@
+name: Release Workflow Dispatcher
+on:
+  workflow_dispatch:
+    inputs:
+      openebs-crds-version:
+        required: true
+        default: ""
+      localpv-provisioner-version:
+        required: true
+        default: 4.2.0
```
Abhinandan-Purkait: Will you change the default after every release? Or it's just for example? Shoudn't we just error out if this is empty rather than defaults?

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,266 @@
...
+          committer: GitHub <noreply@github.com>
+          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          title: "Prepare release branch after release ${{ github.ref_name }}"
+          labels: |
+            area/helm
+            kind/ci
+          draft: false
+          signoff: true
+          branch: "create-pull-request/patch-${{ env.BASE }}"
+          token: ${{ secrets.GITHUB_TOKEN }}
```
Abhinandan-Purkait: Blank line?

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+          nix-shell --pure --run "./scripts/helm/update-chart-version.sh --tag $TAG --publish-release" ./shell.nix
+
+      - name: Debug Workflow Directory
+        run: ls -l ./charts
+
+      - name: Publish openebs umbrella helm chart
+        uses: stefanprodan/helm-gh-pages@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          charts_dir: .
```
Abhinandan-Purkait: Blank line?

scripts/helm/test-update-chart-version.sh
```diff
@@ -0,0 +1,96 @@
...
+         --tag "v1.2.3" --dry-run --chart-version "1.2.3-prerelease" --publish-release
+
+run_test "Test 15: Actual release rc tag to modify the chart versions" \
+         "1.2.3-rc" \
+         --tag "v1.2.3-rc" --dry-run --chart-version "1.2.3-prerelease" --publish-release
+
+if [ -n "$FAILED" ]; then
+  echo "Some of the tests have failed..."
+  exit 1
+fi
```
Abhinandan-Purkait: Blank line? Is this test run anywhere in the pipeline?

scripts/helm/update-chart-version.sh
```diff
@@ -0,0 +1,286 @@
...
+  if [[ -n $VERSION ]]; then
+    if [[ -z $DRY_RUN ]];then
+      update_chart_yaml "$VERSION" "$VERSION" "$LOCALPV_PROVISIONER_VERSION" "$ZFS_LOCALPV_VERSION" "$LVM_LOCALPV_VERSION" "$MAYASTOR_VERSION"
+    else
+      echo "$VERSION"
+    fi
+  else
+    die "Failed to update the chart versions"
+  fi
+fi
```
Abhinandan-Purkait: blank line?

scripts/helm/update-dependencies-version-upon-release.sh
```diff
@@ -0,0 +1,74 @@
...
+      exit 0
+      ;;
+    *)
+      help
+      die "Unknown option: $1"
+      ;;
+  esac
+done
+
+update_chart_yaml
```
Abhinandan-Purkait: Blank line.

scripts/helm/validate-chart-version.sh
```diff
@@ -0,0 +1,103 @@
...
+  RELEASE_VERSION=$(extract_major_minor "$BRANCH_NAME")
+  if [[ "$CHART_VERSION" != "$RELEASE_VERSION."*"-prerelease" ]]; then
+    die "Chart version must be in format $RELEASE_VERSION.X-prerelease for release branch"
+  fi
+  if [[ "$APP_VERSION" != "$RELEASE_VERSION."*"-prerelease" ]]; then
+    die "App version must be in format $RELEASE_VERSION.X-prerelease for release branch"
+  fi
+else
+   die "Unknown branch name: $BRANCH_NAME"
+fi
```
Abhinandan-Purkait: Same here.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: No, its just an example of how to set values. All parameters are required. If not provided/empty the workflow won't run and will ask the user to populate the field.

Abhinandan-Purkait (approved):

.github/workflows/release-workflow-dispatcher.yml
tiagolobocastro: So it won't show up as the default value when running the workflow and user still has to manually type it? If so then it's good, otherwise I agree with Abhinandan, I think better to leave empty to avoid mistakes

.github/workflows/release-workflow-dispatcher.yml
tiagolobocastro: Shouldn't it run on develop CI instead?

.github/workflows/release-workflow-dispatcher.yml
tiagolobocastro, line 1: should we break down this file, not all jobd here are triggered by workflow dispatch?

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,266 @@
+name: Release Workflow Dispatcher
+on:
+  workflow_dispatch:
+    inputs:
+      openebs-crds-version:
```
tiagolobocastro: Do we need this as input or just set it to openebs version? Can't remember if we keep bumping this or not?

.github/workflows/release-workflow-dispatcher.yml
tiagolobocastro: > Not sure, we remove the '-prerelease' when we create a release. Patch releases won't be marked as latest. Maybe we can discuss this ?

Yeah but the chart without -prerelease is not actually checked in, it's only pushed to the helm registry?

.github/workflows/release.yml
tiagolobocastro: I'm confused, how is extending the chart-image helping here?

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Yes, thats correct. I just followed the design and workflow of other repos. It's the same behaviour across.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: I'll remove the default values and make them empty.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Not required anymore. We are setting it to openebs version already. Removed.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: I would keep them all in one file. They might not all be triggered by workflow dispatcher but are all part of the branch preparation.

.github/workflows/release.yml
pchandra19: I have extended chart-images.yml to be used as a resubale workflow and calling it from branch preparation to avoid dupilcation of code.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Done. Included the job in pr.yml

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,267 @@
...
+        required: true
+      helm-version:
+        required: true
+      app-version:
+        required: true
+  push:
+    branches:
+      - "release/**"
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+**"
```
niladrih, line 23: Branch preparation during tagging?

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,267 @@
...
+
+      - name: Modify chart version upon release branch creation.
+        run: |
+          branch_name=${{ github.event.inputs.release-branch }}
+          nix-shell --pure --run "./scripts/helm/update-chart-version.sh --branch $branch_name --type release --chart-version ${VERSION}-develop  --localpv-provisioner-version ${LOCALPV_PROVISIONER_VERSION} --zfs-localpv-version ${ZFS_LOCALPV_VERSION} --lvm-localpv-version ${LVM_LOCALPV_VERSION} --mayastor-version ${MAYASTOR_VERSION}" ./shell.nix
+        env:
+          VERSION: ${{ github.event.inputs.helm-version }}
+          LOCALPV_PROVISIONER_VERSION: ${{ github.event.inputs.localpv-provisioner-version }}
+          ZFS_LOCALPV_VERSION: ${{ github.event.inputs.zfs-localpv-version }}
+          LVM_LOCALPV_VERSION: ${{ github.event.inputs.lvm-localpv-version }}
```
niladrih, line 85: nit: let's call the envs LOCALPV_ZFS_VERSION, LOCALPV_LVM_VERSION, LOCALPV_HOSTPATH_VERSION

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,267 @@
...
+          echo "zfs_localpv_version=$(curl -s $GITHUB_API_URL/openebs/zfs-localpv/refs/heads/develop/deploy/helm/charts/Chart.yaml | yq '(.version | sub("-develop$"; ""))')" >> "$OUTPUT_FILE"
+          echo "lvm_localpv_version=$(curl -s $GITHUB_API_URL/openebs/lvm-localpv/refs/heads/develop/deploy/helm/charts/Chart.yaml | yq '(.version | sub("-develop$"; ""))')" >> "$OUTPUT_FILE"
+          echo "mayastor_version=$(curl -s $GITHUB_API_URL/openebs/mayastor-extensions/refs/heads/develop/chart/Chart.yaml | yq '(.version | sub("-develop$"; ""))')" >> "$OUTPUT_FILE"
+
+      - name: Debug extracted versions
+        run: |
+          echo "Extracted LocalPV Provisioner Version: ${{ steps.extract-versions.outputs.localpv_provisioner_version }}"
+          echo "Extracted ZFS LocalPV Version: ${{ steps.extract-versions.outputs.zfs_localpv_version }}"
+          echo "Extracted LVM LocalPV Version: ${{ steps.extract-versions.outputs.lvm_localpv_version }}"
+          echo "Extracted Mayastor Version: ${{ steps.extract-versions.outputs.mayastor_version }}"
```
niladrih, line 170:
```suggestion
          echo "Extracted LocalPV Hostpath Version: ${{ steps.extract-versions.outputs.localpv_provisioner_version }}"
          echo "Extracted LocalPV ZFS Version: ${{ steps.extract-versions.outputs.zfs_localpv_version }}"
          echo "Extracted LocalPV LVM Version: ${{ steps.extract-versions.outputs.lvm_localpv_version }}"
          echo "Extracted Mayastor Version: ${{ steps.extract-versions.outputs.mayastor_version }}"
```

scripts/helm/update-chart-version.sh
```diff
@@ -0,0 +1,276 @@
...
+  diff_out=$(diff -B <(yq '.' "$2") <(yq "$1" "$2"))
+  error=$?
+  if [ "$error" != "0" ] && [ "$error" != "1" ]; then
+    exit "$error"
+  fi
+  if [ -n "$diff_out" ]; then
+    echo "$diff_out" | patch --quiet --no-backup-if-mismatch "$2" -
+  fi
+  set -euo pipefail
+}
```
niladrih, line 69: Could use `source "$ROOT_DIR/mayastor/scripts/utils/yaml.sh"` instead and use the yq_ibl from the extensions repo. See the ./scripts/helm/images.sh script on this repo.

.github/workflows/release.yml
tiagolobocastro: But this won't work, the contents on this job won't get updated

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Yes. On tag (or release creation) we have to move the release branch forward.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Not fixing at the moment.

.github/workflows/release-workflow-dispatcher.yml
pchandra19: Not fixing at the moment

.github/workflows/release.yml
pchandra19: Added the steps to each workflow.

scripts/helm/update-chart-version.sh
pchandra19: Well, that may be a smart way to do stuff but I'd not like to over complicate stuff for 12 LOC.

.github/workflows/pr.yml
```diff
@@ -20,3 +20,23 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: recursive
+
+      - uses: cachix/install-nix-action@v22
+      - name: Pre-populate nix-shell
+        run: |
+              export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
```
tiagolobocastro: seems too indented compare to usual

.github/workflows/release-workflow-dispatcher.yml
```diff
@@ -0,0 +1,321 @@
...
+          LOCALPV_PROVISIONER_VERSION: ${{ steps.extract-versions.outputs.localpv_provisioner_version }}
+          ZFS_LOCALPV_VERSION: ${{ steps.extract-versions.outputs.zfs_localpv_version }}
+          LVM_LOCALPV_VERSION: ${{ steps.extract-versions.outputs.lvm_localpv_version }}
+          MAYASTOR_VERSION: ${{ steps.extract-versions.outputs.mayastor_version }}
+
+      - name: Pre-populate helm nix-shell
+        run: nix-shell ./scripts/helm/shell.nix --run "echo"
+
+      - name: Generate image list
+        run: nix-shell ./scripts/helm/shell.nix --run "./scripts/helm/images.sh generate --dependency-update --exit-code"
```
tiagolobocastro: Don't think we want to use exit-code here since we actually want to modify the metadata here, otherwise this would fail?

scripts/helm/update-chart-version.sh
```diff
@@ -0,0 +1,276 @@
...
+      ;;
+  esac
+done
+
+# # Validate required arguments
+# if [[ -z "$LOCALPV_PROVISIONER_VERSION" || -z "$ZFS_LOCALPV_VERSION" || -z "$LVM_LOCALPV_VERSION" || -z "$MAYASTOR_VERSION" ]]; then
+#   echo "Error: All mandatory dependencies versions must be provided."
+#   help
+#   exit 1
+# fi
```
tiagolobocastro: Remove if not needed

tiagolobocastro (commented):
Pls check the exit-code comment, would be good if you could test this on a fork

scripts/helm/update-chart-version.sh
pchandra19: Done.

.github/workflows/pr.yml
pchandra19: Fixed.

pchandra19:
I am closing this PR , have created https://github.com/openebs/openebs/pull/3906 with the workflow dispatch as discussed.

PR #3901, opened 2025-04-09
fix for breaking change on nix build

abhilashshetty04:


r1jt (approved):

niladrih (approved):
This is a good idea, until openebs-upgrade has its own nix-build target.

PR #3902, opened 2025-04-10
Fix the plugin scripted build

tiagolobocastro:
ci: build the upgrade crate
    
    Ensures that doesn't break by accident.
    
---

    ci: run release.sh on pr test
    
    We should really look into refactoring the release.sh to use the utils
    build for the plugin, maybe using a new argument to avoid breaking changes.
    Also we should check if we still need the older build model, or can the
    utils be adopted as the default build?
    
---

    fix(plugin): use correct paths for build


abhilashshetty04 (approved):

.github/workflows/pr.yml
```diff
@@ -20,3 +20,50 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
+      - uses: DeterminateSystems/nix-installer-action@v14
+        if: needs.plugin-changed.outputs.rust_changed == 'true'
+      - name: Test building the static binaries
+        if: needs.plugin-changed.outputs.rust_changed == 'true'
+        # todo: modify downstream release.sh to build plugin via utils
+        #  perhaps it's time to rename utils or convert all to the utils build
+        run: nix-build -A utils.debug.x86_64.linux-musl.kubectl-openebs --arg incremental false
```
abhilashshetty04, line 57: We have this already

https://github.com/openebs/openebs/blob/develop/.github/workflows/pr-build.yml#L33

tiagolobocastro: Yes, I simply combined multiple PR files in 1

PR #3906, opened 2025-04-17
chore(ci): Branch Preparation using workflow dispatch

pchandra19:
This PR addresses : https://github.com/openebs/openebs/issues/3835

Couple of points to note:
-The approach uses workflow dispatch to update the versions. 
-We will have to ensure that the local engine follow the correct versioning and release branch format. It can cause failures during helm dependency update if the version being set is not found (due to chart version constraint), so please make sure the exact versions are passed which match are present in local engine repos.

### DIFF
--- a/.github/workflows/branch-preparation.yml
+++ b/.github/workflows/branch-preparation.yml
@@ -1,0 +1,138 @@
+name: Branch Preparation
+
+on:
+  workflow_dispatch:
+    inputs:
+      localpv-provisioner-version:
+        required: true
+      zfs-localpv-version:
+        required: true
+      lvm-localpv-version:
+        required: true
+      mayastor-version:
+        required: true
+      branch:
+        required: true
+      helm-version:
+        required: true
+      app-version:
+        required: true
+
+jobs:
+  create-release-branch:
+    runs-on: ubuntu-latest
+    outputs:
+      branch_exists: ${{ steps.check-branch.outputs.exists }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: develop
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: Check if branch exists
+        id: check-branch
+        run: |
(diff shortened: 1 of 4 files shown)


Abhinandan-Purkait (approved):

.github/workflows/branch-preparation.yml
```diff
@@ -0,0 +1,138 @@
...
+          set -euo pipefail
+          sudo debconf-communicate <<< "set man-db/auto-update false" || true
+          sudo dpkg-reconfigure man-db || true
+          nix-shell ./scripts/k8s/shell.nix --run "./scripts/k8s/deployer.sh start --mayastor --zfs --lvm --label"
+
+      - name: Install helm chart
+        run: nix-shell ./scripts/helm/shell.nix --run "./scripts/helm/install.sh --mayastor --zfs --lvm --wait"
+
+      - name: Verify image list
+        run: nix-shell ./scripts/helm/shell.nix --run "./scripts/helm/images.sh verify"
```
tiagolobocastro, line 108: I think we can skip these because they will be tested on the PR anyway

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+          submodules: recursive
+
+      - uses: cachix/install-nix-action@v22
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
+
+      - name: Remove prerelease label from dependencies
```
tiagolobocastro, line 25: from dependencies but also from the chart itself right?

tiagolobocastro (approved):
We still need a workflow for modifying the develop?

pchandra19:
> We still need a workflow for modifying the develop?

Same workflow will work for develop as well.

PR #3907, opened 2025-04-18
feat: add upgrade plugin

niladrih:
Changes:
- Renames upgrade crate to openebs-upgrade
- Adds `kubectl openebs upgrade` command
- Adds `kubectl get upgrade-status` command
- Adds Event watcher which watches for initial Upgrade Event


plugin/src/cli_utils/upgrade/k8s/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+                }
+            }
+
+            openebs_chart_cm.ok_or(anyhow!("Failed to figure out the release name of the 'openebs' helm chart, there aren't any 'openebs' charts installed in the '{namespace}' namespace"))
+        }
+        unsupported_storage_driver => Err(anyhow!(
+            "'{unsupported_storage_driver}' storage driver for helm is not supported"
+        )),
+    }
+}
```
Abhinandan-Purkait: Add blank line.

plugin/src/cli_utils/upgrade/k8s/resources.rs
```diff
@@ -0,0 +1,316 @@
...
+            labels: Some(upgrade_labels!()),
+            name: Some(format!("{release_name}-upgrade-config-map-{version}", release_name = release_name.as_str(), version = upgrade_obj_suffix())),
+            namespace: Some(namespace.to_string()),
+            ..Default::default()
+        },
+        data: Some(data),
+        immutable: Some(true),
+        ..Default::default()
+    }
+}
```
Abhinandan-Purkait: Blank line

plugin/src/constants.rs
```diff
@@ -0,0 +1,17 @@
...
+}
+
+/// The default image tag to the upgrade-job image.
+pub const UPGRADE_JOB_IMAGE_TAG: &str = "develop";
+/// The default image registry for container images.
+pub const DEFAULT_IMAGE_REGISTRY: &str = "docker.io";
+/// ConfigMap mount path for upgrade.
+pub const UPGRADE_CONFIG_MAP_MOUNT_PATH: &str = "/upgrade-config-map";
+
+pub const PRODUCT_NAME: &str = "openebs";
```
Abhinandan-Purkait: blank line

Abhinandan-Purkait (approved):

plugin/Cargo.toml
```diff
@@ -17,23 +17,28 @@ tls = ["openapi/tower-client-tls", "rest-plugin/tls"]
...
-k8s-openapi = { version = "0.22.0", features = ["v1_24"] }
-kube = { version = "0.94.2", features = ["derive", "runtime"] }
 lazy_static = "1.5.0"
 prettytable-rs = "0.10.0"
 serde = "1.0.214"
 serde_json = {version = "1.0.132"}
 serde_yaml = "0.9.34"
 snafu = "0.8.5"
-tokio = { version = "1.43.1", features=["full"] }
+tokio = { version = "1.41.0", features=["full"] }
```
tiagolobocastro: why are crates here being downgraded?

plugin/src/cli_utils/upgrade/k8s/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+            .ok_or(anyhow!("No data in helm {driver}"))?
+            .into_iter()
+            .find_map(|(k, v)| k.eq("release").then_some(v))
+            .ok_or(anyhow!(
+                "No value mapped to the 'release' key in helm {driver}"
+            ))
+    }};
+}
+
+pub fn upgrade_name_concat(release_name: &str, component_name: &str) -> String {
```
tiagolobocastro, line 41: add doc comments to pubs

plugin/src/cli_utils/upgrade/k8s/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    format!("{release_name}-{component_name}-{release_version}", release_version = upgrade_obj_suffix())
+}
+
+
+pub async fn delete_older_upgrade_events(release_name: &str, component_name: &str, kube_event_client: Api<Event>) -> Result<()> {
```
tiagolobocastro: older than what?
add doc comments explaining pls

plugin/src/cli_utils/upgrade/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
+use tokio::{spawn, try_join};
+use tokio::task::JoinHandle;
+use upgrade::common::kube::client::list_pods;
+use openebs_upgrade::error::FailedTokioSpawn;
+
+pub mod cli;
+pub mod k8s;
+
+pub struct ImageProperties {
```
tiagolobocastro, line 45: What is this for? Add doc comments please

plugin/src/cli_utils/upgrade/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    pub pull_secrets: Option<Vec<k8s_openapi::api::core::v1::LocalObjectReference>>,
+    pub registry: String,
+    pub pull_policy: Option<String>,
+}
+
+impl ImageProperties {
+    pub async fn new_from_helm_release(ns: &str, release_name: &str, args: &UpgradeArgs) -> Result<Self> {
+        let hostpath_localpv_image_name = format!("{release_name}-localpv-provisioner");
+        let mut openebs_containers: HashSet<&str> = [
+            "api-rest",
```
tiagolobocastro, line 77: I guess this would be clearer with comments, what are these for, and why these specifically?

plugin/src/cli_utils/upgrade/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+                                        ..Default::default()
+                                    }),
+                                    ..Default::default()
+                                }),
+                                ..Default::default()
+                            },
+                            EnvVar {
+                                // Ref: https://github.com/helm/helm/blob/main/cmd/helm/helm.go#L76
+                                name: "HELM_DRIVER".to_string(),
+                                value: env::var("HELM_DRIVER").ok(),
```
tiagolobocastro: Shouldn't this come from the cli argument?

plugin/src/cli_utils/upgrade/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    let release_name = match args.release_name.as_ref() {
+        Some(name) => name.clone(),
+        None => helm_release_name(namespace, args.helm_storage_driver.as_str()).await?
+    };
+
+    delete_older_upgrade_events(release_name.as_str(), "upgrade", Api::namespaced(k8s_client.clone(), namespace)).await?;
+
+    create_upgrade_resources(args, release_name.as_str(), namespace, k8s_client).await?;
+
+
```
tiagolobocastro, line 313: is this formatted? There's a lot of new lines here.

plugin/src/cli_utils/upgrade/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    K: Resource + Clone + std::fmt::Debug,
+{
+    let pp = PostParams::default();
+    client
+        .create(&pp, resource)
+        .await
+        .map(|_| ())
+        .or_else(|err| match err {
+            kubeError::Api(response) if response.reason.eq("AlreadyExists") => {
+                println!("{kind} '{name}' already exists", kind = resource.kind(), name = resource.name_unchecked());
```
tiagolobocastro: Shouldn't we check the contents match as well?
If the resources are specifically for upgrade, another way could be deleting them before hand, if they exist.

plugin/src/constants.rs
```diff
@@ -0,0 +1,17 @@
...
+
+/// The name suffix to the kubernetes upgrade-job resources and related resources.
+pub fn upgrade_obj_suffix() -> String {
+    version_info!().version_tag.unwrap_or(UPGRADE_JOB_IMAGE_TAG.to_string()).replace('.', "-")
+}
+
+/// The default image tag to the upgrade-job image.
+pub const UPGRADE_JOB_IMAGE_TAG: &str = "develop";
+/// The default image registry for container images.
+pub const DEFAULT_IMAGE_REGISTRY: &str = "docker.io";
```
tiagolobocastro, line 16: This is for our upgrade image I'm assuming?

niladrih:
@tiagolobocastro I know quite of bit of doc comments are missing, and they've got a pub on them, so I'll add that in soon, all of the code bits and the build bits are in.

plugin/src/cli_utils/upgrade/mod.rs
niladrih: The names are unique across versions. So, we should be okay to assume that the resources is the correct one. WDYT?

plugin/src/constants.rs
niladrih: Yes!

plugin/src/cli_utils/upgrade/mod.rs
niladrih: It is now..

plugin/src/cli_utils/upgrade/mod.rs
niladrih: It's being accepted in the same way that helm expects it. Helm introductory docs don't mention this input. I don't expect users to explicitly state this configuration. 

It's expected that the shell is being used for both the plugin and helm.

plugin/src/cli_utils/upgrade/mod.rs
niladrih: Noted, I will add more docs here.

plugin/src/main.rs
```diff
@@ -54,6 +56,14 @@ impl CliArgs {
                 operations.cli_args.namespace = ns;
                 operations.cli_args.kubeconfig = path
             }
+            cli_utils::Operations::Upgrade(ref mut upgrade_args) => {
+                upgrade_args.namespace = ns;
+                upgrade_args.kubeconfig = path
+            }
+            cli_utils::Operations::Get(ref mut get_able) => {
```
abhilashshetty04: Do we need to add `Get` command as a top level verb

Cant it be after `upgrade`

For eg: `kubectl openebs upgrade get status` / ` kubectl openebs upgrade status`

Plugin is developed using `kubectl-openebs resource verb` pattern

plugin/src/cli_utils/upgrade/mod.rs
tiagolobocastro: Then should be ok I guess

plugin/src/main.rs
tiagolobocastro: > Plugin is developed using kubectl-openebs resource verb pattern

On the mayastor yes, but now on openebs, we have the engine as the first command right?

So it might be ok to keep upgrade as well? 
EDIT: to clarify, I mean using upgrade as top level, as Abhilash showed above

plugin/src/cli_utils/upgrade/cli.rs
```diff
@@ -0,0 +1,70 @@
...
+pub struct UpgradeArgs {
+    #[arg(skip)]
+    pub namespace: String,
+
+    #[arg(skip)]
+    pub kubeconfig: Option<PathBuf>,
+
+    /// Helm release name for the openebs helm chart.
+    #[arg(global = true, long, short)]
+    pub release_name: Option<String>,
```
tiagolobocastro, line 19: what's the different between namespace and release-name?

plugin/src/cli_utils/upgrade/cli.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    /// Specify the container registry for the upgrade-job image.
+    #[arg(global = true, long)]
```
tiagolobocastro: Why are these global?

plugin/src/cli_utils/upgrade/mod.rs
tiagolobocastro: I mean, you have added the cli args:
```
#[arg(env = "HELM_DRIVER", default_value = "")]
    pub helm_storage_driver: String,
```
But here you're only using the env var, so it's kind pointless to add it to the cli args?

plugin/src/cli_utils/upgrade/mod.rs
tiagolobocastro: Actually it may not be, ex: what if this content is from a previously failed upgrade?

plugin/src/cli_utils/upgrade/mod.rs
niladrih: Makes sense. It doesn't need to be on the cli args.

plugin/src/cli_utils/upgrade/mod.rs
niladrih: There are delete command to delete the resources. And it'll be printed that it already exists, which implies that creation was skipped?

plugin/src/cli_utils/upgrade/cli.rs
niladrih: Kubernetes namespace and helm release name. Example: release 'prod_openebs' in namespace 'storage'

plugin/src/main.rs
niladrih: I've changed the `kubectl openebs get upgrade-status` to `kubectl openebs upgrade status`. No change to `kubectl openebs upgrade`.

abhilashshetty04 (approved):

plugin/src/cli_utils/upgrade/mod.rs
niladrih: I have removed the additional env read, and have the CLI arg input everywhere.

plugin/src/cli_utils/upgrade/cli.rs
niladrih: Removed the global attributes

PR #3911, opened 2025-04-28
chore: add loki and alloy to umbrella chart

Abhinandan-Purkait:
### Changes

1. Adds loki to singleBinary monolithic chart as new dependency.
2. Adds alloy as new dependency.
3. Disables mayastor loki and alloy dependency.
4. Adds configuration for alloy.
   1. We can now defined multiple label selectors for scraping.
   2. We can now define tenant id, to segregate logs, based on tenant.
5. Updates the storage configs for the newer components.


charts/templates/loki-storage/storage/localpv-loki.yaml
tiagolobocastro, line 1: Do we really need this?
Even in mayastor we have moved away from "manual" for a long time now

charts/values.yaml
```diff
@@ -64,3 +79,277 @@ engines:
...
+  minio:
+    replicas: 3
+    drivesPerNode: 1
+    mode: distributed
+    # Disable this if you want to enabled external s3 bucket, and uncomment the storage section above.
+    enabled: true
+    persistence:
+      enabled: true
+      storageClass: "openebs-minio-localpv"
+      size: 2Gi
```
tiagolobocastro, line 155: I guess this should take a Comment with appropriate prefix so it gets documented on the README?

charts/values.yaml
```diff
@@ -64,3 +79,277 @@ engines:
...
+          store: tsdb
+          object_store: s3
+          schema: v13
+          index:
+            prefix: loki_index_
+            period: 24h
+    commonConfig:
+      replication_factor: 3
+
+    # Configure these if you don't want to use minio and rather an external s3 bucket.
```
tiagolobocastro: Can we point to documentation url instead?

charts/values.yaml
```diff
@@ -64,3 +79,277 @@ engines:
...
+      # max_chunk_age: 6m
+    tracing:
+      enabled: true
+    querier:
+      max_concurrent: 1
+    limits_config:
+      ingestion_burst_size_mb: 1000
+      ingestion_rate_mb: 10000
+
+  singleBinary:
```
tiagolobocastro, line 130: What's this single binary?

charts/values.yaml
```diff
@@ -64,3 +79,277 @@ engines:
...
+    # Disable this if you want to enabled external s3 bucket, and uncomment the storage section above.
+    enabled: true
+    persistence:
+      enabled: true
+      storageClass: "openebs-minio-localpv"
+      size: 2Gi
+
+  deploymentMode: SingleBinary
+
+  lokiCanary:
```
tiagolobocastro, line 159: Do we need to explicitly disable all this? It's quite verbose

charts/values.yaml
```diff
@@ -64,3 +79,277 @@ engines:
...
+  bloomCompactor:
+    replicas: 0
+  bloomGateway:
+    replicas: 0
+
+alloy:
+  enabled: true
+
+  logging_config:
+    # -- Enable debugging on alloy components.
```
tiagolobocastro: Do we need this documented?

charts/values.yaml
```diff
@@ -64,3 +79,277 @@ engines:
...
+        {{- if eq $j $i -}}
+        {{- $part = append $part (get $labels $key) -}}
+        {{- else -}}
+        {{- $part = append $part ".*" -}}
+        {{- end -}}
+        {{- end -}}
+        {{- $regexParts = append $regexParts (join ";" $part) -}}
+        {{- end -}}
+        {{- join "|" $regexParts -}}
+        {{- end -}}
```
tiagolobocastro: missing newline?

charts/values.yaml
```diff
@@ -8,6 +8,8 @@ openebs-crds:
 localpv-provisioner:
   rbac:
     create: true
+  analytics:
+    enabled: true
```
tiagolobocastro: What is this change doing, was it not enabled?

charts/templates/loki-storage/storage/localpv-loki.yaml
Abhinandan-Purkait: Yeah, it would be useful if someone wants to use their own default storage solution rather than hostpath.

charts/values.yaml
Abhinandan-Purkait: I initially thought would do that, but the documentation is not very clear and the options are segregated and not in one place.

charts/values.yaml
Abhinandan-Purkait: It's the mode of loki that we are using for this chart.

charts/values.yaml
Abhinandan-Purkait: Yes we need to other wise it will install a lot of components which we don't need.

charts/values.yaml
Abhinandan-Purkait: Documented as in? This is the scrape configuration. It's mostly same as that of the promtail, but with few extra tweaks, like using multiple logging labels etc

charts/values.yaml
Abhinandan-Purkait: Removed

charts/values.yaml
Abhinandan-Purkait: Done.

charts/templates/loki-storage/storage/localpv-loki.yaml
tiagolobocastro: Can't they just set the storage class name?

charts/values.yaml
tiagolobocastro: Maybe we can put this in a doc and then just point to it

charts/values.yaml
tiagolobocastro: fair enough then :(

charts/values.yaml
tiagolobocastro: I mean, do we really need this documented on the README?

```suggestion
    # Enable debugging on alloy components.
```

charts/values.yaml
tiagolobocastro: I mean, we can keep, was just questioning if this was a change because if it was then would need more documentation.

tiagolobocastro (approved):

charts/templates/loki-storage/storage/localpv-loki.yaml
Abhinandan-Purkait: Yes, they can, this is more like for ease of deploying, i.e don't need to know the storage class just use manual and it automatically picks up the default storage class.

charts/values.yaml
Abhinandan-Purkait: Sure. I will add a document for the storage configurations and point it to this.

charts/values.yaml
Abhinandan-Purkait: Probably not. Removing it.

charts/values.yaml
```diff
@@ -64,3 +79,252 @@ engines:
...
+
+  singleBinary:
+    replicas: 3
+    drivesPerNode: 1
+    persistence:
+      enabled: true
+      storageClass: "openebs-loki-localpv"
+      accessModes:
+        - ReadWriteOnce
+      size: 2Gi
```
dsharma-dc, line 141: Is this the limit of pvc size loki can use for logs? Should this be configurable?

charts/values.yaml
```diff
@@ -64,3 +79,252 @@ engines:
...
+
+  minio:
+    replicas: 3
+    drivesPerNode: 1
+    mode: distributed
+    enabled: true
+    persistence:
+      enabled: true
+      storageClass: "openebs-minio-localpv"
+      size: 2Gi
```
dsharma-dc, line 155: Similarly this size.

charts/values.yaml
Abhinandan-Purkait: Done.

dsharma-dc (approved):

charts/templates/loki-storage/storage/localpv-storageclass.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - name: StorageType
+        value: "hostpath"
+      - name: BasePath
+        value: {{ tpl ( .Values.loki.localpvScConfig.minio.basePath | quote ) . }}
+    openebs.io/cas-type: local
+  name: {{ tpl .Values.loki.localpvScConfig.minio.name . | required "StorageClass name for loki localpv storage cannot be empty" }}
+provisioner: openebs.io/local
+reclaimPolicy: {{ .Values.loki.localpvScConfig.minio.reclaimPolicy }}
+volumeBindingMode: {{ .Values.loki.localpvScConfig.minio.volumeBindingMode }}
+{{- end }}
```
niladrih: +Newline

niladrih (approved):

charts/values.yaml
Abhinandan-Purkait: It is configurable, its just the default value.

charts/values.yaml
Abhinandan-Purkait: Yes, it's configurable.

charts/values.yaml
tiagolobocastro: Do we have any idea as to how much storage the singleBinary will consume?
Also worth noting that with the hostpath this will effectively be unbounded anyway right?

charts/loki-storage.md
```diff
@@ -0,0 +1,90 @@
...
+
+---
+
+### 2. **High Availability (HA) Deployment (Object Storage)**
+
+For **HA (High Availability)** deployments, **object storage** is **mandatory**. This ensures that your logs are replicated and highly available across multiple nodes, reducing the risk of data loss and providing fault tolerance.
+
+- **Recommended for**: Production environments with multiple Loki instances, where high availability and horizontal scaling are required.
+
+By default with openebs loki is deployed in this mode, i.e with multiple replicas and with an minio distributed object storage.
```
tiagolobocastro: a ? not sure

```suggestion
By default with openebs loki is deployed in this mode, i.e with multiple replicas and with a minio distributed object storage.
```

charts/templates/loki-storage/storage/localpv-loki.yaml
tiagolobocastro: But this doesn't scrub data, right? I remember there were a few issues with the manual approach
And for ease, you can just set storage class name to default? So same

charts/loki-storage.md
tiagolobocastro, line 1: You might want to add this file to the helm ignore if you don't want it on the helm tar btw

tiagolobocastro (approved):

charts/templates/loki-storage/storage/localpv-loki.yaml
Abhinandan-Purkait: Removed it for now.

charts/loki-storage.md
Abhinandan-Purkait: Done.

PR #3912, opened 2025-04-28
chore: point submodules to develop

Abhinandan-Purkait:
- Updates the extensions submodules to develop.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "mayastor"]
 	path = mayastor
 	url = ../mayastor-extensions.git
-	branch = release/2.8
+	branch = develop
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,8 +152,8 @@ dependencies = [
  "regex",
  "ring",
  "rustls-native-certs 0.7.3",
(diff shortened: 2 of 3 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "mayastor"]
 	path = mayastor
 	url = ../mayastor-extensions.git
-	branch = release/2.8
+	branch = develop
```
tiagolobocastro, line 4: this should now be enforced by the PR check when we're updating the versions

Abhinandan-Purkait: Sorry what do you mean by PR check  when we're updating the versions?

tiagolobocastro: I mean, the workflow that updates the mayastor dependency for the chart, should also ensure the submodule is updated

tiagolobocastro (approved):

.gitmodules
Abhinandan-Purkait: Okay, yeah I think @pchandra19 is working on it. I will merge this for now. Seems like plugin build is not working on extensions.

PR #3914, opened 2025-04-29
Update github runners to oci gh arc runners

koksay:
CNCF has hosted ephemeral GitHub runners on Oracle that we want projects to use instead of the GitHub-hosted ones, which now incur a cost to use.

Please direct any questions to me, @jeefy, @krook, and @RobertKielty. (or #cncf-ci-infra channel on the CNCF Slack)

### DIFF
--- a/.github/workflows/branch-preparation.yml
+++ b/.github/workflows/branch-preparation.yml
@@ -49,7 +49,7 @@ jobs:
 
   prepare-branch:
     needs: create-release-branch
-    runs-on: oracle-8cpu-32gb-x86-64
+    runs-on: oracle-vm-8cpu-32gb-x86-64
     outputs:
       pr_title: ${{ steps.set-title.outputs.title }}
     steps:
--- a/.github/workflows/branch-preparation.yml
+++ b/.github/workflows/branch-preparation.yml
@@ -49,7 +49,7 @@ jobs:
 
   prepare-branch:
     needs: create-release-branch
-    runs-on: ubuntu-latest-8-cores
+    runs-on: oracle-8cpu-32gb-x86-64
     outputs:
       pr_title: ${{ steps.set-title.outputs.title }}
     steps:
--- a/.github/workflows/chart-images.yml
+++ b/.github/workflows/chart-images.yml
@@ -13,7 +13,7 @@ env:
 
 jobs:
   helm-images:
-    runs-on: oracle-8cpu-32gb-x86-64
+    runs-on: oracle-vm-8cpu-32gb-x86-64
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/chart-images.yml
+++ b/.github/workflows/chart-images.yml
@@ -13,7 +13,7 @@ env:
 
 jobs:
   helm-images:
-    runs-on: ubuntu-latest-8-cores
+    runs-on: oracle-8cpu-32gb-x86-64
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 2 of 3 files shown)


tiagolobocastro:
Hey @koksay 
Other repos will also need modifying, ex: https://github.com/openebs/mayastor/blob/develop/.github/workflows/unit-int.yml

Looks like the tests are failing:
```
error: cannot connect to socket at '/nix/var/nix/daemon-socket/socket': No such file or directory
```

Are these runners not supporting docker inside them? IIRC a while back I had spoken with @RobertKielty about support for ARM runners and the runners were containers, not supporting docker in docker.

koksay:
@tiagolobocastro You're right, those runners are containers, but they currently support DIND. It appears that a couple of mounts are missing. I will update them and then ask you to rerun the failed job.

koksay:
@tiagolobocastro can you rerun the helm-images job?

tiagolobocastro:
> @tiagolobocastro can you rerun the helm-images job?

Done, looks like failed with the same error:
https://github.com/openebs/openebs/actions/runs/14727177890/job/41372867335

koksay:
ok thanks, I will check this further

tiagolobocastro:
@koksay is there a timeline when the current runners would be disabled?

I think I also saw that KVM is not enable? We have some tests which use KVM to spawn VM's

If we can't get these new runners working for us, is there an alternative, ex: CNCF VM's or baremetal which we could use as runners?

tiagolobocastro:
Also for the mayastor tests specifically, running as a container might prove tricky as we're connecting nvme devices in the kernel and sometimes testing may lead to a kernel crash, thus potentially affecting other runs.

Probably for those specific ones we should run as VM's regardless of fixing this issue here?

koksay:
Yes, we need VMs for your pipelines. We're working on running those on VMs instead of (or in addition to) containers, but the system is not ready yet. Currently, you're using the ones from GitHub, and the problem is the cost. They will not be disabled, but we need to move away from them asap.

I'm closing this PR as it will not work this way. I will create another one when the VM infrastructure is ready. @jeefy fyi...

koksay:
Hey, I'm reopening this one as we have the option to use VMs as well.

tiagolobocastro:
> Hey, I'm reopening this one as we have the option to use VMs as well.

Awesome! We're going further now:

```
ERROR: failed to create cluster: failed to list nodes: command "docker ps -a --filter label=io.x-k8s.kind.cluster=kind --format '{{.Names}}'" failed with error: exit status 1

Command Output: permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.48/containers/json?all=1&filters=%7B%22label%22%3A%7B%22io.x-k8s.kind.cluster%3Dkind%22%3Atrue%7D%7D": dial unix /var/run/docker.sock: connect: permission denied
```

Is the user not part of docker group perhaps?

koksay:
You're absolutely right. I made some changes. Can you retest that failing check?

Abhinandan-Purkait:
I re ran the job. It seems to have failed again. Here is error.

> ERROR: failed to create cluster: failed to list nodes: command "docker ps -a --filter label=io.x-k8s.kind.cluster=kind --format '{{.Names}}'" failed with error: exit status 1
Command Output: permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.48/containers/json?all=1&filters=%7B%22label%22%3A%7B%22io.x-k8s.kind.cluster%3Dkind%22%3Atrue%7D%7D": dial unix /var/run/docker.sock: connect: permission denied

koksay:
I overlooked some shell behaviour. I tested it and seems fine now, can you trigger it again please?

Abhinandan-Purkait:
Thanks! Seems to have passed now.

tiagolobocastro (approved):

PR #3915, opened 2025-04-30
OEP 3916 - Volume Cloning

adam-lithus:
As per https://github.com/openebs/mayastor/issues/1718

@tiagolobocastro

### DIFF
--- a/designs/replicated-pv/mayastor/volume-clone.md
+++ b/designs/replicated-pv/mayastor/volume-clone.md
@@ -1,0 +1,198 @@
+---
+oep-number: OEP 3905
+title: Volume Cloning Feature for OpenEBS Mayastor
+authors:
+  - "@adamcharnock"
+owners:
+  - "@tiagolobocastro"
+  - "@adamcharnock"
+editor: TBD
+creation-date: 2025-04-30
+last-updated: 2025-04-30
+status: provisional
+---
+
+# Volume Cloning Feature for OpenEBS Mayastor
+
+## Table of Contents
+
+1. [Overview](#overview)
+2. [Motivation](#motivation)
+3. [Goals](#goals)
+4. [Non-Goals](#non-goals)
+5. [Proposal](#proposal)
+6. [User Stories](#user-stories)
+7. [Implementation Details](#implementation-details)
+8. [Testing](#testing)
+
+---
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


adam-lithus:
I didn't have too much time available to write this, but I did the best with what I had in between other tasks. I hope it is ok, and if anything is amiss please let me know.

designs/replicated-pv/mayastor/volume-clone.md
```diff
@@ -0,0 +1,198 @@
+---
+oep-number: OEP 3905
```
tiagolobocastro: Please link here the tracking issue which I've raised: #3916

Oh which github handle should I have used there, adamcharnock or adam-lithus?

```suggestion
oep-number: OEP 3916
```

tiagolobocastro (approved):
Awesome, thank you!

designs/replicated-pv/mayastor/volume-clone.md
adam-lithus: Ah yes, I've updated this (now I understand how the numbering works!). And let's go with `adam-lithus`, as this is work related for me

designs/replicated-pv/mayastor/volume-clone.md
```diff
@@ -0,0 +1,198 @@
...
+This proposal introduces a volume cloning feature for Mayastor. A clone is a copy of
+an existing volume that can be used independently from the source volume. The feature
+will leverage Mayastor's existing snapshot and restore capabilities to efficiently
+implement cloning with copy-on-write (COW) functionality.
+
+Volume cloning will enable users to quickly create duplicates of volumes for various
+use cases such as development, testing, data analysis, and backup, without having to
+copy all the data.
+
+## Motivation
```
michaelbeaumont, line 37: IMO all of this is very noisy and just buries the lede. I'm not sure any of these motivations are particular to cloning, all of these use cases are solved with snapshots AFAICT. We already have CoW with snapshots. The whole point of https://github.com/openebs/mayastor/issues/1718 is very particular to the Kubernetes CSI, where instead of going, _at the API level_, from PVC -> snapshot -> PVC we go directly from PVC -> PVC.

adamcharnock: Thanks for pointing this out @michaelbeaumont. I've updated this section to be more to the point.

niladrih (approved):

PR #3919, opened 2025-05-08
chore(ci): add an extra check for release branch before blocking publish

pchandra19:


.github/workflows/chart-release.yml
```diff
@@ -23,7 +23,8 @@ jobs:
         id: check_version
         run: |
           VERSION=$(yq '.version' charts/Chart.yaml)
-          if [[ "$VERSION" == *-develop* ]]; then
+          BRANCH="${GITHUB_REF#refs/heads/}"
+          if [[ "$BRANCH" == release/* && "$VERSION" == *-develop* ]]; then
```
Abhinandan-Purkait, line 27: echo "Chart version is develop and not prerelease: $VERSION"

Abhinandan-Purkait (approved):

.github/workflows/chart-release.yml
pchandra19: I believe anything with a - in semVer is prerelease, hence called it a develop prerelease.

PR #3927, opened 2025-05-14
chore(ci): add workflow to push upgrade image to dockerhub

pchandra19:


.github/workflows/images.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push the release images
+        run: |
+          TAG=$(nix-shell --run "cat charts/Chart.yaml | yq .appVersion")
+          nix-shell --run "./scripts/release.sh --image upgrade.job --tag v${TAG}"
```
niladrih: This is okay for regular pushes to the branch, but for tag pushes, the tag should be the pushed tag.

pchandra19: Yes ! I missed that. Let me update the workflow.

niladrih (approved):

PR #3928, opened 2025-05-15
feat(supportability): add system dump operation to plugin

Abhinandan-Purkait:
- Integrates the mayastor-extensions supportability with umbrella plugin.
- Adds the dump for local engines


plugin/src/cli_utils/localpv/lvm/mod.rs
```diff
@@ -1,12 +1,14 @@
...
 use plugin::resources::utils::OutputFormat;
+use plugin::ExecuteOperation;
 
+use clap::Parser;
 use snafu::Snafu;
 use std::path::PathBuf;
 
+pub(crate) mod node;
+pub(crate) mod supportability;
+pub(crate) mod volume;
```
niladrih, line 10: Should these be pub, looking beyond this crate to possible derivations built from this library?

plugin/src/cli_utils/localpv/lvm/supportability.rs
niladrih, line 1: Functions on this file don't have doc comments. While they're not exposed to the public scope, it's good to know what they do.

plugin/src/cli_utils/supportability/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+    #[clap(global = true, short, long)]
+    etcd_endpoint: Option<String>,
+
+    /// Output directory path to store archive file
+    #[clap(global = true, long, short = 'd', default_value = "./")]
+    output_directory_path: String,
+
+    /// Kubernetes namespace of openebs service
+    #[clap(global = true, long, short = 'n', default_value = "openebs")]
+    namespace: String,
```
niladrih: We're not assuming defaults for the namespace typically. The namespace is passed down from the caller functions, it doesn't need to be a clap arg.

plugin/src/cli_utils/supportability/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Path to kubeconfig file.
+    #[clap(skip)]
+    kubeconfig: Option<PathBuf>,
```
niladrih, line 38: This makes sense, the kubeconfig is passed down from the caller functions.

dsharma-dc (approved):

plugin/src/cli_utils/localpv/lvm/mod.rs
Abhinandan-Purkait: No they don't need to be pub, atleast at the moment. There is no other consumer of this at the moment.

plugin/src/cli_utils/localpv/lvm/supportability.rs
Abhinandan-Purkait: Sure, I will add them. Thanks

plugin/src/cli_utils/supportability/mod.rs
Abhinandan-Purkait: This has been the behaviour of the system dump all these while for mayastor so kept the same behaviour here. This is just to add a layer on convenience as mostly users would install it in openebs ns, so they don't need to pass -n openebs all the time.

niladrih: The plugin behaviour is to use the default namespace. The openebs plugin doesn't have this behaviour of assuming the namespace of the deployment. This is only exclusive to the mayastor plugin.
cc: @tiagolobocastro @abhilashshetty04

Abhinandan-Purkait: Why did we decide to change the behavior for openebs then? Should it be same for all plugins, even the previous openebs ctl used to work this way.

Abhinandan-Purkait: It's very unlikely that someone will install it in default namespace.

tiagolobocastro: No, let's not do this, we had already decided on this for mayastor v3 and openebs (since it's a new plugin, so no breaking changes), namespace is a required parameter

Abhinandan-Purkait: Anyway I have kept the behaviour same as the current plugin. Have reverted this. PTAL

tiagolobocastro: I mean, so should have no default and user must specify

Abhinandan-Purkait: Done.

niladrih (approved):

PR #3947, opened 2025-06-04
chore(ci): fixes for upgrade job, publish-charts and mktemp issue

pchandra19:


.github/workflows/images.yml
```diff
@@ -63,6 +63,7 @@ jobs:
             # Remove 'refs/tags/' prefix to get the tag name
             TAG="${GITHUB_REF#refs/tags/}"
             TAG="${TAG#v}"
+            nix-shell --pure --run "./scripts/helm/update-version-upon-release.sh" ./shell.nix
```
tiagolobocastro: Doesn't need to be specified

```suggestion
            nix-shell --pure --run "./scripts/helm/update-version-upon-release.sh"
```

tiagolobocastro (approved):

.github/workflows/images.yml
tiagolobocastro: Duplicate nix installs in the yaml above, can remove this from above as well:

```yaml

      - uses: cachix/install-nix-action@v31.3.0
        with:
          github_access_token: ''

      - name: Inject github token to nix config
        run: |
          cat <<EOF | sudo tee -a /etc/nix/nix.conf
          access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
          trusted-users = root runner
          EOF

      - name: Restart Nix daemon
        run: sudo systemctl restart nix-daemon
```

PR #3954, opened 2025-06-09
include tag while building plugin

abhilashshetty04:


.github/workflows/release-binaries.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         with:
           enable_kvm: false
       - run: |
-          nix-build -A utils.release.${{ matrix.arch }}.${{ matrix.target }}.kubectl-openebs --arg incremental false
+          nix-build -A utils.release.${{ matrix.arch }}.${{ matrix.target }}.kubectl-openebs --arg incremental false --tag "${{ github.event.release.tag_name }}"
```
tiagolobocastro: there's no release event here, only tag or branch push, so this doesn't seem right?
Also should only use it when it's a tag push... this workflow is also triggered for branch pushes

.github/workflows/pr.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         if: needs.plugin-changed.outputs.rust_changed == 'true'
         # todo: modify downstream release.sh to build plugin via utils
         #  perhaps it's time to rename utils or convert all to the utils build
-        run: nix-build -A utils.debug.x86_64.linux-musl.kubectl-openebs --arg incremental false
+        run: nix-build -A utils.debug.x86_64.linux-musl.kubectl-openebs --arg incremental false --tag "${{ github.event.release.tag_name }}"
```
tiagolobocastro: I'm confused what we're trying to do here as well, there's no tag here, this is a PR

abhilashshetty04: Here probably not required. Ill remove this. This just validates the build

.github/workflows/release-binaries.yml
abhilashshetty04: Ive modified it to pass tag only if it is tagged push. 

Should we run tar and upload on every commit merge? Probably we can create a separate workflow if that is not needed.

abhilashshetty04: we have a wf that looks for release: https://github.com/openebs/openebs/blob/develop/.github/workflows/publish-release-artifacts.yml

That downloads and pushes artefacts to krew.

tiagolobocastro (approved):
Pls cherry-pick to release/4.3

.github/workflows/release-binaries.yml
tiagolobocastro: > Should we run tar and upload on every commit merge? Probably we can create a separate workflow if that is not needed.

At the moment this is the only way to get access to the binary on a specific commit so anyone can download it

Abhinandan-Purkait (approved):

PR #3962, opened 2025-06-13
Update submodule to point to release/2.9

abhilashshetty04:


Abhinandan-Purkait (approved):

Abhinandan-Purkait:
Looks good but just wondering, once we have mayastor tagged should we point it to the tag before cutting a tag here? cc @tiagolobocastro @pchandra19

tiagolobocastro:
> Looks good but just wondering, once we have mayastor tagged should we point it to the tag before cutting a tag here? cc @tiagolobocastro @pchandra19

Yes we should do that; for now it's probably fine since we're releasing now and there will be no more changes on release/2.9
This can be something the workflow dispatch would when modifying the chart dependency version

tiagolobocastro (approved):

PR #3966, opened 2025-06-13
fix(helm): add imagePullSecrets to pre-upgrade hook (#3892)

### LINKED ISSUE #3892: OpenEBS Helm Chart missing imagePullSecrets in pre-upgrade hook
In every subchart, it is possible to set imagePullSecrets to fetch the images from a private registry. 

However in values -> preUpgradeHook this it is missing and no reference is found in https://github.com/openebs/openebs/blob/main/charts/templates/pre-upgrade-hook.yaml .

I assume this is a problem, as I do not know how to trigger the pre-upgrade hook.

nneram:
- Helm chart pre-upgrade hook job fails in private registries
- `imagePullSecrets` were missing from the job spec
- Users behind private registries can't upgrade without manual patching

- Adds `.Values.preUpgradeHook.imagePullSecrets` to the hook's Job spec
- Ensures hook jobs can pull images when secrets are required

- Deploy OpenEBS Helm chart in a cluster using a private registry
- Set `imagePullSecrets` in `values.yaml`
- Helm upgrade should now complete without image pull error

- None expected
- Applies only to the upgrade hook job

5. Other details:
Fixes https://github.com/openebs/openebs/issues/3892

### DIFF
--- a/charts/templates/pre-upgrade-hook.yaml
+++ b/charts/templates/pre-upgrade-hook.yaml
@@ -92,7 +92,10 @@ spec:
         - "-c"
         args:
         - "(kubectl annotate --overwrite crd volumesnapshots.snapshot.storage.k8s.io volumesnapshotclasses.snapshot.storage.k8s.io volumesnapshotcontents.snapshot.storage.k8s.io helm.sh/resource-policy=keep || true) && (kubectl -n {{ .Release.Namespace }} delete deploy -l openebs.io/component-name=openebs-localpv-provisioner --ignore-not-found)"
-{{- if .Values.preUpgradeHook.tolerations }}
+        {{- if .Values.preUpgradeHook.tolerations }}
         tolerations:
-{{ toYaml .Values.preUpgradeHook.tolerations | indent 10 }}
-{{- end }}
+        {{ toYaml .Values.preUpgradeHook.tolerations | indent 10 }}
+        {{- end }}
+      {{- if .Values.preUpgradeHook.imagePullSecrets }}
+      imagePullSecrets: {{ toYaml .Values.preUpgradeHook.imagePullSecrets | nindent 8 }}
+      {{- end }}
--- a/charts/templates/pre-upgrade-hook.yaml
+++ b/charts/templates/pre-upgrade-hook.yaml
@@ -82,6 +82,10 @@ spec:
         {{- end }}
     spec:
       serviceAccountName: "{{ .Release.Name }}-pre-upgrade-hook"
+      {{- with .Values.preUpgradeHook.tolerations }}
+      tolerations:
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
       restartPolicy: Never
       containers:
       - name: pre-upgrade-job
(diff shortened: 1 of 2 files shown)


charts/values.yaml
```diff
@@ -47,6 +47,14 @@ mayastor:
 
 # -- Configuration options for pre-upgrade helm hook job.
 preUpgradeHook:
+  # -- Node tolerations for server scheduling to nodes with taints
+  ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+  ##
+  tolerations: []
+  # -- Optional array of imagePullSecrets containing private registry credentials
+  ## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+  imagePullSecrets: []
```
tiagolobocastro, line 56: Looks like we have an image key entry below, so we could put it there?
ex: `image.pullSecrets`

nneram: Yeah, we could. But I think the best practice is to put `imagePullSecrets` at the pod level. I found this example:
[https://github.com/cetic/helm-job/blob/master/templates/job.yaml](https://github.com/cetic/helm-job/blob/master/templates/job.yaml)
But if you prefer, I can put it under the `image` section instead.

tiagolobocastro: Yes that makes sense, let's leave it as is then

tiagolobocastro (approved):

charts/templates/pre-upgrade-hook.yaml
```diff
@@ -92,7 +92,10 @@ spec:
...
-{{- if .Values.preUpgradeHook.tolerations }}
+        {{- if .Values.preUpgradeHook.tolerations }}
         tolerations:
-{{ toYaml .Values.preUpgradeHook.tolerations | indent 10 }}
-{{- end }}
+        {{ toYaml .Values.preUpgradeHook.tolerations | indent 10 }}
+        {{- end }}
+      {{- if .Values.preUpgradeHook.imagePullSecrets }}
+      imagePullSecrets: {{ toYaml .Values.preUpgradeHook.imagePullSecrets | nindent 8 }}
+      {{- end }}
```
Abhinandan-Purkait: Add a blank line here.

Abhinandan-Purkait (approved):

niladrih (approved):
LGTM

charts/templates/pre-upgrade-hook.yaml
niladrih: The tolerations templating doesn't work for me, PTAL @nneram.

niladrih (changes requested):

niladrih:
@nneram -- this one works:
```yaml
{{- if .Values.preUpgradeHook.tolerations }}
      tolerations:
{{ toYaml .Values.preUpgradeHook.tolerations | indent 8 }}
{{- end }}
```

charts/templates/pre-upgrade-hook.yaml
nneram: I changed the code to:

```
      {{- with .Values.preUpgradeHook.tolerations }}
      tolerations:
      {{- toYaml . | nindent 6 }}
      {{- end }}
```

I think this formatting is prettier and easier to read since the block starts at the line.

I tested both tolerations and imagePullSecrets using helm template and kubectl apply on the Job only, and it works as expected.

Sorry for missing this in the first version!

niladrih (approved):

niladrih:
@nneram thanks for the change! Could you rebase your branch please?

tiagolobocastro:
> @nneram thanks for the change! Could you rebase your branch please?

I've done it with the github UI

Siyler1 (approved):

PR #3969, opened 2025-06-16
oep: mark pool cordon as implementable

tiagolobocastro:


designs/replicated-pv/mayastor/pool-cordon.md
```diff
@@ -106,19 +112,67 @@ As such we're proposing that the CR reflects only the cordon state of the pool,
...
 #### Kubectl Plugin
 
-TODO
+Add new subcommands to cordon with the pool resource.
 
 ### Test Plan
 
-TODO
+In order to test this feature we need a scenario where we'd make use the pool which we are cordoning.
+If there are multiple options available, what if the scheduler always picks the other pools and we never exercise the cordoning logic for example.
```
Abhinandan-Purkait, line 230: What does this mean?

Abhinandan-Purkait (approved):

designs/replicated-pv/mayastor/pool-cordon.md
tiagolobocastro: We need to make sure the tests ensure the logic is working;
ex: if the scheduling never prefers the pool for some reason (ignoring the cordoning), then we're not testing the cordon filtering logic.

Abhinandan-Purkait: I see, in that case we probably need to have exactly identical pools so that the other factors are ruled out? Not sure how simple it can be.

tiagolobocastro: Yeah, maybe add this pool in particular, and add others when this one fails to schedule.

niladrih (approved):
Makes sense to me

avishnu (approved):

PR #3970, opened 2025-06-17
Cherry-pick PR 3967 to release/4.3

niladrih:
The helm secret structures carry the default values from the chart and user configured values. We were previously reading the chart.values structure to check if mayastor is enabled, however the config structure carries user configuration and must also be considered.

### DIFF
--- a/plugin/src/cli_utils/upgrade/helm_values.rs
+++ b/plugin/src/cli_utils/upgrade/helm_values.rs
@@ -9,6 +9,7 @@ use serde::Deserialize;
 #[derive(Deserialize, Debug)]
 pub struct HelmRelease {
     chart: HelmChart,
+    config: Option<HelmConfigValues>,
 }
 
 #[derive(Deserialize, Debug)]
@@ -29,7 +30,6 @@ struct HelmValues {
 
 #[derive(Deserialize, Debug)]
 struct Engines {
-    // local: LocalEngines,
     replicated: ReplicatedEngines,
 }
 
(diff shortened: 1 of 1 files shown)


plugin/src/cli_utils/upgrade/helm_values.rs
```diff
@@ -43,6 +43,26 @@ struct EngineEnabled {
...
+}
+
+#[derive(Deserialize, Debug)]
+struct HelmConfigValuesEngines {
+    replicated: Option<HelmConfigValuesEnginesReplicated>,
+}
+
+#[derive(Deserialize, Debug)]
+struct HelmConfigValuesEnginesReplicated {
+    mayastor: Option<HelmConfigEngineEnabled>,
```
Abhinandan-Purkait, line 58: Why not directly Option<bool> here? As the HelmConfigEngineEnabled is specifically talking about the enablement so I don't think it will have more fields in future?

Abhinandan-Purkait (approved):

plugin/src/cli_utils/upgrade/helm_values.rs
niladrih: The whole thing may be missing from the secret. These things show up only if someone adds configuration explicitly.

niladrih: Also, there's no guarantee we'll not have more fields in the future, there's no reason to make this assumption?

PR #3992, opened 2025-07-24
ci: add staging ci and modify release ci

Abhinandan-Purkait:
- Adds a staging CI that is supposed to be run before tagging.
- Adds scripts to mirror images from one registry to another.
- Adds scripts to pull and push kubectl plugin binary from ghcr to releases.
- Modifies release to only pull and release via staging.
- Adds oras and crane to shell


.github/workflows/staging.yml
```diff
@@ -0,0 +1,159 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push the release images
+        run: |
+          nix-shell --pure --run "./scripts/helm/update-version-upon-release.sh"
+          nix-shell --run "./scripts/release.sh --tag ${TAG} --registry ghcr.io/${{ env.IMAGE_ORG }}/dev"
+
+  kubectl-plugin:
```
tiagolobocastro: let's not duplicate both the image build and the kubectl plugin build
Ex: we could either dispatch the other workflow and collect its result or just dispatch it and do the work there

.github/workflows/release.yml
```diff
@@ -22,14 +22,121 @@ jobs:
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Mirror images from dev to Docker Hub
+        run: |
+          nix-shell --pure --run "./scripts/staging/mirror-images.sh \
+            --source-namespace ghcr.io/${{ env.IMAGE_ORG }}/dev \
+            --target-namespace docker.io/${{ secrets.DOCKERHUB_USERNAME }} \
```
tiagolobocastro: Shouldn't it be this?

```suggestion
            --target-namespace docker.io/${{ secrets.IMAGE_ORG }} \
```

.github/workflows/release.yml
```diff
@@ -22,14 +22,121 @@ jobs:
...
+        with:
+          fetch-depth: 0
+          submodules: recursive
+
+      - uses: cachix/install-nix-action@v31.3.0
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
```
tiagolobocastro: Here and above, should we use another shell.nix? We don't need the rust compiler and all here

.github/workflows/release.yml
```diff
@@ -22,14 +22,121 @@ jobs:
...
+          echo "IMAGE_ORG=${IMAGE_ORG}" >> $GITHUB_ENV
+
+      - name: Set TAG from github ref name
+        run: |
+          TAG="${{ github.ref_name }}"
+          echo "TAG=${TAG}" >> $GITHUB_ENV
+
+      - name: Download and Upload kubectl artifacts
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
tiagolobocastro:
```suggestion
      - name: Download and Upload kubectl artifacts
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          TAG: ${{ github.ref_name }}
```

.github/workflows/release.yml
```diff
@@ -22,14 +22,121 @@ jobs:
...
+        run: |
+          [ -z "${{ secrets.IMAGE_ORG }}" ] && IMAGE_ORG=openebs || IMAGE_ORG=${{ secrets.IMAGE_ORG }}
+          echo "IMAGE_ORG=${IMAGE_ORG}" >> $GITHUB_ENV
+
+      - name: Set TAG from github ref name
+        run: |
+          TAG="${{ github.ref_name }}"
+          echo "TAG=${TAG}" >> $GITHUB_ENV
+
+      - name: Download and Upload kubectl artifacts
```
tiagolobocastro: nit: 
```suggestion
      - name: Mirror kubectl artifacts
```

.github/workflows/release.yml
```diff
@@ -22,14 +22,121 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          nix-shell --pure --run "./scripts/staging/pull-kubectl-from-ghcr.sh \
+            --tag ${TAG} \
+            --namespace ghcr.io/${{ env.IMAGE_ORG }}/dev/kubectl-openebs \
```
tiagolobocastro: What's the difference between env.IMAGE_ORG and github.repository_owner?

.github/workflows/release.yml
```diff
@@ -22,14 +22,121 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: Set Image Org
+        run: |
+          [ -z "${{ secrets.IMAGE_ORG }}" ] && IMAGE_ORG=openebs || IMAGE_ORG=${{ secrets.IMAGE_ORG }}
+          echo "IMAGE_ORG=${IMAGE_ORG}" >> $GITHUB_ENV
```
tiagolobocastro: Can't we just use `github.repository_owner` ?

.github/workflows/release.yml
```diff
@@ -22,14 +22,121 @@ jobs:
...
+            --version "$TAG" \
+            --destination charts-staged
+
+          tar -xzf charts-staged/openebs-"$TAG".tgz -C charts-final
+
+          mv charts-final/openebs/* charts
 
       - name: Publish openebs umbrella helm chart
         uses: stefanprodan/helm-gh-pages@v1.7.0
-        env:
```
tiagolobocastro, line 31: Is this not needed?

scripts/staging/mirror-images.sh
```diff
@@ -0,0 +1,55 @@
...
+            ;;
+    esac
+done
+
+if [[ -z "$SOURCE_NAMESPACE" ]] || [[ -z "$TARGET_NAMESPACE" ]] || [[ -z "$TAG" ]]; then
+    log_fatal "Usage: $0 --source-namespace <source> --target-namespace <target> --tag <tag>"
+fi
+
+IMAGES=(
+    "openebs-upgrade-job"
```
tiagolobocastro: Can we fetch this from the release.sh?

scripts/staging/mirror-images.sh
```diff
@@ -0,0 +1,55 @@
...
+ROOT_DIR="$SCRIPT_DIR/../.."
+
+source "$ROOT_DIR/mayastor/scripts/utils/log.sh"
+
+SOURCE_NAMESPACE=""
+TARGET_NAMESPACE=""
+TAG=""
+
+while [[ $# -gt 0 ]]; do
+    case $1 in
```
tiagolobocastro: Use 2 spaces to keep consistent with the other scripts in the repo

scripts/staging/pull-kubectl-from-ghcr.sh
```diff
@@ -0,0 +1,67 @@
...
+    "aarch64-apple-darwin"
+    "x86_64-windows-gnu"
+)
+
+mkdir -p artifacts
+
+for PLATFORM in "${PLATFORMS[@]}"; do
+    echo "Pulling kubectl-openebs for ${PLATFORM}..."
+
+    oras pull "${NAMESPACE}:${VERSION}-${PLATFORM}"
```
tiagolobocastro: Isn't it a single tar file now?

.github/workflows/images.yml
```diff
@@ -4,8 +4,6 @@ on:
     branches:
```
pchandra19, line 4: Will we be pushing images to DockerHub in case of release/develop ? Shouldn't this be GHCR instead ?

.github/workflows/release.yml
Abhinandan-Purkait: `github.repository_owner` will always be openebs, while IMAGE_ORG can be anything that we set, it comes from secret, so incase we want to modify it. But for the time being I would make every `github.repository_owner` as things stand today for simplicity

.github/workflows/release.yml
Abhinandan-Purkait: Yes we can as I mentioned above.

.github/workflows/release.yml
Abhinandan-Purkait: No, works fine without that, infact none of the locals use that so far. You want me to keep it?

scripts/staging/mirror-images.sh
Abhinandan-Purkait: I don't see the full name anywhere there?

.github/workflows/images.yml
Abhinandan-Purkait: No, as decided we would not keep these dev and prerel on the staging, correct me if I am wrong @tiagolobocastro

.github/workflows/staging.yml
Abhinandan-Purkait: Makes sense. Making that change. I would keep the chart release as it is as we are doing a lot of different things on the staging.

Abhinandan-Purkait: On release.yml I can probably reuse the chart-release.yml but need to check.

.github/workflows/release.yml
Abhinandan-Purkait: I am planning on removing the IMAGE_ORG thing altogether and use the `github.repository_owner` so it will be that now.

.github/workflows/release.yml
Abhinandan-Purkait: Indeed. Let me add that.

.github/workflows/release.yml
tiagolobocastro: Alright, if it works then it's fine to remove it

scripts/staging/mirror-images.sh
tiagolobocastro: hmm some code might be needed to fetch it from the derivation then?
or some other way to make sure this is kept up to date

scripts/staging/pull-kubectl-from-ghcr.sh
Abhinandan-Purkait: Yeah, done.

.github/workflows/staging.yml
Abhinandan-Purkait: Done. I will leave the chart bit as it is, as it's not straightforward as such.

scripts/staging/mirror-images.sh
Abhinandan-Purkait: Yeah, not sure. Left it as it is for the time being.

.github/workflows/images.yml
tiagolobocastro: Since these are floating images there's no need to have a staging for these, it's fine to push do dockerhub

.github/workflows/images.yml
```diff
@@ -52,19 +55,27 @@ jobs:
...
         run: |
-          if [[ "${{ github.ref_type }}" == "tag" ]]; then
-            # Remove 'refs/tags/' prefix to get the tag name
-            TAG="${GITHUB_REF#refs/tags/}"
-            TAG="${TAG#v}"
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            TAG="${{ inputs.tag }}"
             nix-shell --pure --run "./scripts/helm/update-version-upon-release.sh"
-          else
+            nix-shell --run "./scripts/release.sh --tag ${TAG} --registry ${{ inputs.namespace }}"
```
tiagolobocastro: nit: you're already using inputs for the namespace
```suggestion
            nix-shell --run "./scripts/release.sh --tag ${{ inputs.tag }} --registry ${{ inputs.namespace }}"
```

.github/workflows/publish-to-krew-index.yml
```diff
@@ -0,0 +1,20 @@
+name: Publish To Krew Index
+on:
+  release:
+    types: [ published ]
+jobs:
+  publish-artifacts:
+    if: ${{ !startsWith(github.event.release.tag_name, 'v4.2') }}
```
tiagolobocastro: I guess we don't need this now

```suggestion
```

.github/workflows/release-binaries.yml
```diff
@@ -50,27 +58,28 @@ jobs:
...
-            nix-build -A utils.release.${{ matrix.arch }}.${{ matrix.target }}.kubectl-openebs --arg incremental false --argstr tag "${{ github.ref_name }}"
-          else
-            nix-build -A utils.release.${{ matrix.arch }}.${{ matrix.target }}.kubectl-openebs --arg incremental false
+          TAG_ARG=""
+          SYSTEM_ARG=""
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            TAG_ARG="--argstr tag ${{ inputs.tag }}"
           fi
+          if [[ "${{ matrix.os }}" == "macos-14" && "${{ matrix.arch }}" == "x86_64" ]]; then
+            SYSTEM_ARG="--argstr system x86_64-darwin"
```
tiagolobocastro: this should be an optional part of the matrix instead, like the suffix

.github/workflows/release.yml
```diff
@@ -1,17 +1,53 @@
-name: Publish chart upon release creation
+name: Release Tagging Workflow
 
 on:
-  release:
-    types:
-      - "published"
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+**'
```
tiagolobocastro: hmm could we just use the release type now rather than the tag?

.github/workflows/staging.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      - uses: cachix/install-nix-action@v31.3.0
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./scripts/staging/shell.nix
+
+      - name: Download kubectl artifacts
+        uses: actions/download-artifact@v4
```
tiagolobocastro, line 50: where is this downloading from?

.github/workflows/staging.yml
```diff
@@ -0,0 +1,104 @@
...
+      - uses: cachix/install-nix-action@v31.3.0
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./scripts/helm/shell.nix
+
+      - name: Update version for staging
+        run: |
+          nix-shell --pure --run "./scripts/helm/update-version-upon-release.sh" ./scripts/helm/shell.nix
```
tiagolobocastro, line 87: Is this the removal of the `-prerelease`?

.gitignore
```diff
@@ -29,3 +29,6 @@ plugin/target/
 .rust-toolchain
 /rust-toolchain.toml
 /kubectl-openebs
+
+# artifacts
+artifacts/
```
tiagolobocastro: only at the root right?

```suggestion
/artifacts/
```

scripts/staging/mirror-images.sh
```diff
@@ -0,0 +1,55 @@
...
+      ;;
+  esac
+done
+
+if [[ -z "$SOURCE_NAMESPACE" ]] || [[ -z "$TARGET_NAMESPACE" ]] || [[ -z "$TAG" ]]; then
+  log_fatal "Usage: $0 --source-namespace <source> --target-namespace <target> --tag <tag>"
+fi
+
+IMAGES=(
+  "openebs-upgrade-job"
```
tiagolobocastro: You can do something like this:
Reuse the $IMAGES from release.sh and then you just need to eval the imageName, like so:

```
for name in $IMAGES; do
    echo $(nix eval -f . "images.release.$name.imageName" --raw --quiet))
done
```

scripts/staging/pull-kubectl-from-ghcr.sh
```diff
@@ -0,0 +1,77 @@
...
+REGISTRY_HOST=$(echo "$NAMESPACE" | cut -d'/' -f1)
+
+echo "Logging in to ${REGISTRY_HOST}..."
+echo "${PASSWORD}" | oras login "${REGISTRY_HOST}" --username "${USERNAME}" --password-stdin
+
+echo "Pulling kubectl bundle..."
+
+oras pull "${NAMESPACE}:${TAG}"
+
+bundle_tar=$(find . -name "kubectl-openebs-all-platforms-*.tar.gz" -type f | head -1)
```
tiagolobocastro: Shouldn't we simply check for the actual tar file?

`if [ -f "kubectl-openebs-all-platforms-$TAG.tar.gz" ];`

Otherwise if we can't find out tag, but we find other tags, then we're assuming success?

scripts/staging/push-kubectl-to-ghcr.sh
tiagolobocastro, line 1: Would it be worth combining these push-pull scripts?

shell.nix
```diff
@@ -25,6 +25,8 @@ mkShell {
     commitlint
     cowsay
     git
+    oras
+    crane
```
tiagolobocastro: Don't add it here, simply import it as you see here for k8sShell and helmShell

.github/workflows/staging.yml
Abhinandan-Purkait: From the workflow right? There's an upload happening in the previous `build-binaries` step.

.github/workflows/staging.yml
Abhinandan-Purkait: Yeah

scripts/staging/pull-kubectl-from-ghcr.sh
Abhinandan-Purkait: Indeed.

scripts/staging/push-kubectl-to-ghcr.sh
Abhinandan-Purkait: I mean we can, but it's already segregated in the staging directory so kept it for easy of reading, otherwise becomes too big.

.github/workflows/staging.yml
tiagolobocastro: Ah so it's downloading from its own workflow?

scripts/staging/push-kubectl-to-ghcr.sh
tiagolobocastro: Aren't the scripts the same up to L55?

tiagolobocastro: I mean it's fine it you want to keep separate, just thought we could merge them

.gitignore
Abhinandan-Purkait: Yes

scripts/staging/mirror-images.sh
Abhinandan-Purkait: Can't source that here right?

scripts/staging/push-kubectl-to-ghcr.sh
Abhinandan-Purkait: Combined them.

scripts/staging/push-pull-kubectl.sh
tiagolobocastro, line 1: nit: `kubectl-oci.sh`, `kubectl.sh` or something like that?

scripts/staging/push-pull-kubectl.sh
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env bash
+# Manage kubectl plugin binaries as GHCR OCI artifacts.
+# Supports both pushing and pulling operations.
+# Usage:
+#   Push: ./kubectl-ghcr-oci.sh push --tag <tag> --namespace <ghcr-path> --username <user> --password <token>
```
tiagolobocastro: Update actual name

scripts/staging/push-pull-kubectl.sh
```diff
@@ -0,0 +1,170 @@
...
+  # Create a combined tarball of all kubectl binaries
+  echo "Creating combined tarball of all kubectl binaries..."
+  local combined_tar="kubectl-openebs-all-platforms-${TAG}.tar.gz"
+
+  tar -czf "${combined_tar}" -C artifacts .
+
+  echo "Pushing combined tarball to ${NAMESPACE}:${TAG}"
+
+  oras push "${NAMESPACE}:${TAG}" \
+    --artifact-type application/vnd.openebs.kubectl.bundle.v1+tar+gzip \
```
tiagolobocastro: I'd suggest to leave this as default

scripts/staging/mirror-images.sh
Abhinandan-Purkait: Used the one as suggested. Although there is a flaw here, if someone runs this on their fork, the origin on gh vm would be <user-id> and the image generated would be `<user-id>/openebs-upgrade-job`, rather than, `openebs/openebs-upgrade-job` so just removing `openebs/` won't work. But anyway, for now works fine.

tiagolobocastro: Then just remove anything before the: `echo ${image##*/}`

scripts/staging/shell.nix
```diff
@@ -0,0 +1,13 @@
+{ pkgs ? import (import ../../nix/sources.nix).nixpkgs {
+    overlays = [ (_: _: { inherit (import ../../nix/sources.nix); }) (import ../../nix/overlay.nix { }) ];
+  }
+}:
+pkgs.mkShell {
+  name = "helm-scripts-shell";
```
tiagolobocastro:
```suggestion
  name = "staging-shell";
```

tiagolobocastro (approved):

scripts/staging/push-pull-kubectl.sh
Abhinandan-Purkait: using kubectl-oci.sh

scripts/staging/push-pull-kubectl.sh
Abhinandan-Purkait: using kubectl-oci.sh

scripts/staging/push-pull-kubectl.sh
Abhinandan-Purkait: yeah, I checked it, it seems to be an actual field, otherwise it shows up as unknown

pchandra19 (approved):

Issue #3893, opened 2025-03-27
CRDs should not be annotated by pre-upgrade hook if management is disabled in chart


solidDoWant:
## Description

The Helm chart pre-upgrade hook always annotates the volume snapshot controller CRDs even if management/ownership is disabled in the chart.

## Expected Behavior

The hook should not modify the CRDs, and should also not need permission to get or patch them.

## Current Behavior

The hook annotations the volume snapshot CRDs.

## Possible Solution

Only run the [annotate command](https://github.com/openebs/openebs/blob/develop/charts/templates/pre-upgrade-hook.yaml#L91) if management of the CRDs is enabled. Also only conditionally give the service account CRD annotation permissions if management is enabled.

## Steps to Reproduce

1. Configure the chart to not manage these CRDs via `openebs-crds.csi.volumeSnapshots.enabled: false`
2. Deploy the chart
3. Observe the behavior of the pre-upgrade hook.

## Your Environment

N/A - see https://github.com/openebs/openebs/blob/develop/charts/templates/pre-upgrade-hook.yaml#L91

Mrbm5:
I've noticed a similar issue as described, the `pre-upgrade-hook` job always runs during the chart installation, even when it seems unnecessary. Is there a specific reason this job is designed to run every time?  Would it be possible to make its execution optional via values.yaml?

For instance, while using the [Local PV Hostpath](https://github.com/openebs/dynamic-localpv-provisioner), According to the documentation, snapshots are not supported for Local PV Hostpath, and I’ve set the relevant CRD and snapshot-related configurations to false in `values.yaml`. However, the `pre-upgrade-hook` still runs, attempts to annotate CRDs that are not even present, and subsequently deletes the `OpenEBS deployment`. There doesn't seem to be a configurable way to opt out of this behavior.

Here’s is the `values.yaml`, where all snapshot and CRD options are explicitly disabled:
```yaml

openebs-crds:
  csi:
    volumeSnapshots:
      enabled: false
      keep: false

localpv-provisioner:
  rbac:
    create: true
  localpv:
    enabled: true
    basePath: "/var/openebs/local"
    resources:
      limits:
        memory: 256Mi
      requests:
        memory: 32Mi
        cpu: 10m
  helperPod:
  analytics:
    enabled: false
  hostpathClass:
    name: openebs-hostpath
    enabled: true
    reclaimPolicy: Retain
    isDefaultClass: true 
    xfsQuota:
      enabled: true

zfs-localpv:
[...]

PR #3996, opened 2025-07-29
feat(chart): enable preUpgradeHook only for upgrades from v3 to v4

niladrih:
The pre-upgrade-hook for the openebs helm chart runs for every upgrade. This need only run for upgrades from openebs v3 to openebs v4.

This change checks for the localpv-provisioner deployment's `chart:` label for the chart name, and checks to see if the version set on the chart name starts with a `3`, using regex. If not, it doesn't run.

If we don't find an appropriate localpv-provisioner deployment, we run anyway, just to be safe.

Additional changes:
- Moves deletion annotations to values.yaml
- Adds `before-hook-creation` to deletion policy, so that it's not necessary to cleanup failed run resource to run a new run of the pre-upgrade hook
- Adds a toggle for the pre-upgrade job, for convenience.
- Set a check to only perform the localpv-provisioner find and read label logic if it's an upgrade action on helm

### DIFF
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -49,8 +49,6 @@ annotations:
   helm.sh/images: |
     - name: etcd
       image: docker.io/bitnami/etcd:3.6.2-debian-12-r0
-    - name: kubectl
-      image: docker.io/bitnami/kubectl:1.25.15
     - name: alloy
       image: docker.io/grafana/alloy:v1.8.1
     - name: loki
--- a/charts/templates/_helpers.tpl
+++ b/charts/templates/_helpers.tpl
@@ -51,3 +51,35 @@ chart: {{ template "openebs.chart" . }}
 heritage: {{ .Release.Service }}
 openebs.io/version: {{ .Values.release.version | quote }}
 {{- end -}}
+
+{{/*
+Returns matched if the Hostpath Localpv Deployment is of v3.x
+Usage:
+  {{- if include "hostpath_is_v3" . }}
+    Do something
+  {{- end }}
+*/}}
+{{- define "hostpath_is_v3" -}}
+  {{/* Name from https://github.com/openebs/charts/blob/openebs-3.10.0/charts/openebs/templates/localprovisioner/deployment-local-provisioner.yaml#L8 */}}
+  {{- $name1 := printf "%s-localpv-provisioner" (include "openebs.fullname" .) -}}
+  {{/* Name from https://github.com/openebs/dynamic-localpv-provisioner/blob/v4.3.0/deploy/helm/charts/templates/deployment.yaml#L5 */}}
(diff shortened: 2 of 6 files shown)


niladrih:
Related to https://github.com/openebs/openebs/issues/3893

charts/values.yaml
```diff
@@ -62,8 +64,12 @@ preUpgradeHook:
   ## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
   imagePullSecrets: []
   # - name: secretName
-  # -- Labels to be added to the pod hook job
-  podLabels: {}
+  # Labels to be added to the pod hook job
+  podLabels:
+    openebs.io/logging: "true"
+  # This helps in debugging by leaving the Job/Pod instances lying around
```
tiagolobocastro, line 70: You mean by default we'll leave the jobs around? or on failure only?

charts/Chart.yaml
```diff
@@ -49,8 +49,6 @@ annotations:
   helm.sh/images: |
     - name: etcd
       image: docker.io/bitnami/etcd:3.6.2-debian-12-r0
-    - name: kubectl
-      image: docker.io/bitnami/kubectl:1.25.15
```
tiagolobocastro, line 53: Need to modify the helm images scripts to make sure we generate the script, we don't want to miss on this image...

charts/values.yaml
niladrih: We're not leaving the job around in success. In failure, we leave it around.
The comment here suggests that one could debug the pre-hook behaviour better by changing these values here, instead of having to tweak the template.

charts/templates/_helpers.tpl
```diff
@@ -51,3 +51,35 @@ chart: {{ template "openebs.chart" . }}
 heritage: {{ .Release.Service }}
 openebs.io/version: {{ .Values.release.version | quote }}
 {{- end -}}
+
+{{/*
+Returns matched if the Hostpath Localpv Deployment is of v3.x
+Usage:
+  {{- if include "hostpath_is_v3" . }}
+    Do something
```
dsharma-dc, line 59: Are we going to reach in here?

Abhinandan-Purkait (approved):

charts/templates/_helpers.tpl
niladrih: No, that's documentation. The `{{/* */}} blocks are comment blocks.

charts/Chart.yaml
tiagolobocastro: Why was this merged with this bug in?

niladrih: Oh, I'm sorry, I didn't interpret this comment properly. I'll fix it. Sorry about this.

PR #4014, opened 2025-08-04
fix: staging shell, script and change namespaces

Abhinandan-Purkait:
- Change the namespaces for plugin and helm to `/dev/plugin` and `/dev/helm`
- Add packages to staging shell
- Fix the mirror-images script to accept ROOT_DIR while not modifying the current.

### DIFF
--- a/scripts/staging/mirror-images.sh
+++ b/scripts/staging/mirror-images.sh
@@ -6,11 +6,12 @@
 set -euo pipefail
 
 SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]:-"$0"}")")"
+CURRENT_ROOT_DIR="${SCRIPT_DIR}/../.."
 
 # if ROOT_DIR is not defined use the one below
-: "${ROOT_DIR:=${SCRIPT_DIR}/../..}"
+: "${ROOT_DIR:=$CURRENT_ROOT_DIR}"
 
-source "$ROOT_DIR/mayastor/scripts/utils/log.sh"
+source "$CURRENT_ROOT_DIR/mayastor/scripts/utils/log.sh"
 NO_RUN=true . "$ROOT_DIR/scripts/release.sh"
 
 IMAGES=()
--- a/scripts/staging/shell.nix
+++ b/scripts/staging/shell.nix
@@ -9,5 +9,12 @@ pkgs.mkShell {
     crane
     yq-go
     jq
+  ] ++ pkgs.lib.optional (builtins.getEnv "IN_NIX_SHELL" == "pure" && pkgs.system != "aarch64-darwin") [
+    docker
+    git
+    curl
+    nix
+    kubernetes-helm-wrapped
+    cacert
   ];
 }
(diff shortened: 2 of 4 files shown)


scripts/staging/mirror-images.sh
```diff
@@ -6,11 +6,12 @@
 set -euo pipefail
 
 SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]:-"$0"}")")"
+CURRENT_ROOT_DIR="${SCRIPT_DIR}/../.."
```
tiagolobocastro, line 9: If using the term "current", then it should be the other way around?

Abhinandan-Purkait: Could be, should I do that then?

tiagolobocastro: More options, ex: `RELEASE_ROOT_DIR`, `PARENT_ROOT_DIR`
Any good naming suggestions @niladrih ?

Abhinandan-Purkait: `PARENT_ROOT_DIR` makes sense

Abhinandan-Purkait: going with the same

pchandra19 (approved):

niladrih (approved):

scripts/staging/shell.nix
```diff
@@ -9,5 +9,11 @@ pkgs.mkShell {
     crane
     yq-go
     jq
+  ] ++ pkgs.lib.optional (builtins.getEnv "IN_NIX_SHELL" == "pure" && pkgs.system != "aarch64-darwin") [
+    docker
+    git
+    curl
+    nix
+    kubernetes-helm-wrapped
```
tiagolobocastro, line 17: Was the cacert not needed?

tiagolobocastro (approved):

scripts/staging/mirror-images.sh
niladrih: PARENT_ROOT_DIR seems fine

scripts/staging/shell.nix
Abhinandan-Purkait: Ohh we probably need it here as well right.

niladrih: curl used to need cacert in the pure shell

Abhinandan-Purkait: Done

scripts/staging/shell.nix
```diff
@@ -9,5 +9,12 @@ pkgs.mkShell {
     crane
     yq-go
     jq
+  ] ++ pkgs.lib.optional (builtins.getEnv "IN_NIX_SHELL" == "pure" && pkgs.system != "aarch64-darwin") [
```
niladrih, line 12: nit: maybe this needs nix fmt

niladrih (approved):

scripts/staging/shell.nix
Abhinandan-Purkait: Nothing changed

PR #4038, opened 2025-08-20
Kubectl Plugin: snapshot delete

tiagolobocastro:
chore: update chart image metadata
    
---

    chore: use stdenv with no CC,AR in the nix-shell
    
    This reduces the derivation size when not using internal rust and also
    can prevent ring rebuilds due to CC,AR env var mismatch.

---

    chore: add pre-commit trampoline
    
    Allows us to erase the the references which pre-commit now seems
    to leave around and which interfere with nix-shell and venv.
    
---

    feat: add snapshot delete subcommand
    
    Similar to the volume, it first ensures that the volumesnapshotcontent is
    not present! If it is, then it bails out with an error.

### DIFF
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -47,10 +47,6 @@ dependencies:
 
 annotations:
   helm.sh/images: |
-    - name: etcd
-      image: docker.io/bitnamilegacy/etcd:3.6.4-debian-12-r0
-    - name: kubectl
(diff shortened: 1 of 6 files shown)


tiagolobocastro:
Resolves https://github.com/openebs/mayastor/issues/1892

Abhinandan-Purkait (approved):

PR #4043, opened 2025-08-26
oep(4011): Mayastor Diskpool raid0 support

urso:
This OEP discusses the addition of Raid0 support to Mayastor LVS diskpools. This enables users to combine multiple devices into a single pool with aggregate capacity.

### DIFF
--- a/designs/replicated-pv/mayastor/raid0.md
+++ b/designs/replicated-pv/mayastor/raid0.md
@@ -1,6 +1,6 @@
 ---
 oep-number: OEP 4011
-title: Add Raid0 DiskPool support to OpenEBS Mayastore
+title: Add Raid0 DiskPool support to OpenEBS Mayastor
 authors:
   - "@urso"
 owners:
@@ -11,7 +11,7 @@ last-updated: 2025-08-26
 status: provisional
 ---
 
-# Add Raid0 DiskPool support to OpenEBS Mayastore
+# Add Raid0 DiskPool support to OpenEBS Mayastor
 
 ## Table of Contents
 
@@ -49,7 +49,7 @@ Mayastor currently requires separate disk pools per device, resulting in multipl
 
 ### Non-Goals
 
-- RAID0 device replacement maybe future enhancement)
+- RAID0 device replacement maybe future enhancement
 - Migration from existing single-device pools to RAID0
 - Other RAID levels (RAID1, RAID5) - separate OEPs
 
--- a/designs/replicated-pv/mayastor/raid0.md
+++ b/designs/replicated-pv/mayastor/raid0.md
@@ -1,0 +1,145 @@
+---
+oep-number: OEP 4011
+title: Add Raid0 DiskPool support to OpenEBS Mayastore
+authors:
+  - "@urso"
+owners:
+  - "@urso"
+editor: TBD
+creation-date: 2025-08-26
+last-updated: 2025-08-26
+status: provisional
+---
+
+# Add Raid0 DiskPool support to OpenEBS Mayastore
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [User Stories](#user-stories)
+    * [Workflow](#workflow)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    * [Risks and Mitigations](#risks-and-mitigations)
+* [Testing](#testing)
+
+## Summary
+
+This proposal adds RAID0 support to Mayastor disk pools, enabling users to combine multiple devices into single pools with aggregate capacity. The implementation leverages SPDK RAID0 functionality while preserving all existing LVS features including encryption, snapshots, and thin provisioning. This enhancement addresses current limitations where users must create separate pools per device, leading to potential sub-optimal disk space utilization across multiple small pools.
+
+## Motivation
+
+Mayastor currently requires separate disk pools per device, resulting in multiple small pools with potential sub-optimal disk space utilization. RAID0 support enables users to consolidate multiple devices into larger pools with improved capacity utilization.
+
+### Goals
+
+- Enable multi-device RAID0 pools with aggregate capacity from multiple devices
+- Preserve all existing LVS features (encryption, snapshots, thin provisioning) on RAID0 pools
+- Maintain backward compatibility with existing single-device pools
+- Provide foundation for future RAID level support (RAID1, RAID5)
+
+**Success Criteria:**
+- Users can create RAID0 pools from multiple devices
+- RAID0 pools function identically to single-device pools from user perspective
+- No regressions in existing single-device functionality
+
+### Non-Goals
+
+- RAID0 device replacement maybe future enhancement)
+- Migration from existing single-device pools to RAID0
+- Other RAID levels (RAID1, RAID5) - separate OEPs
+
+## Proposal
+
+### User Stories
+
+1. **Story 1**: As a user, I want to create large storage pools (e.g., combining 4x 2TB devices into an 8TB pool) to support applications requiring substantial storage capacity.
+
+### Workflow
+
+#### Single Device Pool (Existing)
+
+1. **Create DiskPool CRD**: Specify single device (no `poolType` needed)
+2. **Apply configuration**: Controller creates pool directly on device
+3. **Pool ready**: Single-device pool available
+
+```yaml
+apiVersion: openebs.io/v1beta3
+kind: DiskPool
+metadata:
+  name: single-pool
+spec:
+  node: node-1
+  disks:
+    - /dev/nvme0n1
+```
+
+#### RAID0 Pool (New)
+
+1. **Prepare devices**: Ensure multiple devices are available
+2. **Create DiskPool CRD**: Specify multiple devices with `poolType: "raid0"`
+3. **Controller validation**: Validates RAID0 requirements (min 2 devices)
+4. **RAID0 creation**: io-engine creates SPDK RAID0 bdev from devices
+5. **Pool creation**: LVS blobstore created on RAID0 bdev
+6. **Pool ready**: RAID0 pool available as single large pool
+
+```yaml
+apiVersion: openebs.io/v1beta3
+kind: DiskPool
+metadata:
+  name: raid0-pool
+spec:
+  node: node-1
+  poolType: "raid0"
+  disks:
+    - /dev/nvme0n1
+    - /dev/nvme1n1
+    - /dev/nvme2n1
+  raid0:                # <- optional
+    stripSize: "64k"
+```
+
+
+### Implementation Details/Notes/Constraints
+
+**Architecture:**
+- Multiple devices → SPDK RAID0 → Encryption (if enabled) → LVS blobstore
+- All existing LVS features work transparently
+- Performance optimized with single encryption layer
+
+**Configuration:**
+- Add optional `poolType` and `raid0` fields to existing DiskPool CRD
+- No CRD version bump needed
+- Backward compatibility maintained
+
+**Core Implementation:**
+- io-engine detects multi-device + `poolType: "raid0"`
+- Uses SPDK `spdk_bdev_raid_create()` API
+- LVS blobstore created on RAID0 bdev (encrypted if specified)
+
+**Key Constraints:**
+- Minimum 2 devices required for RAID0
+- Single device failure fails entire pool
+- No migration from existing pools
+
+### Risks and Mitigations
+
+**Risk**: Any single device failure destroys entire pool (inherent RAID0 behavior)
+
+**Mitigations:**
+1. **Clear Documentation**: Explicitly document RAID0 failure characteristics and appropriate use cases to ensure users understand the trade-offs
+2. **Device Health Monitoring**: Ensure existing device monitoring works properly with RAID0 pools to provide early warning of potential failures
+
+Note: SPDK supports device replacement capabilities that could be added as a future enhancement to provide additional mitigation options.
+
+## Testing
+
+- RAID0 pool creation with multiple devices
+- Configuration validation (minimum devices, strip size)
+- Volume operations work on RAID0 pools
+- Single-device pools continue working unchanged
+- Device failure behavior


designs/replicated-pv/mayastor/raid0.md
```diff
@@ -0,0 +1,145 @@
...
+- Provide foundation for future RAID level support (RAID1, RAID5)
+
+**Success Criteria:**
+- Users can create RAID0 pools from multiple devices
+- RAID0 pools function identically to single-device pools from user perspective
+- No regressions in existing single-device functionality
+
+### Non-Goals
+
+- RAID0 device replacement maybe future enhancement)
```
Abhinandan-Purkait:
```suggestion
- RAID0 device replacement maybe future enhancement
```

designs/replicated-pv/mayastor/raid0.md
```diff
@@ -0,0 +1,145 @@
...
+metadata:
+  name: raid0-pool
+spec:
+  node: node-1
+  poolType: "raid0"
+  disks:
+    - /dev/nvme0n1
+    - /dev/nvme1n1
+    - /dev/nvme2n1
+  raid0:                # <- optional
```
Abhinandan-Purkait, line 102: Should this be an enum, to support future types without having to change anything?

```
enum RaidType {
     Raid0(Raid0Params)
     Raid1(Raid1Params)
     Raid5(Raid5Params)
}

Comment (urso) [reply designs/replicated-pv/mayastor/raid0.md:102]: When converting to PoolArgs or gRPC I actually use an enum like this. I was not sure how idiomatic that would be for the CRD and opted for `poolType` + `raid0` namespace. I think both approaches would be extensible.

Plus the 'raid0' here is optional, as `poolType` determines the type. If `raid0` is missing we will assume the users intends to use default settings.

File: designs/replicated-pv/mayastor/raid0.md [hunk 3]
```diff
@@ -0,0 +1,145 @@
+---
+oep-number: OEP 4011
+title: Add Raid0 DiskPool support to OpenEBS Mayastore
```
tiagolobocastro:
```suggestion
title: Add Raid0 DiskPool support to OpenEBS Mayastor
```

designs/replicated-pv/mayastor/raid0.md
tiagolobocastro: Would this be a bit neater?
```yaml
spec:
  node: node-1
  raid0:
    disks:
      - /dev/nvme0n1
      - /dev/nvme1n1
      - /dev/nvme2n1
    stripeSize: 64k
```

tiagolobocastro: is it too early to start thinking about how conversions would be done via the spec? 
Ex: convert raid0 to raid10

designs/replicated-pv/mayastor/raid0.md
```diff
@@ -0,0 +1,145 @@
...
+    - /dev/nvme2n1
+  raid0:                # <- optional
+    stripSize: "64k"
+```
+
+
+### Implementation Details/Notes/Constraints
+
+**Architecture:**
+- Multiple devices → SPDK RAID0 → Encryption (if enabled) → LVS blobstore
```
tiagolobocastro, line 110: What configures the device ordering?
```
      RAID 0 Configuration
  ┌────────────┐   ┌────────────┐
  │  Disk 1    │   │  Disk 2    │
  │────────────│   │────────────│
  │   A1       │   │   A2       │
  │   B1       │   │   B2       │
  │   C1       │   │   C2       │
  └────────────┘   └────────────┘
```
Does the order in the CR dictate the order on-disk ?
Is there a super-block on the disk which stored this configuration?

designs/replicated-pv/mayastor/raid0.md
```diff
@@ -0,0 +1,145 @@
...
+  - "@urso"
+owners:
+  - "@urso"
+editor: TBD
+creation-date: 2025-08-26
+last-updated: 2025-08-26
+status: provisional
+---
+
+# Add Raid0 DiskPool support to OpenEBS Mayastore
```
tiagolobocastro:
```suggestion
# Add Raid0 DiskPool support to OpenEBS Mayastor
```

designs/replicated-pv/mayastor/raid0.md
urso: We already have a 'disks' setting. The new fields I introduced are fully backwards compatible due to being optional (default single disk config still functions as of today).

assuming we have this:

```
spec:
  node: node-1
  raid0:
    disks:
      - /dev/nvme0n1
      - /dev/nvme1n1
      - /dev/nvme2n1
    stripeSize: 64k
```

would we change the schema for single disk configs to this:

```
spec:
  node: node-1
  disk: /dev/nvme0n1
```

All in all I'm fine with that. My concern was keeping the schema backwards compatible.

> is it too early to start thinking about how conversions would be done via the spec?
Ex: convert raid0 to raid10

For me this was a non-goal. Is this something that would make sense to support?

designs/replicated-pv/mayastor/raid0.md
urso: Well, we use SPDK raid0 implementation for this. I didn't notice the SPDK sources to add a superblock. So yes, the order of disks in the CR matters and can break the raid device when changed between restarts.

urso: Looks like I had checked out the wrong SPDK branch. Superblocks seem to be supported (requires UUID being set, though) and can be used for array reconstructions. IMO it is preferable to write superblocks. The superblock is written to each device in the raid array.

Given that device UUIDs will be required (which is currently optional) I think we have to make this either optional or force users to configured UUIDs.

tiagolobocastro: Yeah I think that would be preferable to me too.
Do you mean a `UUID` for the "raid" bdev? We already have a uuid as optional parameter for pool creation so we could use that. We could for example use the `uid` from the `DiskPool` `CR` itself.

urso: For all devices involved. The raid bdev will need one or the configuration fails.
Unless explicitely configured SPDK creates random UUIDs for devices, which will be a problem between restarts. For aio/iouring we could try to read/derive an UUID by reading the ID or serial number via lsblk. For SPDK PCIe user space driver we also should have some APIs to read IDs. Other device types like memory would require the user to configure the UUID explicitely. In case we have an input device without proper UUID we will have to error out though (or allow user to disable superblocks).

tiagolobocastro: Ah I see...
We've been thinking about putting GPT on the disks to help with identification when using diskpool encryption, would that also help here? And use the part-uuid?

designs/replicated-pv/mayastor/raid0.md
tiagolobocastro: Yeah I see the appeal of keeping back-compat :)
I'm ok with it like that, I'll let others comment their thoughts too @Abhinandan-Purkait @abhilashshetty04 @dsharma-dc

designs/replicated-pv/mayastor/raid0.md
```diff
@@ -0,0 +1,145 @@
...
+metadata:
+  name: raid0-pool
+spec:
+  node: node-1
+  poolType: "raid0"
+  disks:
+    - /dev/nvme0n1
+    - /dev/nvme1n1
+    - /dev/nvme2n1
+  raid0:                # <- optional
```
abhilashshetty04, line 102: User having to specify raid level (raid0) on `pooltype` and then the value of PoolType i.e raid0 becoming key in CR doesnt seem like a good UX.

Can we enclose all RAID specific configs in one key. For example:

```
raidSpec:
    level: raid0  -- > in future concat, raid5, raid5f etc.
    stripeSize: 64k
```
    
 Here, raidSpec is required param if user has more then 1 disk in the CR.

designs/replicated-pv/mayastor/raid0.md
```diff
@@ -0,0 +1,145 @@
...
+  node: node-1
+  disks:
+    - /dev/nvme0n1
+```
+
+#### RAID0 Pool (New)
+
+1. **Prepare devices**: Ensure multiple devices are available
+2. **Create DiskPool CRD**: Specify multiple devices with `poolType: "raid0"`
+3. **Controller validation**: Validates RAID0 requirements (min 2 devices)
```
abhilashshetty04, line 85: Some questions regarding disks:

1. What are the sizing considerations while selecting devices to be part of raid 0 pool. Should they all be equal? 
2. Can user add new device to the raid0 Pool. If yes, What changes it may do internally when new device gets added to raid0 pool. I mean from the existing striped data perspective. Does it try to rebalance?

designs/replicated-pv/mayastor/raid0.md
urso: Maybe let me give backwards compatible example without default raid0 settings as well:

```
apiVersion: openebs.io/v1beta3
kind: DiskPool
metadata:
  name: raid0-pool
spec:
  node: node-1
  poolType: "raid0"
  disks:
    - /dev/nvme0n1
    - /dev/nvme1n1
```

designs/replicated-pv/mayastor/raid0.md
urso: This would work and I think be backwards compatible with the current spec. I think I like it.

designs/replicated-pv/mayastor/raid0.md
urso: If we put a GPT on each disk when we initialize them the first time we could use that information to identify the disks. As the GPT header has an UUID field we would even be fine if the initial UUID is fully random.

My idea for encryption was to have the encryption layer on top of the raid layer. This ensures that we encrypt only once for replicating raid types. Adding GPT to the disks would mean that the raid bdev itself must also have GPT information.

I really like the idea of having superblocks. So my proposal for now would be to attempt to read the device serial number and derive an UUID value from that + ask users to provide an UUID for devices that will not have such information. In addition we can make superblocks an optional feature allowing users to opt-out for the time being.
If/Once GPT becomes available in Mayastor we should migrate to use them for new disks, but unfortunately would need to keep backwards compatibility for existing installations.

designs/replicated-pv/mayastor/raid0.md
urso: This depends on the SPDK raid0 implementation.

> 1. What are the sizing considerations while selecting devices to be part of raid 0 pool. Should they all be equal?

Raid0 enforce uniform sizes. Disks can have different sizes for sure. But actual block count is determined by the disk with the least number of blocks. Blocks of bigger disks will not be used.

> 2. Can user add new device to the raid0 Pool. If yes, What changes it may do internally when new device gets added to raid0 pool. I mean from the existing striped data perspective. Does it try to rebalance?

No, it is not possible to add devices once the raid0 is online. One can potentially grow the underlying devices. The raid0 module processes device resize events and grows its own size in that case I think. I'm not sure if Mayastor/LVS can directly benefit from the raid growing in size or if we need some more changes.

designs/replicated-pv/mayastor/raid0.md
abhilashshetty04: I suggest having all raid specific configurations under 1 optional struct. Otherwise, we would have to extend under spec whenever we add support for new raid type.

Please go though the dsp operator code. We handle the CRD migration there. Also, do you mean rollback when you see backward compatible. If yes, we dont support rollback of CRDs. We introduce newer fields as optional. So they all are CRD migration/upgrade safe.

Also, lets not call it poolType in the root of spec. We have a type in Pool Grpc. That defines if pool is backed by Lvs or Lvm.

So 

 ```
apiVersion: "openebs.io/v1beta3"
kind: DiskPool
metadata:
  name: <pool-name>
  namespace: <namespace>
spec:
  node: <node-name>
  disks: ["/dev/disk/by-id/<id>"]
  raidConfig:
    type: "raid0"
    params: RaidType
```

```
enum RaidType {
     Raid0(Raid0Params)
     Raid1(Raid1Params)
     Raid5(Raid5Params)
}
```

In the above example, raidConfig is optional. But we should fail the Pool creation if user applies CR with more then 1 disk and does not provide raidConfig.

File: designs/replicated-pv/mayastor/raid0.md (hunk 7 above)
Comment (abhilashshetty04) [reply designs/replicated-pv/mayastor/raid0.md:85]: We should mention this in OEP?

>> No, it is not possible to add devices once the raid0 is online. One can potentially grow the underlying devices. The raid0 module processes device resize events and grows its own size in that case I think. I'm not sure if Mayastor/LVS can directly benefit from the raid growing in size or if we need some more changes.

We are working on Pool expansion now. So users might try to grow all disks in the raid group uniformly and try pool expansion. I've seen crypto bdev resize when base bdev gets resized. So we will support Online expansion there. worth checking this for raid bdev as well?

Comment (urso) [reply designs/replicated-pv/mayastor/raid0.md:85]: > We should mention this in OEP?

We can document this behavior later. But to me it is actually expected.

>  We are working on Pool expansion now. So users might try to grow all disks in the raid group uniformly and try pool expansion. I've seen crypto bdev resize when base bdev gets resized. So we will support Online expansion there. worth checking this for raid bdev as well?

The raid0 recomputes its size whenever one of its children is resized.

File: designs/replicated-pv/mayastor/raid0.md (hunk 6 above)
Comment (urso) [reply designs/replicated-pv/mayastor/raid0.md:102]: > Please go though the dsp operator code. We handle the CRD migration there. Also, do you mean rollback when you see backward compatible. If yes, we dont support rollback of CRDs. We introduce newer fields as optional. So they all are CRD migration/upgrade safe.

No. With backwards compatibility I mean that system/users will not be required to update existing DiskPool documents. The single-disk configuration (which is the only possible configuration today) will continue to work as is.
The idea is to make 'raidConfig' optional and only required if disks.len>1.

For example configuration for single disk without raid:

```
apiVersion: "openebs.io/v1beta3"
kind: DiskPool
metadata:
 name: <pool-name>
 namespace: <namespace>
spec:
 node: <node-name>
 disks: ["/dev/disk/by-id/<id>"]
```

And configuration for raid0 with 3 disks:


```
apiVersion: "openebs.io/v1beta3"
kind: DiskPool
metadata:
 name: <pool-name>
 namespace: <namespace>
spec:
 node: <node-name>
 disks:
   - "<disk1>"
   - "<disk2>"
   - "<disk3>"
 raidConfig:
   type: "raid0"
```

As 'raidConfig' is an optional field in the CRD spec we don't need to update the api version either, unless we really wish to update the version.

> In the above example, raidConfig is optional. But we should fail the Pool creation if user applies CR with more then 1 disk and does not provide raidConfig.

Absolutely. I think `raidConfig.type` is mandatory if disks.len>1

Comment (abhilashshetty04) [reply designs/replicated-pv/mayastor/raid0.md:102]: Sure, In past we had always used `optional` fields. however, we had bumped crd version. we are not doing it now in latest tasks. Please refer these PR. Here we are adding CRD validation tests also.

https://github.com/openebs/mayastor-control-plane/pull/1014
https://github.com/openebs/mayastor-control-plane/pull/1017

Lets also add `config: RaidConfig`. This will help future development. Different raid type might require different arguments. Lets separate them now itself.


```
apiVersion: "openebs.io/v1beta3"
kind: DiskPool
metadata:
 name: <pool-name>
 namespace: <namespace>
spec:
 node: <node-name>
 disks: ["/dev/disk/by-id/<id>"]
 raid:
   type: RaidType
   config: RaidConfig
```
   
```
enum RaidType {
     Raid0,
}
enum RaidConfig {
     Raid0(Raid0Config),
}
```


We will just extend RaidConfig and RaidType in future.

>> Absolutely. I think raidConfig.type is mandatory if disks.len>1

Even `RaidConfig` should be mandatory.  raid is optional.

tiagolobocastro: > ... We handle the CRD migration there.

This migration in hindsight we should not use moving forward, as it breaks creation of code which applies previous api version...
We could keep serving the old api version until such time we actually want to make a breaking release of mayastor.

> Even RaidConfig should be mandatory. raid is optional.

Yeah this we can do without breaking, make `raid` optional, so that if not specified then it's behaving as it is currently, and if it's specified then we can make any other fields within `raid` required.

designs/replicated-pv/mayastor/raid0.md
tiagolobocastro: Yeah this is what differs from plain concat right, and this allows resize of the raid0 device as opposed to concat, which would not be resizeable without an additional layer of mapping.

urso: I haven't checked other types then raid0, but indeed we need to check resize capabilities per raid type we want to support. Although there is some common raid bdev, resize computation is delegated to the actual raid level modules if I remember correctly.

designs/replicated-pv/mayastor/raid0.md
Abhinandan-Purkait: Sounds good!

designs/replicated-pv/mayastor/raid0.md
tiagolobocastro: > I really like the idea of having superblocks. So my proposal for now would be to attempt to read the device serial number and derive an UUID value from that + ask users to provide an UUID for devices that will not have such information.

Since the last version we now only allow users to give us /dev/by-xxx so that should be enough as a start?
For automatically setting up GPT we probably should tackle that as a separate OEP

tiagolobocastro (approved):
Let us know when you want to merge the provisional OEP, can be merged anytime and as often as needed.
When you change the status to implementable then it must be approved by maintainers as it'll signal acceptance

urso:
> Let us know when you want to merge the provisional OEP, can be merged anytime and as often as needed.
When you change the status to implementable then it must be approved by maintainers as it'll signal acceptance

Thanks. I will be out the next 2 weeks. I think I would prefer to merge the provisional OEP. I would like to create a separate PR with more low level details, so we can have a more focused discussion in the new PR.

designs/replicated-pv/mayastor/raid0.md
urso: > Since the last version we now only allow users to give us /dev/by-xxx so that should be enough as a start?

Potentially yes. Need to have a more detailed look. Let's discuss this level of detail in a more focused low-level details discussion PR for this OEP. As I already have a prototype there are quite a few details I would like to add/discuss separately.

tiagolobocastro:
> > Let us know when you want to merge the provisional OEP, can be merged anytime and as often as needed.
> > When you change the status to implementable then it must be approved by maintainers as it'll signal acceptance
> 
> Thanks. I will be out the next 2 weeks. I think I would prefer to merge the provisional OEP. I would like to create a separate PR with more low level details, so we can have a more focused discussion in the new PR.

Sounds good, please rebase and I'll go ahead and merge this

PR #4044, opened 2025-08-27
oep(4012): Mayastor Nexus QoS settings proposal

urso:
This OEP discusses the addition of QoS settings to the Mayastor Nexus.

### DIFF
--- a/designs/replicated-pv/mayastor/qos.md
+++ b/designs/replicated-pv/mayastor/qos.md
@@ -7,7 +7,7 @@ owners:
   - "@urso"
 editor: TBD
 creation-date: 2025-08-27
-last-updated: 2025-08-27
+last-updated: 2025-09-04
 status: provisional
 ---
 
@@ -22,16 +22,16 @@ status: provisional
   * [Non-Goals](#non-goals)
 * [Proposal](#proposal)
   * [User Stories](#user-stories)
-     * [Story 1: StorageClass with QoS Parameters](#story-1-storageclass-with-qos-parameters)
+     * [Story 1: Volume Creation with QoS](#story-1-volume-creation-with-qos)
      * [Story 2: Runtime QoS Modification](#story-2-runtime-qos-modification)
-     * [Story 3: Adding QoS to Non-QoS StorageClass](#story-3-adding-qos-to-non-qos-storageclass)
+     * [Story 3: Adding QoS to Existing Volume](#story-3-adding-qos-to-existing-volume)
   * [Implementation Details](#implementation-details)
   * [Risks and Mitigations](#risks-and-mitigations)
 * [Testing](#testing)
 
 ## Summary
 
-This proposal adds Quality of Service (QoS) support to Mayastor volumes, enabling users to set IOPS and bandwidth rate limits per volume. QoS limits can be configured via StorageClass parameters at provisioning time or modified through PVC annotations at runtime without disrupting active workloads. This provides resource isolation and predictable performance characteristics for applications in multi-tenant storage environments.
+This proposal adds Quality of Service (QoS) support to Mayastor volumes, enabling users to set IOPS and bandwidth rate limits per volume. QoS limits can be configured through Volume Attribute Classes at provisioning time or modified at runtime without disrupting active workloads. This provides resource isolation and predictable performance characteristics for applications in multi-tenant storage environments.
 
 ## Motivation
 
@@ -40,7 +40,7 @@ Mayastor needs QoS controls to enable resource isolation and predictable perform
 ### Goals
 
 - Implement IOPS and bandwidth rate limiting per volume using SPDK's native QoS capabilities
-- Enable runtime QoS modification via PVC annotations without volume restart
+- Enable runtime QoS modification via Volume Attribute Classes without volume restart
 
 ### Non-Goals
 
@@ -50,29 +50,28 @@ Mayastor needs QoS controls to enable resource isolation and predictable perform
 
 ## Proposal
 
-This is where we get down to the nitty gritty of what the proposal actually is.
+**Kubernetes Version Support:**
+Volume Attribute Classes is supported in Kubernetes 1.31+ (beta) and 1.34+ (stable).
 
 ### User Stories
 
-#### Story 1: StorageClass with QoS Parameters
+#### Story 1: Volume Creation with QoS
 
-As a user, I want to create a storage class with QoS parameters so that volumes provisioned from this StorageClass have IOPS and bandwidth limits applied.
+As a user, I want to create a volume with IOPS and bandwidth limits applied so that my application has predictable storage performance.
 
-**StorageClass:**
+**Volume Attribute Class:**
 ```yaml
-apiVersion: storage.k8s.io/v1
-kind: StorageClass
+apiVersion: storage.k8s.io/v1beta1
+kind: VolumeAttributesClass
 metadata:
-  name: mayastor
-provisioner: io.openebs.csi-mayastor
+  name: standard-performance
+driverName: io.openebs.csi-mayastor
 parameters:
-  protocol: "nvmf"
-  repl: "3"
   qos-iops-limit: "5000"
   qos-bandwidth-limit-mb: "200"
 ```
 
-**PVC using the StorageClass:**
+**PVC with QoS:**
 ```yaml
 apiVersion: v1
 kind: PersistentVolumeClaim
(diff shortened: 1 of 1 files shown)


designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,202 @@
...
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: mayastor
+provisioner: io.openebs.csi-mayastor
+parameters:
+  protocol: "nvmf"
+  repl: "3"
+  qos-iops-limit: "5000"
+  qos-bandwidth-limit-mb: "200"
```
tiagolobocastro: Could also move unit to the value?

```suggestion
  qos-bandwidth-limit: "200 MiB"
```

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,202 @@
...
+
+This proposal adds Quality of Service (QoS) support to Mayastor volumes, enabling users to set IOPS and bandwidth rate limits per volume. QoS limits can be configured via StorageClass parameters at provisioning time or modified through PVC annotations at runtime without disrupting active workloads. This provides resource isolation and predictable performance characteristics for applications in multi-tenant storage environments.
+
+## Motivation
+
+Mayastor needs QoS controls to enable resource isolation and predictable performance in multi-tenant storage environments. SPDK provides native QoS capabilities through rate limiting that can be leveraged to implement per-volume IOPS and bandwidth controls.
+
+### Goals
+
+- Implement IOPS and bandwidth rate limiting per volume using SPDK's native QoS capabilities
```
tiagolobocastro, line 42: Would the `QoS` be enforced at the replica level or the volume target (nexus) ?

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+**PVC with runtime QoS override:**
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: database-storage
+  annotations:
+    openebs.io/qos-iops-limit: "8000"      # Override: increased from 2000
```
tiagolobocastro: So far modifications to volume parameters have been made via the plugin, so there's currently no mechanism for syncing from PVC into VolumeSpec, so something would have to be added for this.

Have you looked into [Volume Attribute Classes](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/) btw?

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,202 @@
...
+spec:
+  storageClassName: mayastor-basic
+  ...
+```
+
+### Implementation Details
+
+**CSI Integration:**
+- CSI driver processes StorageClass parameters for initial QoS configuration
+- SPDK QoS integration via existing JsonGrpc proxy
```
tiagolobocastro, line 134: Can't remember if this enabled by default currently.
We could also add a QoS interface, ex: [stats](https://github.com/openebs/mayastor-dependencies/blob/develop/apis/io-engine/protobuf/v1/stats.proto)

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,202 @@
...
+- Annotation names match StorageClass parameter names using `openebs.io/` namespace prefix
+- `qos-iops-limit` / `openebs.io/qos-iops-limit` - IOPS rate limit, becomes `rw_ios_per_sec`
+- `qos-bandwidth-limit-mb` / `openebs.io/qos-bandwidth-limit-mb` - Combined read/write bandwidth limit, becomes `rw_mbytes_per_sec`
+- `qos-read-bandwidth-limit-mb` / `openebs.io/qos-read-bandwidth-limit-mb` - Read-only bandwidth limit, becomes `r_mbytes_per_sec`
+- `qos-write-bandwidth-limit-mb` / `openebs.io/qos-write-bandwidth-limit-mb` - Write-only bandwidth limit, becomes `w_mbytes_per_sec`
+
+**PVC Annotation Monitoring:**
+- Dedicated PVC watcher added to CSI driver monitors annotation changes on existing PVCs
+- Detects QoS annotation additions, modifications, and removals
+- Validates parameters before applying changes to live volumes
```
tiagolobocastro: We could combine this and VolumeSpec, and let agent-core apply the live changes, since on republish it would have to apply the changes again?

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,202 @@
...
+**Validation and Constraints:**
+- Parameter validation ensures non-negative values and meets SPDK minimum limits
+- QoS applied at nexus level (published volumes only)
+- SPDK minimum limits: 1000 IOPS, 10 MB/s bandwidth
+- Units: IOPS (operations/second), bandwidth in megabytes/second
+
+### Risks and Mitigations
+
+**Resource Contention:**
+- Risk: Multiple volumes with high QoS limits could exceed underlying storage capabilities, causing system-wide performance degradation despite individual volume limits.
```
tiagolobocastro, line 165: For this we should expose accrued QoS on a given node or pool, allowing users to check if any resource is potentially oversubscribed?

designs/replicated-pv/mayastor/qos.md
tiagolobocastro: Btw have you looked into doing QoS at the initiator side instead of storage? And what pros/cons there are?

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,202 @@
...
+     * [Story 1: StorageClass with QoS Parameters](#story-1-storageclass-with-qos-parameters)
+     * [Story 2: Runtime QoS Modification](#story-2-runtime-qos-modification)
+     * [Story 3: Adding QoS to Non-QoS StorageClass](#story-3-adding-qos-to-non-qos-storageclass)
+  * [Implementation Details](#implementation-details)
+  * [Risks and Mitigations](#risks-and-mitigations)
+* [Testing](#testing)
+
+## Summary
+
+This proposal adds Quality of Service (QoS) support to Mayastor volumes, enabling users to set IOPS and bandwidth rate limits per volume. QoS limits can be configured via StorageClass parameters at provisioning time or modified through PVC annotations at runtime without disrupting active workloads. This provides resource isolation and predictable performance characteristics for applications in multi-tenant storage environments.
```
Abhinandan-Purkait: Can we use the ModifyVolume csi wf rather than using the annotations? ref: https://github.com/container-storage-interface/spec/blob/master/spec.md#controllermodifyvolume

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Motivation
+
+Mayastor needs QoS controls to enable resource isolation and predictable performance in multi-tenant storage environments. SPDK provides native QoS capabilities through rate limiting that can be leveraged to implement per-volume IOPS and bandwidth controls.
+
+### Goals
+
+- Implement IOPS and bandwidth rate limiting per volume using SPDK's native QoS capabilities
+- Enable runtime QoS modification via PVC annotations without volume restart
```
Abhinandan-Purkait: As I suggested, we should use the CSI workflow, that would probably save us lot of efforts on creating a new controller to watch for PVC annotation changes.

designs/replicated-pv/mayastor/qos.md
Abhinandan-Purkait: Yeah, even I was wondering the same, we should have a look at it.

Abhinandan-Purkait: Or we can probably just let it be and modifications just supported via plugin, since we do that anyway today, that might simplify things to start with.

designs/replicated-pv/mayastor/qos.md
urso: I was considering this as I would prefer adding units to the value myself. But the proper unit is `MiB/s`, not `MiB`. If we are fine with 'MiB' as unit I will gladly change that.

designs/replicated-pv/mayastor/qos.md
urso: It would be enforced at the Nexus only.

> Btw have you looked into doing QoS at the initiator side instead of storage? And what pros/cons there are?

I think we are a little more limited here. I'm not aware of any iops/bandwidth settings for the Linux NVMe modules. So one would either use TC (Traffic Control tool) to impose limits on the TCP connection or cgroups blkio.
While TC might be a good option to ensure network being well configured, it is somewhat non-intuitive on how these will translate to IOPS for example. Regarding cgroups: while k8s allows you to configure cpu/memory limits I'm not aware of any disk related cgroups settings being made available to users. But even if blkio were available, this causes other issues like the application being affected in total, even if multiple volumes are mounted or sidecar containers for logs/backups requiring their own limits.

designs/replicated-pv/mayastor/qos.md
urso: > Have you looked into [Volume Attribute Classes](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/) btw?

Ah no, I haven't seen this new feature. Thank you for bringing it to my attention. I'm a little on the fence here. Using annotations gives us some flexibility to change a single device at any time. On the other hand `Volume Attribute Classes` seem to become a new standard and it would be preferable to support that.

Short term this would mean it limits us to kubernetes 1.34 (which is not released yet), while still actively maintained versions will not be able to support that feature (unless users are willing to use alpha/beta feature I think).

> Or we can probably just let it be and modifications just supported via plugin, since we do that anyway today, that might simplify things to start with.

Sounds good.

designs/replicated-pv/mayastor/qos.md
urso: Hm, I don't think we need to expose anything else. I would hope it is a matter monitoring the node itself. But it wouldn't hurt to make users aware of that in the Documentation or FAQ.

designs/replicated-pv/mayastor/qos.md
urso: I think this is related to the Volume Attribute Class discussion, isn't it?

Abhinandan-Purkait: Indeed.

Abhinandan-Purkait: For that you would need this I think.

urso:
@Abhinandan-Purkait @tiagolobocastro I think almost all discussions point towards 'Volume Attribute Classes'. Given that this is a standard now and it looks like it is easier to implement I will update the proposal document to use 'Volume Attribute Classes' in favor over PVC annotations.

Shall we close this PR and create a new one?

designs/replicated-pv/mayastor/qos.md
tiagolobocastro: hmm good point...
@Abhinandan-Purkait wdyt?

tiagolobocastro:
> @Abhinandan-Purkait @tiagolobocastro I think almost all discussions point towards 'Volume Attribute Classes'. Given that this is a standard now and it looks like it is easier to implement I will update the proposal document to use 'Volume Attribute Classes' in favor over PVC annotations.
> 
> Shall we close this PR and create a new one?

The first step I see anyhow is adding support for modifying the volume spec via rest, and one could modify via plugin for example.

Would need to check but if `Volume Attribute Classes` requires using an unreleased version of K8s, then I'd probably stay on the fence, and have it as a future improvement, since anyway we'd probably want to do that for the other volume parameters (ex: replicas, topology, encryption).

If for your use case you want to modify the parameters via PVC rather than the plugin, then we'd need a new controller. That should be ok too, but probably we should not make liberal use of this for too many new features *if* we are to move to `Volume Attribute Classes` in the near future.

designs/replicated-pv/mayastor/qos.md
tiagolobocastro: I'm thinking in similar terms to thin provisioning for example, where we expose a pool's `overcommitment` and remaining free space. A user can monitor this today and setup alert rules based on the diskpool metrics exporter.
Though it would not be the same here, since we're doing it at the volume level rather than pool level.
Anyway we can start with Docs and FAQ

urso:
> Would need to check but if Volume Attribute Classes requires using an unreleased version of K8s

I think we are fine.
Beta since 1.31: https://kubernetes.io/blog/2024/08/15/kubernetes-1-31-volume-attributes-class/
Stable in 1.34 which was actually released 2 days ago.

designs/replicated-pv/mayastor/qos.md
urso: > Can't remember if this enabled by default currently.

I think it is enabled. Checking io-engine grpc server I can tell a JsonService is registered for v0 and v1 API always.

designs/replicated-pv/mayastor/qos.md
tiagolobocastro: Yeah I'm not sure if it's possible to have fine grained control when using cgroups, or even if it's possible to configure that from within a CSI driver... 
So let's keep storage-side `QoS`

tiagolobocastro:
> > Would need to check but if Volume Attribute Classes requires using an unreleased version of K8s
> 
> I think we are fine. Beta since 1.31: https://kubernetes.io/blog/2024/08/15/kubernetes-1-31-volume-attributes-class/ Stable in 1.34 which was actually released 2 days ago.

We're fine there then.
A potential downside btw, is that it's not possible to control on a per-volume basis AFAICT, you'd have to create a new attribute class?

urso:
> A potential downside btw, is that it's not possible to control on a per-volume basis AFAICT, you'd have to create a new attribute class?

Correct. Still you can create a new attribute class and change the attribute class of a volume at any time. In worst case you will have to create a configuration per PVC, although I think this is very unlikely someone wants to do just that. At least for my use-case (multi-tenant shared storage) I would define reusable "performance profiles" via the Volume Attribute Class.

My original proposal wants to add the settings to the storageClass + annotations as fallback. This would require me to create a storageClass configuration per 'performance profile' I have. What I like about Volume Attribute Class though is that it is somewhat orthogonal to storageClass. I can configure storage and performance profiles somewhat independently. I definitely can see how this is a great feature for our use-case. We can also easily change the performance profile of volumes by assigning them to another attribute class, which allows us to downgrade/upgrade existing volumes in a controlled and standardized manner.

To me having PVC annotations is kinda nice for life incidents, but I can create attribute classes in that case if I really need to. The added complexity annotations would require is not worth it.

designs/replicated-pv/mayastor/qos.md
urso: Out of scope for this proposal, but one thing I wonder is:
Would it make sense to take QoS settings into account when allocation the volumes for the Nexus? I think right now we choose the node with most available space (assuming thin provisioning). Available bandwidth might be another limiting factor one would want to take into account.

tiagolobocastro: Yeah that could be a future enhancement.
Would we then take into account as a preference or enforcement? If enforcement then what do do if there's no available bandwidth, treat as a failure?

urso:
@Abhinandan-Purkait @tiagolobocastro I updated the proposal to use Volume Attribute Classes. Maybe you want to have another look. 
To me one discussion point that needs clarification is how we should do bandwidth units.

designs/replicated-pv/mayastor/qos.md
Abhinandan-Purkait: Hmm, but by specifying we are also suggesting the users need to make the conversion for units > or < mib/mb. What about taking the size as input rather than rate? Like `qos-bandwidth-limit-size: "200 MiB"`, then it's suggestive that we are speaking out the size bit here?

Abhinandan-Purkait (approved):
Looks good to me in terms of the proposal. Thanks!

designs/replicated-pv/mayastor/qos.md
urso: yeah, might work. maybe `qos-bandwidth-limit-per-sec: "200MiB"` ?

Alternatively we could introduce a 'Rate' type that parses "200MiB/s" (and/or alternative "MiBps").

Abhinandan-Purkait: `qos-bw-limit-per-sec: "200MiB"` sounds good to me as well 👍🏻

tiagolobocastro: either sounds good to me

designs/replicated-pv/mayastor/qos.md
tiagolobocastro: Also a way of querying the current value sets may be required for reconciliation for example
And if so, it may be worth adding these to the nexus api for example, meaning they'd get collected together with the nexus information

tiagolobocastro (approved):

designs/replicated-pv/mayastor/qos.md
urso: > If enforcement then what do do if there's no available bandwidth, treat as a failure?

We don't really know the proper bandwidth in advance and it might even depend on access patterns. In my mind this is mostly best-effort. You still need monitoring to verify the system is performing as expected.

designs/replicated-pv/mayastor/qos.md
urso: I think that RPC exists already. The `bdev_get_bdevs` response also includes an `assigned_rate_limits` field.

urso: Looking into this we already fetch the nexus status from the io-engine. To not make an extra RPC call I plan to add the current qos configuration to the nexus status response. I think this is what you were hinting at.

tiagolobocastro: Yes that's right, we could reuse the same one

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,182 @@
...
+**QoS Parameters:**
+- SPDK rate limit types (`rw_ios_per_sec`, `rw_mbytes_per_sec`, `r_mbytes_per_sec`, `w_mbytes_per_sec`) mapped to VolumeAttributesClass parameters
+- `qos-iops-limit` - IOPS rate limit, becomes `rw_ios_per_sec`
+- `qos-bandwidth-per-sec` - Combined read/write bandwidth limit, becomes `rw_mbytes_per_sec`
+- `qos-read-bandwidth-per-sec` - Read-only bandwidth limit, becomes `r_mbytes_per_sec`
+- `qos-write-bandwidth-per-sec` - Write-only bandwidth limit, becomes `w_mbytes_per_sec`
+
+**Volume Modification Workflow:**
+- Kubernetes triggers `ControllerModifyVolume` RPC when PVC's `volumeAttributesClassName` changes
+- CSI driver validates new VolumeAttributesClass parameters
```
dsharma-dc, line 146: What are these validations?

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,182 @@
...
+**Status Reporting:**
+- Applied QoS limits added to volume status when configured for operational visibility
+- Validation errors and conflicts reported via `ControllerModifyVolume` RPC response
+- Status reflects actual SPDK-applied limits from VolumeAttributesClass parameters
+
+**Validation and Constraints:**
+- Parameter validation ensures non-negative values and meets SPDK minimum limits
+- QoS applied at nexus level (published volumes only)
+- SPDK minimum limits: 1000 IOPS, 10 MB/s bandwidth
+- Units: IOPS (operations/second), bandwidth in megabytes/second
```
dsharma-dc, line 160: ok, looks like these are the validations that CSI driver has to do?

dsharma-dc (approved):

designs/replicated-pv/mayastor/qos.md
tiagolobocastro: Looks like `lvm-localpv` has cgroup rate limits: 
- <https://github.com/openebs/lvm-localpv/blob/f103ba5e3fa2521a3bc3e6e9c53608942b950b5d/pkg/lvm/mount.go#L278>
- <https://github.com/openebs/lib-csi/tree/develop/pkg/device/iolimit>

designs/replicated-pv/mayastor/qos.md
```diff
@@ -0,0 +1,182 @@
...
+**Volume Attribute Class:**
+```yaml
+apiVersion: storage.k8s.io/v1beta1
+kind: VolumeAttributesClass
+metadata:
+  name: standard-performance
+driverName: io.openebs.csi-mayastor
+parameters:
+  qos-iops-limit: "5000"
+  qos-bandwidth-per-sec: "200MiB"
```
tiagolobocastro, line 71:
```suggestion
  qosIopsLimit: "5000"
  qosBandwidthPerSec: "200MiB"
```

I think these should be camel case to keep in line with SC parameters

urso:
After some more discussion we think it rather makes sense to consider a QoS implementation to be client side similar to `lvm-localpv`.

Closing this OEP.

PR #4060, opened 2025-09-18
oep(4059): add mayastor rwx block support with kubevirt

tiagolobocastro:

### DIFF
--- a/designs/replicated-pv/mayastor/rwx-kubevirt.md
+++ b/designs/replicated-pv/mayastor/rwx-kubevirt.md
@@ -1,0 +1,265 @@
+---
+oep-number: OEP 4059
+title: RWX Block Volume Support in Mayastor for KubeVirt
+authors:
+  - "@tiagolobocastro"
+owners:
+  - "@tiagolobocastro"
+editor: TBD
+creation-date: 18/09/2025
+last-updated: 18/09/2025
+status: implementable
+---
+
+# RWX Block Volume Support in Mayastor for KubeVirt
+
+<br>
+
+## Table of Contents
+
+- [🎯 Summary](#-summary)
+- [📌 Motivation](#-motivation)
+  - [🧱 Goals](#-goals)
+  - [🚫 Non-Goals](#-non-goals)
+- [🧵 Proposal](#-proposal)
+  - [📚 User Stories](#-user-stories)
+    - [🧑‍💻 Story 1](#-story-1)
+    - [🧑‍💻 Story 2](#-story-2)
+  - [🛠️ Implementation Details/Notes/Constraints](#️-implementation-detailsnotesconstraints)
+    - [Public OpenAPI](#public-openapi)
+    - [Internal gRPC](#internal-grpc)
+    - [Kubectl Plugin](#kubectl-plugin)
+  - [Test Plan](#test-plan)
+    - [Behaviour specification](#behaviour-specification)
+  - [Risks and Mitigations](#risks-and-mitigations)
+    - [Mitigations](#mitigations)
+- [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks \[optional\]](#drawbacks-optional)
+- [Alternatives \[optional\]](#alternatives-optional)
+
+<br>
+
+## 🎯 Summary
+
+This proposal introduces RWX (ReadWriteMany) block volume support in Mayastor to enable "shared" block storage for KubeVirt virtual machines. The enhancement aims to support VM migration across nodes, improving HA, maintenance, etc
+
+<br>
+
+## 📌 Motivation
+
+Our users' KubeVirt VMs currently rely on PVCs with `ReadWriteOnce` access mode, limiting their ability to share block devices across nodes.
+Kubevirt requires PVCs with `ReadWriteMany` access mode to enable its [live migration](https://kubevirt.io/user-guide/compute/live_migration/).
+
+<br>
+
+### 🧱 Goals
+
+- Implement RWX block volume support in Mayastor.
+- Ensure compatibility with KubeVirt VM disk attachments.
+- Provide CSI-compliant RWX semantics for block volumes.
+- Support Persistent Reservations (optional) for fencing.
+
+<br>
+
+### 🚫 Non-Goals
+
+- RWX support for filesystem volumes.
+- RWX support for non-KubeVirt workloads.
+- VM-level fencing or quorum management (we don't want to share, but rather to switch!).
+
+<br>
+
+## 🧵 Proposal
+
+As per my understanding, KubeVirt itself handles the flushing and switching of the storage to the new node, which means we don't even have to implement reservations.
+The CSI spec allows for many access modes, example:
+
+- ***MULTI_NODE_SINGLE_WRITER***
+  A volume can be published at multiple nodes simultaneously. Only one of the node can be used as read/write. The rest will be readonly.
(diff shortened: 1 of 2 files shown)


designs/replicated-pv/mayastor/rwx-kubevirt.md
```diff
@@ -0,0 +1,132 @@
+---
+oep-number: OEP 3903
+title: Cordon Mayastor Pools
```
Abhinandan-Purkait:
```suggestion
title: RWX Block Volume Support
```

designs/replicated-pv/mayastor/rwx-kubevirt.md
```diff
@@ -0,0 +1,132 @@
...
+        Rest
+    end
+
+    csi-controller
+
+    kubelet-1 --> |CSI Api| csi-node-1
+    kubelet-2 --> |CSI Api| csi-node-2
+    csi-controller --> |Public Api| Rest
+    csi-node-1 --> |Public Api| Rest
+    csi-node-2 --> |Public Api| Rest
```
Abhinandan-Purkait: Why does csi node need to interact with rest api? What information are we gathering here?

designs/replicated-pv/mayastor/rwx-kubevirt.md
tiagolobocastro: sigh, also had forgotten to fix oep number... thanks!

designs/replicated-pv/mayastor/rwx-kubevirt.md
tiagolobocastro: Was only here to show we have a connection, but actually doesn't need to, only confusing things... removed it

Abhinandan-Purkait (approved):

designs/replicated-pv/mayastor/rwx-kubevirt.md
```diff
@@ -0,0 +1,261 @@
...
+<br>
+
+### 🛠️ Implementation Details/Notes/Constraints
+
+Following a top-down approach, here are the required modifications:
+
+1. Extend Mayastor CSI to advertise `MultiNodeMultiWriter`, but only for block mode volumes.
+   1. this means we'd reject it for fs volumes, since that cannot possibly work!
+2. Core agent must now accept publishes, even if the volume is already published, but for a different node
+3. Core agent must collate all valid initiator nqns when building the access control list for the NVMe-oF target.
```
dsharma-dc, line 115: Are we going to limit and validate this to 2 initiators, because that's the only valid case of migration I think.

dsharma-dc:
It's not clear to me how are we going to ensure data access synchronisation once the two initiators are talking to the same volume target. e.g. this is something NFS inherently takes care of using hierarchical locking ( with nfsv4.x).

designs/replicated-pv/mayastor/rwx-kubevirt.md
```diff
@@ -0,0 +1,261 @@
...
+- Ensure compatibility with KubeVirt VM disk attachments.
+- Provide CSI-compliant RWX semantics for block volumes.
+- Support Persistent Reservations (optional) for fencing.
+
+<br>
+
+### 🚫 Non-Goals
+
+- RWX support for filesystem volumes.
+- RWX support for non-KubeVirt workloads.
```
abhilashshetty04, line 68: How are we going to know if the Block type pvc is going to be plugged into kubevirt pod (I guess its going to be used by virt-launcher)?

designs/replicated-pv/mayastor/rwx-kubevirt.md
```diff
@@ -0,0 +1,261 @@
...
+### Risks and Mitigations
+
+To ensure high-availability, when the connection from the initiator to the target becomes failed we trigger failover of the volume target to other nodes.
+With multi-attach we now have multiple initiator-target connections, which may request for failover all at once, or in short sequence! This could wreak avoc in the system, leading into a self-feeding loop of failovers.
+
+<br>
+
+#### Mitigations
+
+In order to mitigate this, we must ensure the failover logic is hardened and resistive to short-succesion failovers, as well as prescriptive forcing the initiator to attempt reconnection against "new-existing" targets. By this I mean, if a previous failover creates a new target, then the next failover requests shoulds try to connect this "new-existing" target, rather than causing an even newewer target to be created, which would then break the first failover!
```
abhilashshetty04:
```suggestion
In order to mitigate this, we must ensure the failover logic is hardened and resistive to short-succesion failovers, as well as prescriptive forcing the initiator to attempt reconnection against "new-existing" targets. By this I mean, if a previous failover creates a new target, then the next failover requests should try to connect this "new-existing" target, rather than causing an even newer target to be created, which would then break the first failover!
```

designs/replicated-pv/mayastor/rwx-kubevirt.md
tiagolobocastro: We can only check for Block, not for the specific app, this has to be well documented

designs/replicated-pv/mayastor/rwx-kubevirt.md
tiagolobocastro: hmm I guess we could do that, though what happens if the second node becomes `NotReady` during migration and then a third node is selected?

designs/replicated-pv/mayastor/rwx-kubevirt.md
tiagolobocastro: thanks, fixed

designs/replicated-pv/mayastor/rwx-kubevirt.md
dsharma-dc: > hmm I guess we could do that, though what happens if the second node becomes `NotReady` during migration and then a third node is selected?

Does Kubevirt ensure this kind of continuity where migration is happening from say node1 to node2, and one of them goes `NotReady` during migration?

tiagolobocastro: @krishnaGajabi do you know?

tiagolobocastro:
> It's not clear to me how are we going to ensure data access synchronisation once the two initiators are talking to the same volume target. e.g. this is something NFS inherently takes care of using hierarchical locking ( with nfsv4.x).

As mentioned, it's a non-goal. It's expected that kubevirt will its own synchronization

designs/replicated-pv/mayastor/rwx-kubevirt.md
krishnaGajabi: KubeVirt supports three live migration strategies: pre-copy, post-copy, and auto-converge. The behavior during failures depends on which strategy is used.

Pre-copy is generally the safest approach. In this mode, all memory pages are first copied to the destination node, and only after the transfer is complete does the VM switch execution to the destination. This ensures that if the destination node fails mid-transfer, the VM keeps running on the source node without interruption.

Other strategies like post-copy and auto-converge have different trade-offs in terms of performance, downtime, and fault tolerance

For more details, see : https://kubevirt.io/user-guide/compute/live_migration/#understanding-different-migration-strategies

tiagolobocastro: At least using pre-copy it seems that migration can be cancelled; so this means we may have to support >2 publishes

dsharma-dc (approved):

niladrih (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
@mergify queue

PR #4061, opened 2025-09-19
Use openebs/kubectl instead of bitnamilegacy/kubectl

niladrih:
This is similar to:
- https://github.com/openebs/mayastor-extensions/commit/b09b15e50244d012538c59bc98919d79a8ae7bb5

This change doesn't require upgrade changes, because upgrade here uses `helm upgrade --reset-then-reuse-values`. The resetting will place the new value above the older one.

### DIFF
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -74,7 +74,7 @@ preUpgradeHook:
     # -- The container image registry URL for the hook job
     registry: docker.io
     # -- The container repository for the hook job
-    repo: bitnamilegacy/kubectl
+    repo: openebs/kubectl
     # -- The container image tag for the hook job
     tag: "1.25.15"
     # -- The imagePullPolicy for the container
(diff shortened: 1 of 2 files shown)


charts/values.yaml
```diff
@@ -74,7 +74,7 @@ preUpgradeHook:
     # -- The container image registry URL for the hook job
     registry: docker.io
     # -- The container repository for the hook job
-    repo: bitnamilegacy/kubectl
+    repo: openebs/kubectl
```
tiagolobocastro, line 77: oh I thought I had done this already, must have missed this one!
I think we need same fix on this repo: <https://github.com/openebs/mayastor-extensions/commit/544a3f14041543736adc1fe0948908377e50ed59>
Please make that change and see if it would catch this

tiagolobocastro (approved):

tiagolobocastro:
just testing the bot...
@mergifyio rebase

charts/values.yaml
niladrih: @tiagolobocastro This change is already there in mayastor-extensions.

niladrih: Oh, I got it.. okay..

Abhinandan-Purkait (approved):

charts/values.yaml
niladrih: Done

scripts/helm/images.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -euo pipefail
+set -x
```
tiagolobocastro: this doesn't need to be merged right?

niladrih: Dammit

niladrih: Fixed. Thanks.

Issue #4076, opened 2025-10-07
[OEP 4076]: Support native LVM tags for LVs in lvm-localpv


jochenseeber:
### Enhancement Description

Add support to set native LVM tags on LVs created by lvm-localpv

- OpenEBS Enhancement Proposal: tbd
- Discussion Link: tbd
- Primary contact (assignee): tbd
- Responsible OWNERs: tbd
- Targeted milestone:
  - tbd
- [ ] Targets
  - [ ] OEP update PR(s):
  - [ ] Code update PR(s):
  - [ ] E2E Test PR(s):
  - [ ] Docs update PR(s):

PR #4077, opened 2025-10-07
oep(4076): add proposal for native LVM tags

jochenseeber:
* Added proposal to support native LVM tags on LVs (#4076)

### DIFF
--- a/designs/local-pv/lvm/native-lvm-tags.md
+++ b/designs/local-pv/lvm/native-lvm-tags.md
@@ -1,0 +1,124 @@
+---
+oep-number: OEP 4076
+title: LocalPV-LVM tag support
+authors:
+  - "@jochenseeber"
+owners: []
+editor: []
+creation-date: 2025-10-07
+last-updated: 2025-10-07
+status: provisional
+---
+
+# Native LVM Tags
+
+## Table of Contents
+
+* [Native LVM Tags](#native-lvm-tags)
+  * [Table of Contents](#table-of-contents)
+  * [Summary](#summary)
+  * [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non\-Goals](#non-goals)
+  * [Proposal](#proposal)
+    * [User Stories](#user-stories)
+      * [Setting tags using the storage class](#setting-tags-using-the-storage-class)
+      * [Setting tags using the persistent volume claim](#setting-tags-using-the-persistent-volume-claim)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+  * [Testing](#testing)
+  * [Graduation Criteria](#graduation-criteria)
+
+## Summary
+
+This proposal suggests supporting native LVM tags for Logical Volumes (LVs) created by lvm-localpv. Tags are strings that are attached to the LV and can be used to identify LVs for LVM commands, LVM configuration, or other commands.
+
+For example, LVs could be tagged with their environment (`env=prod` or `env=test`), by their required backup strategy (`backup=daily` or `backup=weekly`). Admins could then use these tags to target a group of LVs in one go.
+
+## Motivation
+
+From the LVM man page:
+
+> Tags are user-defined strings that can be attached to PVs, VGs and LVs. Tags can be displayed with commands `pvs/vgs/lvs -o tags`. Certain commands will accept a tag name in place of a PV, VG, or LV name. In these cases, the command will operate on each PV/VG/LV with the given tag. Tags should be prefixed with `@` to avoid ambiguity.
+>
+> Characters allowed in tags are: `A–Z a–z 0–9 _ + . - / = ! : # &`
+
+Tags would greatly simplify admin work by allowing admins to target groups of related LVs (e.g. all production LVs) instead of individual LVs. For example:
+
+* Configuration Management: LVM's configuration file can use tags to apply specific settings to groups of volumes, allowing for fine-grained control over the storage infrastructure.
+* Selective Backups: An admin could easily script a backup routine that targets all volumes with the tag `backup=daily`.
+* Targeted LVM Operations: When performing maintenance, an admin could use LVM commands to check all logical volumes tagged as `env=dev` in a single operation.
+
+### Goals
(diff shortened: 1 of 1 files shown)


designs/local-pv/lvm/native-lvm-tags.md
```diff
@@ -0,0 +1,124 @@
...
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: production-database
+allowVolumeExpansion: true
+parameters:
+  volgroup: lvmvg
+  tags: env=prod,backup=never # Just checking if you're paying attention ;-)
```
tiagolobocastro, line 77: 😄

designs/local-pv/lvm/native-lvm-tags.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Goals
+
+* Allow the user to set and add tags to created LVs
+* Specify tags in the storage class for tags that depend on the storage class
+* Specify tags in the PVC's volume attribute class for tags that depend on the individual request
+
+### Non-Goals
+
+* Removing tags
```
tiagolobocastro, line 59: Is this in case it conflicts with something the user might do directly with lvm?
If so a couple of suggestions:
- we could prefix with `openebs.io/$key=$value` so we know it's ours
- empty would remove it, ex: `env=,backup=maybe`

designs/local-pv/lvm/native-lvm-tags.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### Setting tags using the persistent volume claim
+
+* As a user, I can specify in the persistent volume claim which tags should be added to any created LV
+* As a user, I can update the tags parameter in the storage class, and new tags will be applied
+
+```yaml
+---
+apiVersion: storage.k8s.io/v1
+kind: VolumeAttributesClass
```
tiagolobocastro, line 89: We don't support `VolumeAttributesClass` today AFAIK

tiagolobocastro (approved):

designs/local-pv/lvm/native-lvm-tags.md
jochenseeber: Tags are arbitrary strings, not key/value pairs, so "key=" is a tag by itself. We could of course interpret it that way, but then what if somebody wants to set "key=" for some reason.

Does Kubernetes provide old versions of resources when listening to updates? Then we could use that to do a "tag diff".

designs/local-pv/lvm/native-lvm-tags.md
jochenseeber: I would suggest adding that.

Abhinandan-Purkait (approved):

designs/local-pv/lvm/native-lvm-tags.md
tiagolobocastro: That would be good, though I think that requires some work, probably we should have an OEP for that if it's a considerable amount of work?
CC @Abhinandan-Purkait

Abhinandan-Purkait: Difficult to say how big it would be but it should be straightforward. We can consider adding VolumeAttributesClass support.

abhilashshetty04: Looking at the documentation, VolumeAttributeClass is meant to be used for Volume specific attributes. VAC in the PVC can be changed later if provisioner supports ModifyVolume RPC.

For the localpv-lvm, Currently we have `vgpattern` and `scheduler`. These are not a mutable attributes of the Volume once created.  Tags i suppose can be consider as lvm does allow us to updated LV tags. We may use it to do raid specific changes in future. Example: triggering a repair, increasing mirror count etc.

From the implementation perspective, Volume attributes are mapped to `mutable_parameters` of CreateVolumeRequest. Its available in v1.9 (current version of localpv-lvm).

designs/local-pv/lvm/native-lvm-tags.md
```diff
@@ -0,0 +1,124 @@
...
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 4Gi
+
+```
+
+### Implementation Details/Notes/Constraints
+
+* When creating a LV, it is tagged with the union of all tags specified in the storage class and in the PVC's VolumeAttributeClass
```
abhilashshetty04, line 113: tag should be passed using VolumeAttributeClass If we introduce it?

designs/local-pv/lvm/native-lvm-tags.md
tiagolobocastro: I propose we consider tag removal part of `VolumeAttributeClass` since #4146 will introduce `VolumeAttributeClass` .

designs/local-pv/lvm/native-lvm-tags.md
tiagolobocastro: > Does Kubernetes provide old versions of resources when listening to updates? Then we could use that to do a "tag diff".

Not that I know of.

So we have a few options:
1. Do not support tag removal, user must handle by themselves
2. Use `-` for removal 
  `kubectl` does removal like so: `kubectl label node x key-`
  This removes label with key name of `key`.
3. Any tag not present in the `lvmVolume CR spec` is removed
  Downside is this does not allow user modifying the tags or perhaps even importing volumes with tags?
4. Use prefix `openebs.io/` and this way we'd only add or remove tags which start with that prefix

niladrih (approved):

Abhinandan-Purkait:
@jochenseeber Can you please sign your commits, then we can take this in

designs/local-pv/lvm/native-lvm-tags.md
jochenseeber: I think number 2 is the best option. I would not want to remove tags that are not "ours", and I also would like to be able to set tags that may be picked up by other components, and that are not prefixed.

niladrih (approved):

tiagolobocastro:
@mergify queue

PR #4107, opened 2025-11-18
oep(4111): format mayastor node status metrics doc

IyanekiB:
This OEP proposes shifting Mayastor node-status metrics into the existing metrics-exporter by adding a lightweight REST client that polls `/v0/nodes` and exposes Prometheus gauges for online, cordoned, and draining states. The control-plane stays the source of truth, while the exporter becomes the single place where metrics are published.

**Status:** implementing

**Related implementation:** https://github.com/openebs/mayastor-control-plane/pull/1035

**CC:** @pjgranieri

### DIFF
--- a/designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md
+++ b/designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md
@@ -1,0 +1,122 @@
+---
+oep-number: OEP 4111
+title: mayastor-node-status-metrics-via-metrics-exporter
+authors:
+  - "@IyanekiB"
+  - "@pjgranieri"
+owners:
+  - "@tiagolobocastro"
+  - "@Abhinandan-Purkait"
+  - "@niladrih"
+editor: TBD
+creation-date: 2025-11-18
+last-updated: 2025-11-18
+status: provisional
+replaces:
+superseded-by:
+---
+
+# Mayastor Node Status Metrics via Metrics Exporter
+
+## Table of Contents
+
+* [Summary](#summary)
+* [Motivation](#motivation)
+  * [Goals](#goals)
+  * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+  * [Implementation Details](#implementation-details)
+  * [Risks and Mitigations](#risks-and-mitigations)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+* [Drawbacks](#drawbacks)
+* [Alternatives](#alternatives)
+* [Infrastructure Needed](#infrastructure-needed)
+* [Testing](#testing)
+
+## Summary
+
+This proposal establishes the metrics-exporter as the single surface for Mayastor node-status metrics by extending it with a lightweight REST client.
+Instead of generating metrics directly in the control-plane REST service, the exporter periodically queries control-plane endpoints, derives node states, and exposes Prometheus-compatible gauges for online, cordoned, and draining states.
+
+## Motivation
+
+Surfacing metrics directly from the REST service risks duplicating business logic, requires additional auth/scrape plumbing, and can produce divergent metric formats.
+Centralizing node state reporting inside the metrics-exporter aligns with how other Mayastor operational metrics are exposed and creates a reusable pattern for REST-backed observations.
+
+This work follows the lessons learned in [PR #1035](https://github.com/openebs/mayastor-control-plane/pull/1035), which introduced node-status metrics at the REST layer.
+Moving this logic into the exporter yields a more maintainable, observable, and testable pipeline.
+
+### Goals
+
+* Add node-status gauges to the existing metrics-exporter binary.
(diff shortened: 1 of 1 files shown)


designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md
```diff
@@ -0,0 +1,121 @@
+---
+oep-number: TBC
+title: mayastor-node-status-metrics-via-metrics-exporter
+authors:
+  - "@IyanekiB"
+  - "@pjgranieri"
+owners:
+  - "@tiagocastroferreira"
```
tiagolobocastro: was this meant to be me? (:

designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md
```diff
@@ -0,0 +1,121 @@
...
+* The exporter remains stateless; no new storage or CRDs are introduced.
+* Architectural flow:
+  1. Control-plane stays the authoritative source of node state.
+  2. Metrics-exporter polls REST, refreshes its cached state, and updates gauges.
+  3. Prometheus scrapes the exporter.
+  4. Grafana dashboards consume the same series without modification.
+
+### Risks and Mitigations
+
+* **REST unavailability:** When the control-plane REST service is unreachable the exporter could emit stale data. Mitigation: surface `mayastor_node_status_scrape_error` counters and keep the last known state with timestamps so operators can alert on scrape gaps.
```
tiagolobocastro, line 89: Adding an error counter makes sense I think;
Though not sure if we should keep the last known state forever, perhaps we could accommodate a few failures, but then we should stop returning the other gauges, wdyt?

tiagolobocastro (approved):
This is great, thank you for raising this

designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md
```diff
@@ -0,0 +1,121 @@
+---
+oep-number: TBC
```
tiagolobocastro: For this please create a github issue here, of type OEP and then `TBC` is just that issue number

niladrih (approved):

designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md
IyanekiB: Yes, that was supposed to be you! I should have the correct handle now

designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md
IyanekiB: Done, the OEP issue is created and the number has been updated in the header

designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md
IyanekiB: Yeah, that makes sense, an error counter plus a small retry window feels reasonable. Once that limit is hit, what’s your preferred behavior? Should we: 

- stop emitting the node-status gauges, 
- clear them, 
- or surface something like `NaN` until the next successful poll?

Abhinandan-Purkait: I think clearing it out makes sense rather than outdated view? Although I am not sure of the best practices in these kind of scenarios.

Abhinandan-Purkait (approved):
LGTM

designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md
tiagolobocastro: I think stopping the emitting of the gauges seems better than return potentially incorrect data; 
but also unsure what the best practices are here?

IyanekiB: That aligns with the approach we were considering. Our plan is to keep node-status gauges active only during a short retry window, and once that limit is exceeded, stop emitting them to avoid serving stale data. We can re-enable the gauges automatically once a successful poll occurs. Open to any alternative patterns if there’s a preferred best practice in the project.

tiagolobocastro: Yeah, this makes sense to me

tiagolobocastro:
Are we good to merge this @IyanekiB ?

IyanekiB:
Yes, we're all set from our side!

tiagolobocastro:
@mergify merge

Abhinandan-Purkait:
@mergify queue

Abhinandan-Purkait:
@mergify requeue

tiagolobocastro:
@mergifyio requeue

PR #4127, opened 2025-12-05
fix(upgrade-job): OpenEBS v3 to v4 upgrade etcd preUpgradeJob toggle

niladrih:
The strategy we follow when upgrading from OpenEBS v3 to OpenEBS v4 is to disable preUpgradeJob for etcd, then enable it after we've reached target version already.

The enable bit was bugged, fixing that.


tiagolobocastro (approved):

niladrih:
@mergify queue

Issue #4059, opened 2025-09-18
[OEP 4059]: RWX Block Volume Support in Mayastor for KubeVirt


tiagolobocastro:
### Enhancement Description

- Clear and concise enhancement description (one-liner): RWX Block Volume Support in Mayastor for KubeVirt
- OpenEBS Enhancement Proposal:
  - <https://github.com/openebs/openebs/pull/4060>
  - <https://github.com/openebs/openebs/blob/develop/designs/replicated-pv/mayastor/rwx-kubevirt.md>
- Discussion Link: 
- Primary contact (assignee): @tiagolobocastro
- Responsible OWNERs:
  - @tiagolobocastro
- Targeted milestone:
  - OpenEBS vx.y
- [ ] Targets
  - [ ] OEP update PR(s):
  - [ ] Code update PR(s):
    - <https://github.com/openebs/mayastor-control-plane/pull/1052>
    - <https://github.com/openebs/mayastor-control-plane/pull/1056>
  - [ ] E2E Test PR(s):
  - [ ] Docs update PR(s):

_Please keep this description up to date. This will help track the evolution of the enhancement efficiently._

yugchaudhari:
@tiagolobocastro I can see that PR is already merged for this. Is this feature ready to ship? In which version are you guys planning to release this?

tiagolobocastro:
Hey @yugchaudhari, this will be release in "experimental" mode for now as we've yet to do sufficient testing to verify it.

There's currently a potential situation where I don't don't what the best course of action - in case there's a network partition during the migration, and the source VM can still talk to the nexus but the target VM can't; in this case the HA logic can't really handle this and will most likely ping-pong the connections until it gives up.

If we want to do something a little smarter here we'd have to change the HA agent to be volume-aware rather than just node-path aware (it currently handles 2 path down detections individually).

itsgottabered:
Hi @tiagolobocastro,

I'd also like to know how/when this will be available for experimenting. We've got a significant footprint of vmware vms we're migrating to Kubevirt on RKE2/Talos and would like to leverage OpenEBS on some of our clusters with rwx block, so I'm happy to do some testing.

Cheers

tiagolobocastro:
Mayastor `v2.11.0` is released, including the [kubeVirt experimental support](https://openebs.io/docs/main/Solutioning/kubevirt)

PR #4145, opened 2026-01-12
volume attribute class OEP for local-pv lvm

rybas-dv:
I suggest we discuss it in this OEP the addition of VAC settings to local pv LVM. I have implemented in a local environment most of the functionality that I described in the OEP. I ask you to review the OEP and, if approved, I will prepare a PR with code.

### DIFF
--- a/designs/local-pv/lvm/volume_attributes_class.md
+++ b/designs/local-pv/lvm/volume_attributes_class.md
@@ -1,0 +1,203 @@
+---
+oep-number: OEP 4146
+title: VolumeAttributesClass support for local-pv lvm
+authors:
+  - "@rybas-dv"
+owners:
+  - "@tiagolobocastro"
+editor: TBD
+creation-date: 2026-01-11
+last-updated: 2026-01-21
+status: provisional
+---
+
+# VolumeAttributeClass Support for Enhanced PVC Configuration
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [User Stories](#user-stories)
+      * [Story 1: Dynamic Performance Tuning](#story-1-dynamic-performance-tuning)
+      * [Story 2: Multi-Tenant Isolation Policies](#story-2-multi-tenant-isolation-policies)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    * [Risks and Mitigations](#risks-and-mitigations)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+* [Drawbacks](#drawbacks)
+* [Alternatives](#alternatives)
+* [Infrastructure Needed](#infrastructure-needed)
+* [Testing](#testing)
+
+## Summary
+
+VolumeAttributeClass (VAC) is a Kubernetes resource that allows users to manage storage settings. This enhancement proposal introduces VAC support to OpenEBS local-pv lvm, allowing for more flexible and dynamic configuration of PersistentVolumeClaims (PVCs) with attributes such as iops and throughput through cgroup v2 io.max.
+
+The implementation extends the existing CSI driver controller side to interpret and apply VolumeAttributeClass parameters during volume provisioning and management, providing a standardized way to customize storage behavior without modifying StorageClass definitions.
+
+## Motivation
+
+Currently, OpenEBS users face limitations when they need to apply specific volume attributes that vary between applications or environments. The primary methods available are:
+
+1. **Multiple StorageClasses**: Creating separate StorageClass for each combination of attributes leads to proliferation and management overhead.
+2. **Specific parameters**: There is no way to dynamically adjust storage settings after creation, or it is not obvious.
+
+VolumeAttributeClass addresses these limitations by providing:
+- A standardized Kubernetes-native way to specify volume attributes.
+- Dynamic attribute application at PVC throughout the all life cycle.
+- Validation through OpenAPI schemas.
+- Better separation of concerns between storage provisioning (StorageClass/PVC) and volume characteristics (VolumeAttributeClass).
+
+### Goals
+
+1. Implement VolumeAttributeClass support in OpenEBS LVM local-pv.
+2. Extend CSI driver to process VAC parameters during all life cycle.
+3. Support backward compatibility with existing StorageClass parameters.
+4. Provide validation for VolumeAttributeClass parameters.
+5. Document common use cases and examples
+
+### Non-Goals
+
+1. Replacement of ioLimits functionality.
+2. Support for VolumeAttributeClass in all OpenEBS storage engines simultaneously
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Dynamic Performance Tuning
+As a database administrator, I want to specify different IOPS profiles for my PostgreSQL PVCs based on their workload importance, so that production databases get high-performance storage while development databases use cost-optimized storage, all using the same StorageClass.
+
+#### Story 2: Multi-Tenant Isolation Policies
+As a cluster administrator, I want to apply different QoS policies to PVCs, so that premium tenants get guaranteed performance while standard tenants use best-effort storage, using vac name in PVC to select appropriate VolumeAttributeClasses.
+
+### Implementation Details/Notes/Constraints
+Following a top-down approach, here are the required modifications:
+
+#### 1. LVMVolume CRD Modifications for VAC Processing
+When processing vac parameters, it is proposed to use unified parameters and directional ones, for example unified:
+```
+---
+apiVersion: storage.k8s.io/v1
+kind: VolumeAttributesClass
+metadata:
+  name: lvm-qos
+driverName: local.csi.openebs.io
+parameters:
+  qosIopsLimit: "200"
+  qosBandwithPerSec: "7000Mi"
+```
+or directional:
+```
+---
+apiVersion: storage.k8s.io/v1
+kind: VolumeAttributesClass
+metadata:
+  name: lvm-qos
+driverName: local.csi.openebs.io
+parameters:
+  qosIopsReadLimit: "100"
+  qosIopsWriteLimit: "200"
+  qosBandwithReadPerSec: "7000Mi"
+  qosBandwithWritePerSec: "6000Mi"
+```
+
+To interpret VAC parameters listed above such as `qos-iops-limit` or `qos-bandwidth-write-per-sec`, we need to extend the `LVMVolume` Custom Resource Definition. The logical approach is to add an `qos` field to the `LVMVolume.spec` to encapsulate performance-related parameters. Parameters in VAC can be used together, but must be checked for conflicts so that the resulting `LVMVolume.spec` resource will look like this:
+```
+apiVersion: local.openebs.io/v1alpha1
+kind: LVMVolume
+metadata:
+  namespace: lvm-localpv
+spec:
+  capacity: "1000341504"
+  qos:
+    readBPS: 7340032000    # 7GB/s read throughput limit
+    readIOPS: 10000        # 10K read IOPS limit
+    writeBPS: 3670016000   # 3.5GB/s write throughput limit
+    writeIOPS: 5000        # 5K write IOPS limit
+  ownerNodeID: kworker1
+  shared: "no"
+  thinProvision: "no"
+  vgPattern: ^lvmg$
+  volGroup: lvmg
+```
+
+Implementation details:
+- Add IO struct to LVMVolumeSpec in Go types.
+- Include validation for parameter boundaries.
+- Ensure backward compatibility (io field optional).
+- Add conversion logic between VAC parameters and LVMVolume io spec.
+
+#### 2. CSI Controller Modifications for VAC Processing
+The CSI driver controller requires the RPC_MODIFY_VOLUME capability to process VolumeAttributeClass parameters. Once declared, we can obtain VAC parameters through GetMutableParameters and convert them to the LVMVolume resource format.
+
+Implementation details:
+- Add IOParams holds collection of supported settings that can be configured in VolumeAttributesClass.
+- Add parses and validates VAC mutable parameters into IOParams.
+- Add IOParamsCreateVolume builds the initial VolumeIO spec for volume creation.
+- Add IOParamsModifyVolume applies mutable parameters to an existing VolumeIO spec.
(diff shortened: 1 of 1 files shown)


designs/local-pv/lvm/volume_attributes_class.md
```diff
@@ -0,0 +1,176 @@
...
+#### 1. LVMVolume CRD Modifications for VAC Processing
+To interpret VAC parameters such as `openebs.io/iops` and `openebs.io/throughput`, we need to extend the `LVMVolume` Custom Resource Definition. The logical approach is to add an `io` field to the `LVMVolume.spec` to encapsulate performance-related parameters.
+```
+apiVersion: local.openebs.io/v1alpha1
+kind: LVMVolume
+metadata:
+  namespace: lvm-localpv
+spec:
+  capacity: "1000341504"
+  io:
```
abhilashshetty04: Shall we call this qos instead?

designs/local-pv/lvm/volume_attributes_class.md
```diff
@@ -0,0 +1,176 @@
...
+
+1. Implement VolumeAttributeClass support in OpenEBS.
+2. Extend CSI driver to process VAC parameters during all life cycle.
+3. Support backward compatibility with existing StorageClass parameters.
+4. Provide validation for VolumeAttributeClass parameters.
+5. Document common use cases and examples
+
+### Non-Goals
+
+1. Replacement of ioLimits functionality.
```
abhilashshetty04, line 65: Does it mean once set user can't modify the parameters?

designs/local-pv/lvm/volume_attributes_class.md
```diff
@@ -0,0 +1,176 @@
...
+- Maintain idempotent operations for limit application.
+
+The scenarios themselves:
+- Initial Application during NodePublishVolume.
+- Runtime Updates via Controller Reconciliation.
+
+Implementation details:
+- Add logic to enqueue processed VAC parametrs in mgmt volume node agent.
+- Add logic which tries to converge to a desired state for the VAC parametrs on the node.
+- Add logic which applies the VAC parametrs on the node to the mounts.
```
abhilashshetty04:
```suggestion
- Add logic which applies the VAC parameters on the node to the mounts.
```

designs/local-pv/lvm/volume_attributes_class.md
```diff
@@ -0,0 +1,176 @@
...
+
+#### 4. Cgroup-based IO Limiting Modifications for VAC Processing
+The IO limits must be applied to both volumeMode: Block and volumeMode: Filesystem PVCs. This is achieved through cgroup v2 IO controller integration at the node level.
```
abhilashshetty04: Can you please elaborate bit more on cgroup v2 IO controller integration. What dependency changes it might involve.

abhilashshetty04:
@rybas-dv DCO is failing.  Latest commit is not signed.

Thanks for the proposal. Looks good overall.

Will it be a no-op if user updates VAC while volume is not published?

designs/local-pv/lvm/volume_attributes_class.md
```diff
@@ -0,0 +1,176 @@
+---
+oep-number: OEP 4145
```
tiagolobocastro: We need a tracking issue for this, it's not meant to be the PR number, ex: <https://github.com/openebs/openebs/issues/4059>

designs/local-pv/lvm/volume_attributes_class.md
```diff
@@ -0,0 +1,176 @@
...
+
+VolumeAttributeClass addresses these limitations by providing:
+- A standardized Kubernetes-native way to specify volume attributes.
+- Dynamic attribute application at PVC throughout the all life cycle.
+- Validation through OpenAPI schemas.
+- Better separation of concerns between storage provisioning (StorageClass/PVC) and volume characteristics (VolumeAttributeClass).
+
+### Goals
+
+1. Implement VolumeAttributeClass support in OpenEBS.
```
tiagolobocastro:
```suggestion
1. Implement VolumeAttributeClass support in OpenEBS LVM local-pv.
```

designs/local-pv/lvm/volume_attributes_class.md
tiagolobocastro: Would this naming be reflected in the VAC as well?
FYI We had an OEP for QOS for mayastor: <https://github.com/openebs/openebs/pull/4044/changes>
There we had a much more verbose naming on the VAC:
```
  qos-iops-limit: "5000"
  qos-bandwidth-per-sec: "200MiB"
```

designs/local-pv/lvm/volume_attributes_class.md
```diff
@@ -0,0 +1,176 @@
...
+## Graduation Criteria
+TODO
+
+## Implementation History
+The `Summary` and `Motivation` sections being merged signaling owner acceptance
+
+## Testing
+For testing, several stages must be implemented: 
+- Unit tests in the code that will cover possible options for configuring and parameterizing the vac. 
+- Integration tests that should show how the mounting itself and the application of parameters in /sys/fs/cgroup will work. There should be a scenario in which several pods are created with different numbers of replicas using volumeMode: Block and volumeMode: Filesystem, in which use different VAC.
```
tiagolobocastro: How to test that each volume has the correct configuration? Would we inspect the sysfs for each volume?

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: initially, I wanted to call it qos, but after diving into cgroup v2 io controller found io.cost.qos there. And if it is decided to use io.cost.qos in the future qos will be an intersection in the naming. Therefore, I suggested calling it by the name of the controller - io. This name is only for the structure in the api like so:
```
type VolumeInfo struct {
......................
	// IO contains VAC realization based on cgroup v2 IO Interface Files.
	// applied at Node side. Nil means "no explicit IO settings applied".
	IO *VolumeIO `json:"io,omitempty"`
..................
}

// VolumeIO now represents per-volume IO management based on cgroup v2 io.max.
// Can be supplemented with cgroup v2 io.cost.qos etc.
type VolumeIO struct {
	ReadIOPS  uint64 `json:"readIOPS,omitempty"`
	WriteIOPS uint64 `json:"writeIOPS,omitempty"`
	ReadBPS   uint64 `json:"readBPS,omitempty"`
	WriteBPS  uint64 `json:"writeBPS,omitempty"`
}
```

in vac it looks like this:
```
---
apiVersion: storage.k8s.io/v1
kind: VolumeAttributesClass
metadata:
  name: lvm-qos-low
driverName: local.csi.openebs.io
parameters:
  openebs.io/iops: "20"
  openebs.io/throughput: "2000Mi"
```

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: After installing VAC, the user can reconfigure the parameters; this is a VAC requirement. Here, I note that the VAC implementation does not intend to remove the existing ability to regulate IOPS via ioLimits from local-pv lvm. In the helm chart values.yaml it is used here:
```
lvmPlugin:
.......................................
  ioLimits:
    enabled: false
    containerRuntime: containerd
    readIopsPerGB: ""
    writeIopsPerGB: ""
```

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: For this we can either use pod fio to check how the result will show the restriction, or do cat /sys/fs/cgroup/.....pod..../io.max

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: Thanks for the help, added: 
https://github.com/openebs/openebs/issues/4146

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: thanks I fixed it

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: thanks I fixed it

designs/local-pv/lvm/volume_attributes_class.md
tiagolobocastro: Do we need the prefix `openebs.io/`?
Would be good if we could sync up with @urso and maybe name these the same across the different storage engines, to avoid confusion.

designs/local-pv/lvm/volume_attributes_class.md
tiagolobocastro: Testing the pod respects limits might be tricky, I'm ok with "cating" the sysfs

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: I suppose some kind of prefix is definitely needed to ensure validation.

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: expanded the information

designs/local-pv/lvm/volume_attributes_class.md
tiagolobocastro: isn't the vac for `local.csi.openebs.io` specifically? so should be clear it's "our" parameter?
the SC doesn't have require a prefix: <https://github.com/openebs/lvm-localpv/blob/develop/docs/storageclasses.md>

rybas-dv: As far as I understand, the VAC [documentation](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/#parameters) doesn't restrict the choice of parameter name. The choice of prefix/name is up to you. I'm simply saying that the prefix is ​​needed for validation. then I use the prefix `local.csi.openebs.io`?

abhilashshetty04: In the VAC docs I see the parameters does not have prefix :

https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/#resizer

abhilashshetty04: In the VAC docs I see the parameters does not have prefix :

https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/#resizer

rybas-dv: Yes, there's no requirement for a prefix in doc. But to validate that the user passed a parameter related to "our" CSI driver, I think the prefix should be the first check.

rybas-dv: So then in PR I prepare a solution without a prefix and instead of IO I do QoS?

rybas-dv: added changes according to what was discussed

tiagolobocastro (approved):

designs/local-pv/lvm/volume_attributes_class.md
```diff
@@ -0,0 +1,203 @@
...
+apiVersion: storage.k8s.io/v1
+kind: VolumeAttributesClass
+metadata:
+  name: lvm-qos
+driverName: local.csi.openebs.io
+parameters:
+  qos-iops-read-limit: "100"
+  qos-iops-write-limit: "200"
+  qos-bandwidth-read-per-sec: "7000Mi"
+  qos-bandwidth-write-per-sec: "6000Mi"
```
tiagolobocastro: Sorry, just hit me that these should probably be camel case (both here and in the other oep) for consistency?

```suggestion
  qosIopsReadLimit: "100"
  qosIopsReadLimit:: "200"
  qosBandwithReadPerSec: "7000Mi"
  qosBandwithWritePerSec: "6000Mi"
```

niladrih (approved):

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: Thanks, applied the changes)

rybas-dv:
I don't understand how a maystor appeared in my PR) I can't delete it because it doesn't exist.

Update:
during git rebase, I accidentally signed the previous extra commit. Fixed it, thanks.

tiagolobocastro:
> I don't understand how a maystor appeared in my PR) I can't delete it because it doesn't exist.

that's a git submodule, you probably committed it by accident;
you can fix it like this:
```
git submodule update --init --recursive --force
git add mayastor
git commit --amend -s
```

rybas-dv:
@tiagolobocastro please tell me what our next steps are?

tiagolobocastro:
We reviewed this during the maintainers meeting, and I think we still had a couple of doubts.
I'll wait for @Abhinandan-Purkait to add those here

tiagolobocastro:
@mergify squash

designs/local-pv/lvm/volume_attributes_class.md
```diff
@@ -0,0 +1,203 @@
...
+- Add parses and validates VAC mutable parameters into IOParams.
+- Add IOParamsCreateVolume builds the initial VolumeIO spec for volume creation.
+- Add IOParamsModifyVolume applies mutable parameters to an existing VolumeIO spec.
+- Implement the logic of ControllerModifyVolume
+
+#### 3. CSI Node Agent Modifications for VAC Processing
+The CSI driver node agent must apply VAC parameters to workloads in two scenarios. For both scenarios, it is partially inherent: 
+- Implement updateVol and syncVol handlers in node controller.
+- Watch for LVMVolume updates and reapply IO limits.
+- Handle volume remounts and pod restarts gracefully.
```
Abhinandan-Purkait, line 149: Yeah, I wanted to understand more on this point. What do you mean by remount here? Would you remount using the mount option remount or are you expecting a pod restart to do the remount?

Abhinandan-Purkait (approved):
LGTM

designs/local-pv/lvm/volume_attributes_class.md
rybas-dv: @Abhinandan-Purkait 
this means is that if the LVMVolume.spec.qos state changes or the pod is restarted, the vac must be applied(reconcile) to the /sys/fs/cgroup/....pod..../io.max using the information from the GetVolumeDevPath and mnt.GetMounts. This section doesn't mean that if the vac changes, the pod must be restarted to apply it.

rybas-dv: @Abhinandan-Purkait can i resolve this conversation?

Abhinandan-Purkait: Fair enough. Thanks!

tiagolobocastro:
@mergify queue

PR #4160, opened 2026-02-11
ci: trivy scan workflow

pchandra19:

### DIFF
--- a/.github/workflows/trivy-ci.yml
+++ b/.github/workflows/trivy-ci.yml
@@ -1,0 +1,119 @@
+name: Trivy Check CI
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: '0 3 1,15 * *'
+
+permissions:
+  contents: read
+  security-events: write
+  actions: read
+
+env:
+  CI: 1
+
+jobs:
+  helm-scan:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
+
+      - uses: cachix/install-nix-action@v31.3.0
+
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
+
+      - name: Helm dependency update
+        working-directory: charts
+        run: nix-shell shell.nix --run "helm dependency update ."
+
+      - name: Render Helm Templates
+        working-directory: charts
+        run: |
+          CHART_NAME=$(yq '.name' Chart.yaml)
+          RELEASE_NAME=${RELEASE_NAME:-$CHART_NAME}
+          nix-shell shell.nix --run \
+            "helm template \"$RELEASE_NAME\" . > rendered.yaml"
+
+      - name: Trivy Helm Misconfiguration Scan
+        uses: aquasecurity/trivy-action@0.34.2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          scan-type: config
+          scan-ref: charts/rendered.yaml
+          format: sarif
+          output: trivy-helm.sarif
+          severity: HIGH,CRITICAL
+
+      - name: Upload Helm SARIF
+        uses: github/codeql-action/upload-sarif@v4
+        with:
+          sarif_file: trivy-helm.sarif
+          category: trivy-helm
+
+  generate-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      images: ${{ steps.set-matrix.outputs.images }}
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
+          fetch-depth: 0
+
+      - uses: cachix/install-nix-action@v31.3.0
+
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./scripts/helm/shell.nix
+
+      - name: Generate image list
+        run: |
+          nix-shell ./scripts/helm/shell.nix --run "./scripts//helm/images.sh generate --dependency-update --exit-code"
+
+      - id: set-matrix
+        run: |
+          images=$(jq -R -s -c 'split("\n") | map(select(length > 0))' charts/images.txt)
+          echo "images=$images" >> "$GITHUB_OUTPUT"
+
+  image-scan:
+    needs: generate-matrix
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        image: ${{ fromJson(needs.generate-matrix.outputs.images) }}
+
+    steps:
+      - name: Trivy Image Scan
+        uses: aquasecurity/trivy-action@0.34.2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          scan-type: image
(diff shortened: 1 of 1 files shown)


.github/workflows/trivy-ci.yml
```diff
@@ -0,0 +1,124 @@
...
+          nix-shell ../shell.nix --run "helm dependency update ."
+
+      - name: Render Helm templates
+        working-directory: charts
+        run: |
+          CHART_NAME=$(yq '.name' Chart.yaml)
+          RELEASE_NAME=${RELEASE_NAME:-$CHART_NAME}
+          nix-shell ../shell.nix --run "
+            helm template "$RELEASE_NAME" . \
+              > rendered.yaml
```
tiagolobocastro: This doesn't include all images.
We have already chart/images.txt, can't we use that?

pchandra19: I initially thought of using images.txt but then was not sure whether we should scan for images mentioned in that file. There are other images as well apart from openebs. 

For openebs images we are implementing scans in individual repos already so wouldn't using images.txt be redundant ?

It's easy and straight forward to implement scan for images in image.txt , let me know if that's required and I can include it.

tiagolobocastro: > I initially thought of using images.txt but then was not sure whether we should scan for images mentioned in that file. There are other images as well apart from openebs.
>
> For openebs images we are implementing scans in individual repos already so wouldn't using images.txt be redundant ?

Good point, but that also applies to the helm templated images, there's other both 3rd party and images from our other repos?

That being said, we could include all images, since they're part of the installed package anyway?
CC @Abhinandan-Purkait @niladrih

pchandra19: Done.

.github/workflows/trivy-ci.yml
```diff
@@ -0,0 +1,166 @@
...
+      - name: Trivy Helm Misconfiguration Scan
+        working-directory: charts
+        run: |
+          nix-shell ../shell.nix --run "
+            trivy config \
+              --severity HIGH,CRITICAL \
+              --format sarif \
+              --output trivy-helm.sarif \
+              rendered.yaml
+          "
```
tiagolobocastro: All this is not needed now since we use images.txt?


```suggestion
```

.github/workflows/trivy-ci.yml
```diff
@@ -0,0 +1,166 @@
...
+          
+      - name: Upload Helm Sarif
+        uses: github/codeql-action/upload-sarif@v4
+        with:
+          sarif_file: charts/trivy-helm.sarif
+          category: trivy-helm
+
+      - name: Scan and upload container images from file
```
tiagolobocastro: This seems to be a lot more verbose compare to the changes on lvm for example, is this because of the input format?

.github/workflows/trivy-ci.yml
```diff
@@ -0,0 +1,166 @@
...
+
+          # Capture start time (used to identify images built by this run)
+          START_TS=$(date +%s)
+
+          # Extract tag from Chart.yaml
+          TAG=v$(yq '.appVersion' charts/Chart.yaml)
+          echo "Release tag: ${TAG}"
+
+          echo "== Build images =="
+          ./scripts/release.sh \
```
tiagolobocastro: Why do we need to build the images, can't we use whatever is on the registry already?

.github/workflows/trivy-ci.yml
pchandra19: We scan these
helm - charts
container images - ones present in images.txt
image - this is the upgrade image

All three are required.

tiagolobocastro: Why do we need to scan the helm chart, images.txt itself should suffice or am I missing something?
(the upgrade container image not being present in images.txt is a bug, which should be fixed separately CC @niladrih )

.github/workflows/trivy-ci.yml
pchandra19: Yes, there are certain limitations with trivy scanning multiple images with same category, so had to do it this way. There are 40 images in images.txt on which vuln scans are running

.github/workflows/trivy-ci.yml
pchandra19: How to know the tag to check and whether or not it has been previously scanned already. Hence we build and push the image locally, run the scan and upload reports. I think it is a simpler yet flexible approach which can run on any branch via dispatch.

.github/workflows/trivy-ci.yml
pchandra19: We need helm chart scan for misconfigurations, insecure configs, RBAC issues etc.

pchandra19: Also, I will scan the images present in images.txt only and remove the extra we do for upgrade image.

pchandra19:
I have fixed/optimized the workflow
- Removed setup trivy step and instead use trivy-action
- fixed recent issues with trivy - https://github.com/aquasecurity/trivy/discussions/10265 and likes
- Used a matrix based approach for image scans

.github/workflows/trivy-ci.yml
tiagolobocastro: > We need helm chart scan for misconfigurations, insecure configs, RBAC issues etc.

Ok, I wasn't aware trivy did that. Will that be flagged in the security alert as well?

If we want to leverage that then a simple helm template is not sufficient as it doesn't render everything, probably you'd want to render using those 2 lines [here](https://github.com/openebs/openebs/blob/41bcd772a411f27e400e8256f99ad99d5098079b/scripts/helm/images.sh#L235).
That covers all dependencies enabled, plus analytics and also the upgrade jobs, which otherwise don't get rendered

.github/workflows/trivy-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+          submodules: recursive
+          fetch-depth: 0
+
+      - uses: cachix/install-nix-action@v31.3.0
+
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
```
tiagolobocastro, line 32:
```suggestion
          nix-shell --pure --run "echo" ./charts/shell.nix
```

.github/workflows/trivy-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      - name: Helm dependency update
+        working-directory: charts
+        run: nix-shell ../shell.nix --run "helm dependency update ."
```
tiagolobocastro:
```suggestion
        run: nix-shell ./shell.nix --run "helm dependency update ."
```

.github/workflows/trivy-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      - name: Render Helm Templates
+        working-directory: charts
+        run: |
+          CHART_NAME=$(yq '.name' Chart.yaml)
+          RELEASE_NAME=${RELEASE_NAME:-$CHART_NAME}
+          nix-shell ../shell.nix --run \
```
tiagolobocastro: Use the shell.nix from the charts, don't need the rust compiler and all here

```suggestion
          nix-shell ./shell.nix --run \
```

tiagolobocastro (approved):

.github/workflows/trivy-ci.yml
pchandra19: Done.

.github/workflows/trivy-ci.yml
pchandra19: Fixed

.github/workflows/trivy-ci.yml
pchandra19: Fixed

.github/workflows/trivy-ci.yml
pchandra19: Fixed

pchandra19:
@mergify queue

pchandra19:
@mergify refresh

pchandra19:
@mergify queue

PR #4174, opened 2026-03-13
oep(4011): Add design discussion

urso:
Raid 0 proposal updates:

- Add Implementation section with HLD and test plan to the RAID0 DiskPool proposal
- Update CRD examples

What's included in the HLD
- Bdev layer stack diagram (base bdevs → RAID0 → crypto → LVS)
- RAID0 bdev state lifecycle (Configuring → Online → Offline, no degraded mode)
- DiskPool CRD changes (field format, validation rules, strip vs stripe naming)
- Compatibility analysis (partial rebuild, pool expansion, etcd persistence, perf stats)
- Test plan

### DIFF
--- a/designs/replicated-pv/mayastor/raid0.md
+++ b/designs/replicated-pv/mayastor/raid0.md
@@ -15,17 +15,26 @@ status: provisional
 
 ## Table of Contents
 
-* [Table of Contents](#table-of-contents)
-* [Summary](#summary)
-* [Motivation](#motivation)
-    * [Goals](#goals)
-    * [Non-Goals](#non-goals)
-* [Proposal](#proposal)
-    * [User Stories](#user-stories)
-    * [Workflow](#workflow)
-    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
-    * [Risks and Mitigations](#risks-and-mitigations)
-* [Testing](#testing)
+- [Table of Contents](#table-of-contents)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Workflow](#workflow)
+- [Implementation](#implementation)
+  - [Implementation Overview](#implementation-overview)
+  - [High Level Design](#high-level-design)
+    - [Component Overview](#component-overview)
+    - [Bdev Layer Stack](#bdev-layer-stack)
+    - [RAID0 Bdev State Lifecycle](#raid0-bdev-state-lifecycle)
+    - [Repository Structure and PR Order](#repository-structure-and-pr-order)
+    - [DiskPool CRD Changes](#diskpool-crd-changes)
+    - [Compatibility Considerations](#compatibility-considerations)
+    - [Design Decisions](#design-decisions)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Testing](#testing)
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


designs/replicated-pv/mayastor/raid0.md
```diff
@@ -94,52 +104,209 @@ metadata:
...
 
+### Test Plan
+
+#### io-engine Unit/Integration Tests
+
+- RAID0 bdev create and destroy lifecycle
+- RAID0 bdev properties: capacity equals sum of members, correct block size, driver is `"raid"`
+- I/O on RAID0 bdev: write, read, verify data integrity
+- Re-open: destroy and recreate RAID bdev, verify data persists
+- Error handling: missing children, duplicate names
```
tiagolobocastro, line 298: The ordering of the disks needs to be strict?
Does the spdk raid0 have an on-disk superblock, perhaps this might address the ordering issue?

designs/replicated-pv/mayastor/raid0.md
```diff
@@ -94,52 +104,209 @@ metadata:
...
+> Note: Other RAID levels behave differently — RAID1 can tolerate all-but-one
+> device failures, and RAID5F can tolerate one device failure.
+
+#### Repository Structure and PR Order
+
+The implementation spans four repositories:
+
+- **spdk-rs** — Expose SPDK `bdev_raid` module and `raid_bdev_*` FFI bindings.
+- **mayastor-dependencies** — Add RAID protobuf messages to the io-engine gRPC proto definitions.
+- **mayastor (io-engine)** — RAID bdev wrapper, LVS integration, gRPC handler updates. Depends on spdk-rs and mayastor-dependencies.
```
tiagolobocastro, line 243: Also the new probe api may require some tweaks for raid0, not sure

tiagolobocastro: Also we now export disk-errors, probably that also needs some tweaking to get stats from the devices under the raid bdev?

tiagolobocastro (approved):
Thanks @urso, this looks really good
Depending on other comments I'd suggest also setting this for implementable

abhilashshetty04 (approved):

designs/replicated-pv/mayastor/raid0.md
urso: Yes, the order must not change. It is tricky though. SPDK has support to create Raid devices with superblocks. But this requires stable UUIDs. SPDK will check the bdev.uuid value and compare it with the UUID stored in the superblock. It fails if the UUIDs mismatch. The problem is that UUIDs are not stable necessarily stable, unless users add the UUID to their diskpool configuration. This can break restarts.

To overcome the issue we could read the Raid0 superblock ourselves first. Find the UUID, add the alias and update the bdev it's UUID to match its superblock. Only limitation is that, if the explicitly configured UUID is not in line with the superblock, then the value from the superblock will be used, which might be a little confusing. As the raid implementation is agnostic to the underlying children volumes, we can't tell if the UUID field was generated or configured by the user.

In my current implementation we have the superblock checks disabled (SPDK will not write the superblock and we can't enable that feature once the raid is active). This means we are not allowed to reconfigure the diskpools once the raid is configured. Now with the order fixed it has been no problem for us. Alternatively to the UUID dance we could have agent-core complain/error if users change the configuration/order. On the other hand one might want to update a diskpool if a disk was actually replaced.

Although I'm in favor of supporting superblocks also to handle the order, we might consider to make this optional.

abhilashshetty04:
Hi @urso ,

Here is the OEP for diskpool device failure handling:

https://github.com/openebs/openebs/blob/develop/designs/replicated-pv/mayastor/pool-disk-failure-handling.md

niladrih (approved):

tiagolobocastro:
There was a few comments remaining, merging this for now, feel free to pick this up and make status implementable sometime

PR #4184, opened 2026-03-31
feat(4183): allow to disable Dynamic LocalPV provisioner in Helm chart

### LINKED ISSUE #4183: Allow to disable the dynamic-localpv provisioner in the Helm chart
## Description

The Helm chart forces installation of the dynamic-localpv provisioner, most likely because it is used to allocate e.g. etcd volumes. However, the admin may want to use another existing storage class for this purpose, and not enable the dynamic-localpv provisioner.

## Context

The admin wants to keep their Kubernetes installation minimal and avoid installing unnecessary components.

## Possible Solution

Add a condition to the sub chart so setting value `engines.local.dynamic.enabled=false` disables the provisioner.

jochenseeber:
Fixes #4183

### DIFF
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -93,6 +93,9 @@ engines:
       # Disabled by default since rawfile is not marked as stable yet.
       # -- Enable/Disable LocalPV Rawfile Storage Engine
       enabled: false
+    hostpath:
+      # -- Enable/Disable Dynamic LocalPV Provisioner
+      enabled: true
   replicated:
     mayastor:
       # -- Enable/Disable Replicated PV Mayastor Storage Engine
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -28,6 +28,7 @@ dependencies:
   - name: localpv-provisioner
     version: 4.5.0-develop
     repository: "https://openebs.github.io/dynamic-localpv-provisioner"
+    condition: engines.local.hostpath.enabled
   - name: zfs-localpv
     version: 2.10.0-develop
     repository: "https://openebs.github.io/zfs-localpv"
--- a/charts/templates/_validation.tpl
+++ b/charts/templates/_validation.tpl
@@ -1,0 +1,48 @@
+{{/* vim: set filetype=mustache: */}}
+
+{{/*
+Validate that localpv-dependent components are not configured to create
+StorageClasses when the Dynamic LocalPV Provisioner is disabled.
+
+Usage:
+  {{ include "openebs.validate.localpv" . }}
+*/}}
+{{- define "openebs.validate.localpv" -}}
+{{- if not .Values.engines.local.hostpath.enabled -}}
+  {{- if and .Values.loki.enabled .Values.loki.localpvScConfig.enabled -}}
+    {{- fail `
+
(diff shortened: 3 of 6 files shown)


charts/values.yaml
```diff
@@ -93,6 +93,9 @@ engines:
       # Disabled by default since rawfile is not marked as stable yet.
       # -- Enable/Disable LocalPV Rawfile Storage Engine
       enabled: false
+    dynamic:
```
tiagolobocastro: Should this be `hostpath`?
CC @niladrih @Abhinandan-Purkait

niladrih: I think it should be hostpath.

Abhinandan-Purkait: Yes, I think so

Abhinandan-Purkait: ~But why do we want to disable this?~

Abhinandan-Purkait (approved):

niladrih (approved):
I like it. I'm going to advise that we hold off merge till we verify the install behaviour manually. But, the implementation looks solid from what I can see.

charts/values.yaml
tiagolobocastro: @jochenseeber could you please update this to `hostpath` ?

jochenseeber: Will do. Give me until Monday.

tiagolobocastro: Sure @jochenseeber, no pressure, and thank you for contributing!

tiagolobocastro:
> I like it. I'm going to advise that we hold off merge till we verify the install behaviour manually. But, the implementation looks solid from what I can see.

@niladrih  have you checked this, are we good to merge?

niladrih (commented):
@jochenseeber -- One more comment. There's a pre-upgrade hook Job which is required for helm upgrades from v3.x to v4.y. That should be disabled if localpv-provisioner is disabled. Simply wrapping it in an `if` block should work.
https://github.com/openebs/openebs/blob/develop/charts/templates/pre-upgrade-hook.yaml

krishnaGajabi:
Just checking , should'nt this changes be in mayastor-extensions repo first?
cc @niladrih/ @tiagolobocastro / @Abhinandan-Purkait

tiagolobocastro:
> @jochenseeber -- One more comment. There's a pre-upgrade hook Job which is required for helm upgrades from v3.x to v4.y. That should be disabled if localpv-provisioner is disabled. Simply wrapping it in an `if` block should work. https://github.com/openebs/openebs/blob/develop/charts/templates/pre-upgrade-hook.yaml

@jochenseeber could you please take a look at this?

niladrih:
We cannot update the branch with a merge commit, as that doesn't carry a DCO.

niladrih:
@Mergifyio squash

Abhinandan-Purkait:
@mergifyio squash

Abhinandan-Purkait:
@mergifyio squash

tiagolobocastro:
@mergify squash

tiagolobocastro:
@mergify squash

tiagolobocastro:
@mergify squash

niladrih:
These changes are included #4211. The commit from this PR was moved to 4211 because the merge update doesn't carry a DCO on it. Closing this PR.

PR #4191, opened 2026-04-15
feat: Add helm global values

### LINKED ISSUE #4171: preUpgradeHook and .mayastor.preUpgradeHook imagePullSecrets key is not uniform with rest of chart
The helm value for setting the ImagePullSecrets for the preUpgradeHook and the mayastor preUpgradeHook uses a different syntax than what the .mayastor.image.pullSecrets key uses. They should be homogenised.

krishnaGajabi:
Fixes https://github.com/openebs/openebs/issues/4171

## Why is this PR required? What issue does it fix?

Currently, the Helm charts do not provide global variables for configuring common parameters such as the image registry, image pull secrets, image pull policy, and analytics settings.

When deploying the OpenEBS Helm charts, users often need to override these values across multiple subcharts and components. This makes it difficult to locate and configure all the required fields consistently. Providing global variables simplifies configuration and improves the user experience.

## What does this PR do?

This PR introduces the following global configuration variables:
```yaml
global:
  # -- Global override for container image registry
  imageRegistry: ""
  # -- Global image pull secrets (merged with local imagePullSecrets and not overridden)
  # - secret
  imagePullSecrets: []
  # -- Global override for image pull policy
  imagePullPolicy: ""
```

These variables allow users to configure common settings in a single place, which will then be applied across the chart components.

## Does this PR require any upgrade changes?

No.
This change is backward compatible and does not require any upgrade steps.

## Verification

The changes were verified using the following scenarios:

Rendered the templates using helm template to confirm correct value substitution.

Deployed the chart on a Kubernetes cluster with 3 worker nodes (kubeadm-based setup).

Verified that the global values are rendered and applied correctly across the components.

### DIFF
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -1,3 +1,12 @@
+global:
+  # -- Global override for container image registry
+  imageRegistry: ""
+  # -- Global image pull secrets (merged with local imagePullSecrets and not overridden)
+  # - secret
+  imagePullSecrets: []
+  # -- Global override for image pull policy
+  imagePullPolicy: ""
+
 openebs-crds:
   csi:
     volumeSnapshots:
@@ -9,6 +18,13 @@ openebs-crds:
 localpv-provisioner:
   rbac:
     create: true
+  globalImageRegistryOverride: true
+  helperPod:
+    image:
+      registry: "docker.io"
+  localpv:
+    image:
+      registry: "docker.io"
 
 # Refer to https://github.com/openebs/zfs-localpv/blob/v2.7.1/deploy/helm/charts/values.yaml for complete set of values.
 zfs-localpv:
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -38,7 +38,7 @@ dependencies:
     repository: "https://openebs.github.io/lvm-localpv"
     condition: engines.local.lvm.enabled
   - name: rawfile-localpv
-    version: 0.12.0
+    version: 0.14.0
     repository: "https://openebs.github.io/rawfile-localpv"
     condition: engines.local.rawfile.enabled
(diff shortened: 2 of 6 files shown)


krishnaGajabi:
CI Test are failing because we have set global.imageRegistry to empty string.On localPV hostpath chart it was set to docker.io. PR is dependent on hostpath changes to be merged first

Abhinandan-Purkait (approved):

charts/values.yaml
```diff
@@ -9,6 +18,13 @@ openebs-crds:
 localpv-provisioner:
   rbac:
     create: true
+  globalImageRegistryOverride: true
+  helperPod:
+    image:
+      registry: "docker.io"
+  localpv:
+    image:
+      registry: "docker.io"
```
niladrih, line 27: Are these required?

krishnaGajabi: Yes they are required, since we have set  globalImageRegsitryOverrride=true, on localpv helm chart it is false which means global is set and local is unset. we have to explicitly set local since global registry will be overriding on openebs chart

niladrih: But you'd have to also set the global to `""`. Otherwise, just setting the local won't remove the global from there, right?

krishnaGajabi: it will take the parent chart settings, on parent chart global ImageRegistry is set to "".

niladrih (approved):

niladrih:
@Mergifyio queue

niladrih:
@Mergifyio dequeue

niladrih:
@Mergifyio squash

niladrih:
@Mergifyio queue

PR #4193, opened 2026-04-22
fix(upgrade): add RBAC for OpenShift resources

niladrih:
Add create, delete, patch, list, get RBAC permissions for all security.openshift.io resources (OpenShift api group).

### DIFF
--- a/plugin/src/cli_utils/upgrade/k8s/resources.rs
+++ b/plugin/src/cli_utils/upgrade/k8s/resources.rs
@@ -216,6 +216,12 @@ pub(crate) fn upgrade_job_cluster_role(
                 verbs: vec!["create", "list", "delete", "get", "patch"].into_vec(),
                 ..Default::default()
             },
+            PolicyRule {
+                api_groups: Some(vec!["security.openshift.io"].into_vec()),
+                resources: Some(vec!["securitycontextconstraints"].into_vec()),
+                verbs: vec!["create", "list", "delete", "get", "patch"].into_vec(),
+                ..Default::default()
+            },
         ]),
         ..Default::default()
     }


plugin/src/cli_utils/upgrade/k8s/resources.rs
```diff
@@ -216,6 +216,12 @@ pub(crate) fn upgrade_job_cluster_role(
                 verbs: vec!["create", "list", "delete", "get", "patch"].into_vec(),
                 ..Default::default()
             },
+            PolicyRule {
+                api_groups: Some(vec!["security.openshift.io"].into_vec()),
+                resources: Some(vec!["*"].into_vec()),
```
tiagolobocastro: This seems like a pretty large hammer
Who's needing this? And also is it needed on mayastor?

niladrih: The helm command uses these. I'm also adding it on mayastor: [extensions_PR](https://github.com/openebs/mayastor-extensions/pull/843).
Based on my tests, we require at least 2 resources from that api group. Do you think we should thin this down? It'll require more testing.

Abhinandan-Purkait (approved):

plugin/src/cli_utils/upgrade/k8s/resources.rs
Abhinandan-Purkait: Yeah, it would be good to trim it down to specific needed resources.

niladrih: Thinned it down. Turns out we needed `securitycontextconstraints` and nothing else.

niladrih:
@mergify queue

PR #4203, opened 2026-05-09
feat(designs): add OEP for offline volume rebuild

yugchaudhari:
## Summary

Provisional OEP proposing automatic rebuild for unpublished volumes when replicas become unhealthy. The control plane temporarily creates a non-shared nexus, lets the existing rebuild engine restore the replicas, then tears the nexus down.

## Tracking issue

openebs/mayastor#1977

## Status

`provisional`: submitting per the merge-early strategy. Open design questions (state tracking, GC interaction, ReadOnly nexus semantics, concurrency limits) are documented in the OEP and intended to be iterated via follow-up PRs.

## Background

This addresses a gap observed in production while decommissioning DiskPools: only mounted/published volumes had their rebuilds triggered automatically; replicas of unpublished volumes (CDI golden images, cold backup volumes, etc.) were left degraded with no recovery path until manual intervention.

CC @tiagolobocastro @Abhinandan-Purkait, issue authors / discussion contributors.

### DIFF
--- a/designs/replicated-pv/mayastor/offline-volume-rebuild.md
+++ b/designs/replicated-pv/mayastor/offline-volume-rebuild.md
@@ -1,0 +1,239 @@
+---
+oep-number: OEP 1977
+title: Offline Volume Rebuild
+authors:
+  - "@yugchaudhari"
+owners:
+  - "@yugchaudhari"
+editor: TBD
+creation-date: 2026-05-09
+last-updated: 2026-05-09
+status: provisional
+see-also:
+  - https://github.com/openebs/mayastor/issues/1977
+---
+
+# Offline Volume Rebuild
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Implementation Details](#implementation-details)
+    - [State Tracking](#state-tracking)
+    - [Trigger Logic](#trigger-logic)
+    - [Rebuild Flow](#rebuild-flow)
+    - [Concurrent Operations](#concurrent-operations)
+    - [Garbage Collection Interaction](#garbage-collection-interaction)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Testing](#testing)
+
+## Summary
+
+Mayastor currently rebuilds replicas only when a volume is **published** (i.e. has an active nexus). When a volume is unpublished and one or more of its replicas become unhealthy, no automatic rebuild occurs because the hot-spare logic depends on a nexus existing.
+
+This OEP proposes an **offline volume rebuild** mechanism: when an unpublished volume is detected as `Degraded` for longer than a configurable threshold, the control plane temporarily creates a non-shared nexus, lets the existing rebuild engine restore the replicas, then tears the nexus down.
+
+## Motivation
+
+In production environments, many volumes spend most of their lifecycle in an unpublished state, typical examples:
+
+- CDI golden images (PVCs that are read once during machine image cloning, never mounted again)
+- Cold backup volumes
+- Volumes attached to stopped/scaled-down workloads
+- Volumes detached during node maintenance
+
+When a DiskPool is decommissioned, a node fails, or a disk is taken offline for maintenance, replicas of these unpublished volumes go into a degraded state. Because the current hot-spare logic gates on `volume_state.target.is_some()`, no automatic rebuild is triggered. Operators must either:
+
+1. Manually publish each affected volume to trigger a rebuild
+2. Accept silent data risk until a future client mounts the volume
+3. Build external tooling to detect and remediate
+
+This was hit recently in production while decommissioning DiskPools, only mounted/published volumes had their rebuilds triggered automatically; replicas of unpublished volumes (CDI golden images, etc.) were left degraded with no recovery path until manual intervention.
+
+### Goals
+
+- Detect degraded unpublished volumes via existing `NexusInfo` health signal
+- Automatically trigger a rebuild for such volumes after a configurable wait period
+- Reuse the existing rebuild engine and nexus lifecycle code as much as possible
+- Avoid impacting workload performance, bound concurrency separately from "regular" rebuilds
+- Handle concurrent publish/unpublish gracefully
+
+### Non-Goals
+
+- Replica-to-replica rebuild without a nexus (the existing rebuild engine runs inside the nexus)
+- Cross-node replica relocation (the temp nexus runs on the same node as a healthy replica)
+- New data-plane API changes, the io-engine already supports unshared nexus creation; only control plane changes are needed
+- Replacing or significantly changing the existing hot-spare logic for published volumes
+
+## Proposal
+
+### User Stories
+
+#### Story 1: DiskPool decommissioning
+
+A cluster operator wants to decommission a DiskPool to retire ageing hardware. They run `kubectl mayastor cordon pool ...` and wait for replicas to be evicted. With offline rebuild enabled, replicas of unpublished volumes are automatically rebuilt onto other pools without operator intervention. Without offline rebuild, the operator must either publish every affected volume manually or accept that those replicas will remain on the cordoned pool.
+
+#### Story 2: Node maintenance
+
+A node is taken offline for maintenance. After it returns, replicas of unpublished volumes that resided on the node are stale (`NexusInfo` reflects this). With offline rebuild, after the configured grace period these replicas are rebuilt automatically. The grace period prevents needless rebuild work for nodes that come back quickly.
+
+#### Story 3: Disk failure on a cold volume
+
+A physical disk hosting a replica of an unpublished volume fails. Without offline rebuild, the volume silently runs at reduced redundancy until the next publish. With offline rebuild, recovery is automatic.
+
+### Implementation Details
+
+#### State Tracking
+
+The temporary nexus used for offline rebuild needs to be tracked across control-plane restarts. Proposed approach: **reuse the existing `target_config` mechanism** with `share = Protocol::None`. The existing `published()` helper (`OperationGuardArc::<VolumeSpec>::published()`) returns `target().is_some()`, so an unshared target still counts as "published" from most code paths' perspective, which is desirable, because:
+
+- GC code that gates on "target exists" continues to protect replicas during rebuild
+- `health_info_id()` continues to point at the active nexus's `NexusInfo`
+- The existing nexus reconciler (`handle_faulted_children`) will work without modification
+
+A small flag (e.g. `target_config.target.share_state == OfflineRebuild`) can distinguish offline-rebuild targets from real publishes for the few places that need the distinction (CSI publish flow, telemetry).
+
+#### Trigger Logic
+
+A new `OfflineRebuildReconciler` is added to `controller/reconciler/volume/mod.rs` alongside `HotSpareReconciler` and `GarbageCollector`. Trigger conditions per volume:
+
+- `volume.policy.self_heal == true`
+- `volume.target().is_none()`: volume is currently unpublished
+- `volume_state.status == Degraded`, `online_clean_replicas < num_replicas` per `NexusInfo`
+- `health_info_id().is_some()`, there is data to rebuild from
+- Time since the volume entered the `Degraded` state exceeds the offline rebuild grace period
+
+If `current_replica_count < num_replicas` (a replica is missing entirely, not merely degraded), the urgent path applies, no grace period.
+
+#### Rebuild Flow
+
+1. Pick a target node via existing `target_node_candidate`. Prefer a node hosting a healthy replica to avoid cross-node copy.
+2. Build a `TargetConfig` with `share = Protocol::None` (unshared).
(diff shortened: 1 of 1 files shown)


designs/replicated-pv/mayastor/offline-volume-rebuild.md
```diff
@@ -0,0 +1,239 @@
...
+- Volumes attached to stopped/scaled-down workloads
+- Volumes detached during node maintenance
+
+When a DiskPool is decommissioned, a node fails, or a disk is taken offline for maintenance, replicas of these unpublished volumes go into a degraded state. Because the current hot-spare logic gates on `volume_state.target.is_some()`, no automatic rebuild is triggered. Operators must either:
+
+1. Manually publish each affected volume to trigger a rebuild
+2. Accept silent data risk until a future client mounts the volume
+3. Build external tooling to detect and remediate
+
+This was hit recently in production while decommissioning DiskPools, only mounted/published volumes had their rebuilds triggered automatically; replicas of unpublished volumes (CDI golden images, etc.) were left degraded with no recovery path until manual intervention.
```
tiagolobocastro: > while decommissioning DiskPools

How were you doing this btw, probably manually updating etcd?
We've merged offline diskpool and node deletion and will be release on v2.11 which should make this much easier.
Also online DiskPool drain will be the next step for these operations.

designs/replicated-pv/mayastor/offline-volume-rebuild.md
```diff
@@ -0,0 +1,239 @@
...
+- Automatically trigger a rebuild for such volumes after a configurable wait period
+- Reuse the existing rebuild engine and nexus lifecycle code as much as possible
+- Avoid impacting workload performance, bound concurrency separately from "regular" rebuilds
+- Handle concurrent publish/unpublish gracefully
+
+### Non-Goals
+
+- Replica-to-replica rebuild without a nexus (the existing rebuild engine runs inside the nexus)
+- Cross-node replica relocation (the temp nexus runs on the same node as a healthy replica)
+- New data-plane API changes, the io-engine already supports unshared nexus creation; only control plane changes are needed
```
tiagolobocastro: If we want to modify the nexus shutdown logic to avoid full rebuilds in case of a crash, then we may need to modify this, depending on the design/implementation.

Though perhaps we can just modify the nexus code, so that it will behave differently when not shared?

designs/replicated-pv/mayastor/offline-volume-rebuild.md
```diff
@@ -0,0 +1,239 @@
...
+- Replacing or significantly changing the existing hot-spare logic for published volumes
+
+## Proposal
+
+### User Stories
+
+#### Story 1: DiskPool decommissioning
+
+A cluster operator wants to decommission a DiskPool to retire ageing hardware. They run `kubectl mayastor cordon pool ...` and wait for replicas to be evicted. With offline rebuild enabled, replicas of unpublished volumes are automatically rebuilt onto other pools without operator intervention. Without offline rebuild, the operator must either publish every affected volume manually or accept that those replicas will remain on the cordoned pool.
```
tiagolobocastro: If the pool is cordoned the replicas won't be evicted automatically. Cordon will only stop new resources from being scheduled in the pool.

There should be a new pool operation, `drain`, which would then cordon and evict the replicas; perhaps `drain` can be built on top of what we're doing in this oep.

designs/replicated-pv/mayastor/offline-volume-rebuild.md
```diff
@@ -0,0 +1,239 @@
...
+
+#### State Tracking
+
+The temporary nexus used for offline rebuild needs to be tracked across control-plane restarts. Proposed approach: **reuse the existing `target_config` mechanism** with `share = Protocol::None`. The existing `published()` helper (`OperationGuardArc::<VolumeSpec>::published()`) returns `target().is_some()`, so an unshared target still counts as "published" from most code paths' perspective, which is desirable, because:
+
+- GC code that gates on "target exists" continues to protect replicas during rebuild
+- `health_info_id()` continues to point at the active nexus's `NexusInfo`
+- The existing nexus reconciler (`handle_faulted_children`) will work without modification
+
+A small flag (e.g. `target_config.target.share_state == OfflineRebuild`) can distinguish offline-rebuild targets from real publishes for the few places that need the distinction (CSI publish flow, telemetry).
```
tiagolobocastro: Need to think about this, I wonder if the nexus having share protocol None would be enough for what we need

designs/replicated-pv/mayastor/offline-volume-rebuild.md
```diff
@@ -0,0 +1,239 @@
...
+- `volume.target().is_none()`: volume is currently unpublished
+- `volume_state.status == Degraded`, `online_clean_replicas < num_replicas` per `NexusInfo`
+- `health_info_id().is_some()`, there is data to rebuild from
+- Time since the volume entered the `Degraded` state exceeds the offline rebuild grace period
+
+If `current_replica_count < num_replicas` (a replica is missing entirely, not merely degraded), the urgent path applies, no grace period.
+
+#### Rebuild Flow
+
+1. Pick a target node via existing `target_node_candidate`. Prefer a node hosting a healthy replica to avoid cross-node copy.
```
tiagolobocastro: or the node where the rebuilding replica lives on I guess?

designs/replicated-pv/mayastor/offline-volume-rebuild.md
```diff
@@ -0,0 +1,239 @@
...
+
+- Detect degraded unpublished volumes via existing `NexusInfo` health signal
+- Automatically trigger a rebuild for such volumes after a configurable wait period
+- Reuse the existing rebuild engine and nexus lifecycle code as much as possible
+- Avoid impacting workload performance, bound concurrency separately from "regular" rebuilds
+- Handle concurrent publish/unpublish gracefully
+
+### Non-Goals
+
+- Replica-to-replica rebuild without a nexus (the existing rebuild engine runs inside the nexus)
```
tiagolobocastro: We do have some generic-ish replica rebuild code, aimed at snapshot-chain rebuild, which is currently not used, and could potentially be used to rebuild replica->replica directly without a nexus, though using a nexus at least as a start would be easier.
Also nexus can be shared, thus allowing for publish of the volume with no disruption.

tiagolobocastro (approved):
Also I'd say we should have another knob to enable/disable offline rebuild, would allow us to disable it by default should we need to release before this is complete/fully tested for example.

Abhinandan-Purkait (approved):
This is great @yugchaudhari  Thanks! I would highlight two points:

1. Make sure the non-shared nexuses(the ones spawned for offline rebuild) are spread over mayastor nodes as much as possible to reduce the blast radius. The goal should be we should not pile up non-shared nexuses on one mayastor node.
2. Probably would be good to add a bdd to replicate a crashed mayastor node, maybe something in lines of killing the mayator process mid offline rebuilds or before nexus info was persisted properly(this might be tricky)

Overall LGTM.

designs/replicated-pv/mayastor/offline-volume-rebuild.md
yugchaudhari: Manually, and the process is uglier than just etcd edits. We were physically removing disks from baremetal nodes (PCI-remove for NVMe, sysfs offline for SCSI/SATA), then manually:

1. Created **trigger pods** that mount the affected unpublished PVCs just to force a nexus, so hot-spare logic would kick in and rebuild
2. Restarted `io-engine` on the faulted node to flush SPDK's in-memory pool/replica cache (which otherwise reports stale replicas as healthy)
3. Sometimes had to restart `io-engine` on the nexus-hosting node too, when the remote SPDK cache also kept stale replica state
4. Removed DiskPool finalizer + DiskPool CR
5. Deleted `PoolSpec` from etcd
6. Deleted orphan `ReplicaSpec` entries from etcd
7. Restarted `agent-core` to reconcile

(We wrote a runbook for this and ran it ~10 times across the cluster, happy to share if useful.)

So this OEP would eliminate steps 1 (the trigger pod hack), and combined with v2.11's offline DiskPool deletion + the upcoming online drain, the whole flow becomes a single `kubectl mayastor drain pool` command. Big win.

One thing worth flagging from our experience: the **SPDK in-memory cache** is the trickiest part. Even after the physical disk is gone, remote nexuses can keep reporting stale replicas as healthy until io-engine is restarted. The offline rebuild reconciler should not be confused by this, it should rely on `NexusInfo` (etcd-persisted) rather than the live SPDK replica state when deciding whether to trigger a rebuild. Worth calling out in the OEP.

designs/replicated-pv/mayastor/offline-volume-rebuild.md
yugchaudhari: Good point, keeping it in the nexus code (treat unshared differently for clean-shutdown semantics) is cleaner than introducing a new flag at the control-plane layer. Will update the OEP to reflect this direction.

designs/replicated-pv/mayastor/offline-volume-rebuild.md
yugchaudhari: You're right, my mistake, cordon stops new scheduling but doesn't evict. Updating the user story to use the proposed `drain` semantics. As you said, this OEP naturally enables `drain` for pools (evict + rebuild offline replicas).

designs/replicated-pv/mayastor/offline-volume-rebuild.md
yugchaudhari: Agreed, `share == None` alone is likely enough. Will remove the proposed extra flag from the OEP. The only place a distinction matters is preventing CSI publish racing with the temp nexus, which is already handled by promoting via `share_nexus` (your suggested approach for the concurrent-publish case).

designs/replicated-pv/mayastor/offline-volume-rebuild.md
yugchaudhari: Yes, "node where the rebuilding replica lives" is more precise. Updating to clarify we prefer nodes hosting one of the volume's replicas (any of them) so we avoid network copy.

yugchaudhari:
> Also I'd say we should have another knob to enable/disable offline rebuild, would allow us to disable it by default should we need to release before this is complete/fully tested for example.

Will add an `offline_rebuild_enabled` config knob (defaulting to **disabled** initially so a release can ship before this is fully tested, as you suggested).

yugchaudhari:
> This is great @yugchaudhari Thanks! I would highlight two points:
> 
> 1. Make sure the non-shared nexuses(the ones spawned for offline rebuild) are spread over mayastor nodes as much as possible to reduce the blast radius. The goal should be we should not pile up non-shared nexuses on one mayastor node.
> 2. Probably would be good to add a bdd to replicate a crashed mayastor node, maybe something in lines of killing the mayator process mid offline rebuilds or before nexus info was persisted properly(this might be tricky)
> 
> Overall LGTM.

1. Good catch, will add to the design: nexus placement should spread across mayastor nodes to reduce blast radius. Specifically, when picking the target node for a temp nexus, prefer nodes with the fewest existing temp/offline-rebuild nexuses among the volume's replica nodes.

2. Will add: BDD scenarios for (a) mayastor process killed mid-rebuild, (b) crash before `NexusInfo` is persisted. Need to verify the recovery path leaves no orphaned temp nexuses and resumes rebuild on next reconcile.

designs/replicated-pv/mayastor/offline-volume-rebuild.md
yugchaudhari: Good point, I'll soften the Non-Goals wording. Replica-to-replica rebuild (via the snapshot-chain rebuild code path) isn't impossible, it's just not the route we're taking initially. Will rephrase as "out of scope for initial implementation; may be revisited if the nexus-based approach proves too heavyweight."

Re: shareable nexus enabling no-disruption publish: that's exactly the concurrent-publish path in the design ("promote via `share_nexus`"), so we're aligned.

designs/replicated-pv/mayastor/offline-volume-rebuild.md
tiagolobocastro: > One thing worth flagging from our experience: the SPDK in-memory cache is the trickiest part. Even after the physical disk is gone, remote nexuses can keep reporting stale replicas as healthy until io-engine is restarted.

This should also be improved now, if device is un-plugged pool will be hot-removed, and nexus child will also be hot-removed.
I didn't have any hot-unpluggable devices to test with, so had to use like: `echo 1 | sudo tee /sys/class/block/sdb/device/delete` and ublk devices for testing.

tiagolobocastro:
I think we can probably merge this as is, and do further updates via PR if you wish.
Just let me know, thank you!

designs/replicated-pv/mayastor/offline-volume-rebuild.md
yugchaudhari: Good to hear! That hot-removal fix probably saves a lot of operator pain. The trigger-pod hack was very cumbersome. Will update the OEP risk row to reflect that this is now handled at the data-plane level (so the offline-rebuild reconciler can trust live state once on a sufficiently new mayastor version).

yugchaudhari:
> I think we can probably merge this as is, and do further updates via PR if you wish. Just let me know, thank you!

Let's merge as-is. The pending OEP update for the Non-Goals (data-plane changes wording) and the SPDK cache risk row are small wording fixes, I'll send them in a follow-up PR. Thanks @tiagolobocastro @Abhinandan-Purkait for the thorough review!

tiagolobocastro:
@mergify queue

yugchaudhari:
@tiagolobocastro PR got dequeued because of unresolved comments, I resolved them now. can you please re-queue?

niladrih (approved):

Abhinandan-Purkait:
@mergifyio queue

PR #4229, opened 2026-06-09
oep(4074): add initial oep for mayastor trim/unmap support

tiagolobocastro:

### DIFF
--- a/designs/replicated-pv/mayastor/unmap.md
+++ b/designs/replicated-pv/mayastor/unmap.md
@@ -1,0 +1,556 @@
+---
+oep-number: OEP 4074
+title: Mayastor trim/unmap support
+authors:
+  - "@tiagolobocastro"
+owners:
+  - "@tiagolobocastro"
+editor: TBD
+creation-date: 09/06/2026
+last-updated: 09/06/2026
+status: provisional
+---
+
+# OEP: Mayastor trim/unmap support
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Gap 1: Rebuild Does Not Propagate Unmaps](#gap-1-rebuild-does-not-propagate-unmaps)
+  - [Gap 2: Synchronous UNMAP Blocks the Reactor](#gap-2-synchronous-unmap-blocks-the-reactor)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Space reclamation after rebuild](#story-1-space-reclamation-after-rebuild)
+    - [Story 2: Flash device efficiency](#story-2-flash-device-efficiency)
+    - [Story 3: No reactor stalls during fstrim](#story-3-no-reactor-stalls-during-fstrim)
+  - [Part 1: Cluster Release During Rebuild](#part-1-cluster-release-during-rebuild)
+    - [Constraints](#constraints)
+    - [Rebuild Initialization](#rebuild-initialization)
+    - [Full Rebuild](#full-rebuild)
+    - [Partial Rebuild](#partial-rebuild)
+    - [Optional Optimization: IO-log Operation Provenance](#optional-optimization-io-log-operation-provenance)
+    - [Replica Geometry Discovery (Control-Plane vs Data-Plane)](#replica-geometry-discovery-control-plane-vs-data-plane)
+    - [Component Change Summary](#component-change-summary)
+  - [Part 2: Async UNMAP Passthrough to Backend Devices](#part-2-async-unmap-passthrough-to-backend-devices)
+    - [Why TRIM Matters for Flash Devices](#why-trim-matters-for-flash-devices)
+    - [Current State](#current-state)
+    - [Async UNMAP for uring bdev](#async-unmap-for-uring-bdev)
+    - [Async UNMAP for aio bdev](#async-unmap-for-aio-bdev)
+    - [Blobstore Integration](#blobstore-integration)
+    - [Component Change Summary](#component-change-summary-1)
+  - [Interaction Between Parts](#interaction-between-parts)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Graduation Criteria](#graduation-criteria)
+- [Test Plan](#test-plan)
+  - [Part 1: Rebuild UNMAP Tests](#part-1-rebuild-unmap-tests)
+  - [Part 2: Backend UNMAP Tests](#part-2-backend-unmap-tests)
+- [Implementation History](#implementation-history)
+- [Alternatives](#alternatives)
+  - [Alternative 1: Modify SPDK Blobstore to Support Sub-Cluster UNMAPs for ZEROES](#alternative-1-modify-spdk-blobstore-to-support-sub-cluster-unmaps-for-zeroes)
+- [References](#references)
+
+---
+
+## Summary
+
+This OEP proposes two complementary features to bring full UNMAP/DEALLOC support
+to Mayastor:
+
+1. **Cluster release during rebuild**: Correctly propagating unmapped (deallocated)
+   clusters from source to destination replica during both full and partial nexus
+   rebuilds, so that thin-provisioned space is actually reclaimed on rebuilt
+   replicas.
+
+2. **Async UNMAP passthrough to backend devices**: Forwarding UNMAP/TRIM commands
+   asynchronously to underlying aio/uring bdevs, enabling flash devices to reclaim
+   blocks and maintain write performance without blocking the SPDK reactor thread.
+
+Also of note, there's currently a bug where unmapped clusters would not be correctly
+reported, see fix [here](https://github.com/openebs/mayastor/pull/1994).
+
+---
+
+## Motivation
+
+Mayastor already supports UNMAP/DEALLOC commands at the nexus level (controlled
+by the `--bs-cluster-unmap` CLI flag). When a filesystem such as XFS issues
+`fstrim` or punches holes, the corresponding blobstore clusters are freed on
+each replica. This allows thin-provisioned storage pools to reclaim space as
+users delete files.
+
+However, two significant gaps remain:
+
+### Gap 1: Rebuild Does Not Propagate Unmaps
+
+When a replica is rebuilt (either after a brief outage triggering a partial
+rebuild, or when a new replica is added triggering a full rebuild), unmapped
+clusters on the source are **not correctly propagated** to the destination.
+
+The root cause is a granularity mismatch: the SPDK blobstore only deallocates
+a cluster when it receives an UNMAP that is **exactly cluster-sized and
+cluster-aligned**, as enforced in `blobstore.c`:
+
+```c
+if (spdk_blob_is_thin_provisioned(blob) && is_allocated &&
+    bs_io_units_per_cluster(blob) == length && g_cluster_unmap_enabled) {
+```
+
+The rebuild operates in fixed `SEGMENT_SIZE` (64 KiB) chunks. Even when a source
+read returns "unallocated", the resulting 64 KiB UNMAP sent to the destination
+does not satisfy the above condition (cluster size is typically 1MiB+), so the
+destination cluster remains allocated with zeroed data. The result is that
+rebuilt replicas consume more space than the source, defeating the purpose of
+thin provisioning.
+
+### Gap 2: Synchronous UNMAP Blocks the Reactor
+
+The SPDK aio and uring bdev modules handle UNMAP by calling `fallocate()` with
+`FALLOC_FL_PUNCH_HOLE` (for files) or `ioctl(BLKDISCARD)` (for block devices)
+**synchronously and inline** on the SPDK reactor thread. These syscalls can
+block for tens to hundreds of milliseconds, stalling all other I/O on that
+reactor core. This can cause:
+
+- I/O timeout escalation
+- Nexus health check failures
+- Cascading performance degradation under `fstrim` workloads
+
+A previous PR ([openebs/spdk#11](https://github.com/openebs/spdk/pull/11))
+implemented async UNMAP for the aio bdev but was later reverted. This OEP
+proposes reviving that work and bringing it upstream directly.
+
+### Goals
+
+- Rebuilt replicas must have the **same cluster allocation state** as the source,
+  including properly freed (unmapped) clusters.
+- UNMAP commands to aio/uring bdevs must be handled **asynchronously** without
+  blocking the SPDK reactor thread.
+- Flash/NVMe devices must receive TRIM notifications, enabling FTL garbage
+  collection, reduced write amplification, and sustained write performance.
+- Thin-provisioned storage pools must reclaim space end-to-end: from filesystem
+  delete → nexus UNMAP → blobstore cluster free → backend device TRIM.
+
+### Non-Goals
+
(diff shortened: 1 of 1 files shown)


patrickdk77:
## Rebuild thin-propagation couples Part 1 and Part 2 (it needs WRITE_ZEROES, not UNMAP)

While prototyping the Part 2 async passthrough, I hit a dependency on Part 1 that seems worth surfacing early: it changes what "correct" looks like for the backend ops, and it affects the capability model both parts share.

### The core constraint

A rebuild has to leave the destination replica byte-for-byte identical to the source. For an unmapped (thin) cluster the source reads back zeros, so after the rebuild propagates that hole the destination must read back zeros for the same range too.

Plain UNMAP (`BLKDISCARD`) does not guarantee that. Discard is only a hint; whether a post-discard read returns zeros is device-dependent, and the kernel removed `discard_zeroes_data` precisely because it could not be relied on. If a rebuild propagated a hole with UNMAP, on many devices the destination would read stale data where the source reads zeros, and the two replicas would silently diverge.

The operation that is safe for rebuild is `WRITE_ZEROES` (`BLKZEROOUT`): it guarantees zeros on read-back and deallocates where the device supports it. So the two paths want different ops:

- data-path trim (`fstrim`): UNMAP (discard) for reclaim, no read-back requirement
- rebuild thin-propagation: WRITE_ZEROES for guaranteed zeros, with reclaim where supported

### Implication: UNMAP and WRITE_ZEROES have to be gated independently, per device

Because the two paths need different ops, the backend has to advertise UNMAP and WRITE_ZEROES independently, based on the device's actual queue limits (`discard_max_bytes` vs `write_zeroes_max_bytes`), rather than flipping them together. A discard-only device (`discard_max_bytes > 0`, `write_zeroes_max_bytes == 0`, common on plain NVMe/dm) can serve data-path trim, but it cannot be a thin-rebuild target via unmap; Part 1 has to fall back to writing real zeros there.

WRITE_ZEROES looks nearly redundant for data-path trim on its own, but it is load-bearing for Part 1. Gating the two ops together, or shipping UNMAP-only, would make thin rebuild incorrect on mixed-capability hardware.

### Where the zero-guarantee comes from splits along blobstore vs backend

- Blobstore/lvol replicas (the primary target): freeing a cluster at the lvol layer already makes it read back as zeros, independent of the backend device. There, rebuild propagation is a cluster-free at the blob level and backend UNMAP passthrough is purely physical reclaim underneath; read-back correctness never depends on the device op.
- aio-backed replicas (raw LV / zvol / file): there is no blobstore returning zeros, so the device itself has to provide the guarantee, i.e. WRITE_ZEROES. Here the backend op is correctness-critical, not just reclaim.

So Part 1's implementation, and the role Part 2 plays in it, split along the same blobstore-vs-backend line the OEP already draws.

### A rebuild-protocol shape that keeps this simple

The direction I'm leaning for Part 1 keeps the source as the authority and the destination capability-aware:

- the source walks its own allocation map and, per rebuild segment, sends either data (allocated) or an "unmapped" marker (freed)
- the destination applies each segment against its own capability: WRITE_ZEROES where the range must read back zero, degrading to a plain zero-write where the device lacks it

That seems to fold several of the listed open questions together. "Replica geometry" becomes the source's live allocation map read inline during the copy, with no separate round-trip, and the IO-log WRITE-vs-UNMAP provenance is largely sidestepped because the source reports current state rather than replaying history. The per-device capability decision is the same probe Part 2 uses.

### Questions

1. Does propagating holes via WRITE_ZEROES (rather than UNMAP) for read-back safety match your intent for Part 1? Or was the blobstore cluster-free path specifically in mind, where the lvol layer provides the zero-guarantee and the device op is reclaim-only?
2. Given the coupling, does it make sense to keep WRITE_ZEROES in the Part 2 backend work even though trim alone barely exercises it?
3. Enablement: a flag and the probe seem complementary rather than either/or. My inclination is a node-level flag (per io-engine / per host) as a coarse off-switch, so an admin can disable backend passthrough on a problematic machine without reconfiguring individual volumes or replicas, while the per-device probe handles capability detection automatically underneath it. Does that match what you had in mind for the CLI flag, and should it default enabled (probe-driven, admin opts out) or default disabled (explicit opt-in)?

tiagolobocastro:
It is true that unmap to the backend may not necessarily ensure zeroes are read, though I'm not sure if we'd require a `byte-for-byte` match per replica on rebuild for unmapped regions, as they are unmapped?

And yes we can achieve the zero read on the cluster via the blobstore itself.

> ### Questions
> 1. Does propagating holes via WRITE_ZEROES (rather than UNMAP) for read-back safety match your intent for Part 1? Or was the blobstore cluster-free path specifically in mind, where the lvol layer provides the zero-guarantee and the device op is reclaim-only?

To clarify, IIRC it was not my idea to propose having them as separate things, it was more just to split the work domains if you will.
Though to answer your question, I did not intend to provide zero-guarantee on part 1, the main intention is to free up unused space; If an application wants to guarantee zeroes then it should write zeroes rather than send unmap.

> 2. Given the coupling, does it make sense to keep WRITE_ZEROES in the Part 2 backend work even though trim alone barely exercises it?

Part 2 is intended to take better care of [flash devices](https://github.com/openebs/openebs/blob/39e61757cd2de8af7280a36ed1bde2cc3a039a7b/designs/replicated-pv/mayastor/unmap.md#why-trim-matters-for-flash-devices).

> 3. Enablement: a flag and the probe seem complementary rather than either/or. My inclination is a node-level flag (per io-engine / per host) as a coarse off-switch, so an admin can disable backend passthrough on a problematic machine without reconfiguring individual volumes or replicas, while the per-device probe handles capability detection automatically underneath it. Does that match what you had in mind for the CLI flag, and should it default enabled (probe-driven, admin opts out) or default disabled (explicit opt-in)?

The flag was there because even though spdk enabled unmap in the blobstore the nexus did not handle it properly so we wanted to keep it disabled until we tackled this; but IIRC the flag was never exposed via helm, it's always disabled.

---

I'd be interested to know what your requirements for unmap are, and if you need zero-guarantee?

patrickdk77:
I wasn't clear, there are two rebuild paths and my concern is only with the aio/uring devices.

**Blobstore/lvol: agreed, nothing to change.** Rebuild reads with
`UNWRITTEN_READ_FAIL`, blobstore returns `UNWRITTEN_BLOCK` on holes, the copier
skips them, and unallocated reads back as zeros on both sides. No byte-for-byte
match needed: all is good.

**aio/io_uring thin backends (ZFS zvol + thin LVM), this is the part I am
looking at.** A plain `preadv`/`io_uring_prep_readv` never returns
`UNWRITTEN_BLOCK`, so the skip never fires and rebuild copies every segment,
the space the data-path trim just reclaimed re-inflates on the next rebuild.

But both backends zero unprovisioned reads and discard-free (dm-thin on a loop
device: fresh LV reads zeros at `data% 0.00`; write 8 MiB -> `1.56`; `blkdiscard`
-> back to `0.00` and reads zeros; zvols behave the same). So rebuild can detect
holes by content: read the segment, and if it's all zeros, propagate a
discard/UNMAP to the target (frees it and makes it read zero) instead of writing
zeros. Fresh target can just skip; partial/in-place must discard to clear stale
data. It's not about zeros-as-data -- it's that these backends re-inflate on
rebuild unless we propagate the hole.

Scope:
- blobstore/lvol -- already thin-aware.
- ZFS zvol + thin LVM -- content zero-detect + discard-propagate in rebuild (new,
  separate from the blobstore skip).
- thick LVM -- fully allocated, out of Part 1; trim there is flash-care passthrough (Part 2).

Do you want the aio/uring thin path in Part 1, or as a follow-up after Part 2?

tiagolobocastro:
> **aio/io_uring thin backends (ZFS zvol + thin LVM), this is the part I am looking at.** A plain `preadv`/`io_uring_prep_readv` never returns `UNWRITTEN_BLOCK`, so the skip never fires and rebuild copies every segment, the space the data-path trim just reclaimed re-inflates on the next rebuild.

But the skip doesn't come from the backend device, only comes from the nexus child/replica from the blobstore.

The part2 is about being able to issue unmap/discard to devices and not detect holes from them.

>     * blobstore/lvol -- already thin-aware.
> 
>     * ZFS zvol + thin LVM -- content zero-detect + discard-propagate in rebuild (new,
>       separate from the blobstore skip).
> 
>     * thick LVM -- fully allocated, out of Part 1; trim there is flash-care passthrough (Part 2).

Do you mean ZFS/LVM as the child/replicas? We did start some initial work to support LVM lvols as replicas directly but that effort never progressed, so that is currently not possible.
Only blobstore is currently supported usecase.

patrickdk77:
> Do you mean ZFS/LVM as the child/replicas? We did start some initial work to
> support LVM lvols as replicas directly but that effort never progressed, so
> that is currently not possible.

Yes, ZFS zvols and LVM LVs as replicas directly. That is where our viewpoints
differ: I have been working from my tree where that already exists.

I added ZFS support and it is fully functional in my repos. The LVM path came
along with it. My goal was never to write this OEP, it was to make unmap work
for ZFS, and once that was in place there was no reason not to cover LVM as
well, since it is the same problem in the same shape. I was testing the unmap
work against LVM regardless, so I have most likely filled in the LVM-replica
gap you are describing as a side effect of getting ZFS working.

That is also why my framing has been aio/uring-first throughout. I only came
across this OEP when I went looking to upstream what I had, and saw it was
scoped to blobstore. Blobstore is the gap in my work rather than the other way
around, so I reconciled my side against the OEP. That is why the aio/uring thin
path reads like an add-on in this thread when on my end it was the starting
point.

tiagolobocastro:
> Yes, ZFS zvols and LVM LVs as replicas directly. That is where our viewpoints differ: I have been working from my tree where that already exists.

Ah, ok now I understand :)
 
> I added ZFS support and it is fully functional in my repos. The LVM path came along with it. My goal was never to write this OEP, it was to make unmap work for ZFS, and once that was in place there was no reason not to cover LVM as well, since it is the same problem in the same shape. I was testing the unmap work against LVM regardless, so I have most likely filled in the LVM-replica gap you are describing as a side effect of getting ZFS working.

That's awesome, would you be looking into upstreaming that?

> That is also why my framing has been aio/uring-first throughout. I only came across this OEP when I went looking to upstream what I had, and saw it was scoped to blobstore. Blobstore is the gap in my work rather than the other way around, so I reconciled my side against the OEP. That is why the aio/uring thin path reads like an add-on in this thread when on my end it was the starting point.

Right, I wasn't considering this use-case at all.
So you need both things:
1. what is my part2, being able to send unmap/discard via aio/uring
2. extra which is also detect unmapped areas

So you're detecting this by reading and checking if it's all zeroes? Are you doing this locally or on the nexus?
And you'd have to use similar approach where you have to read chunks of the largest-size allocation-unit right? Otherwise again if destination allocation-unit is larger then it wouldn't discard? (unless you keep reading until non-zero or some trick like that)

patrickdk77:
> That's awesome, would you be looking into upstreaming that?

That is the plan, and it is why I turned up on this OEP instead of opening
something separately.

The end goal is a working ZFS backend, checksum-validated data, that handles
unmap requests. Upstreaming that as one lump would be miserable to review, and
splitting it naively would be a flood of PRs, so I broke it into stacked
pieces. Unmap is the logical first step: the ZFS work depends on it, and there
was already an open request for exactly that here. Filling this request, with
the unmap pieces I need alongside it, seemed better than running a parallel
track. The LVM and ZFS replica items then follow as another PR or two on top.

Concretely the trim/unmap stack is three PRs across three repos, in order:

- openebs/spdk: async UNMAP/WRITE_ZEROES for the aio and uring bdevs, each op
  advertised independently per device off its queue limits
- openebs/spdk-rs: libspdk pin bump to that commit
- openebs/mayastor: the io-engine side, a `fallocate=` opt-in on `aio://` URIs
  plus the per-device capability probe

The ZFS backend is a separate stack on top (dependencies, io-engine,
control-plane), and the LVM replica items fall out of the same work.

> So you need both things:
> * what is my part2, being able to send unmap/discard via aio/uring
> * extra which is also detect unmapped areas

Exactly those two, and that split is how I staged it. Part 2 is built and
validated. Detection was my next todo item, but I stopped short of starting it
to align with the OEP and the project direction first. If the intent here is
different from what I had in mind then I would rather refocus now than build it
out and discover afterwards that it does not fit Part 1. That is what I was
fishing for with the earlier questions.

> So you're detecting this by reading and checking if it's all zeroes? Are you
> doing this locally or on the nexus?

Read-and-test-for-zero is the direction I sketched, but none of it is written
yet, so take it as intent rather than a description. Local versus nexus is part
of what I would be working out first. The two shapes as I see them:

- On the nexus: the rebuild already reads every segment, so the test rides on
  work that is happening anyway. The nexus would need to know the destination's
  granularity to pick a test size.
- On the replica: the source could report "unmapped" without sending data,
  which is the same shape as the existing blobstore skip, where the source read
  returns an unwritten status and the copier skips the segment. Whether that
  can ride the existing read path as-is for aio-backed children is one of the
  things I would need to establish.

If you have a preference between those, that is worth knowing before I spend
the time on it.

> And you'd have to use similar approach where you have to read chunks of the
> largest-size allocation-unit right? Otherwise again if destination
> allocation-unit is larger then it wouldn't discard?

Right, and this is a genuine gap in what I have today: nothing in my current
code looks at `discard_granularity`, the capability probe reads only
`discard_max_bytes` and `write_zeroes_max_bytes`. If the destination's
allocation unit is larger than the chunk being tested, the discard can free
nothing at all, silently. So detection has to work in units of the
destination's granularity rather than the rebuild segment size or the source's
geometry, which means the destination's capability has to be known before the
source decides what to probe. That is another reason I want the shape agreed
before writing it.

tiagolobocastro:
> Read-and-test-for-zero is the direction I sketched, but none of it is written yet, so take it as intent rather than a description. Local versus nexus is part of what I would be working out first. The two shapes as I see them:
> 
>     * On the nexus: the rebuild already reads every segment, so the test rides on
>       work that is happening anyway. The nexus would need to know the destination's
>       granularity to pick a test size.
> 
>     * On the replica: the source could report "unmapped" without sending data,
>       which is the same shape as the existing blobstore skip, where the source read
>       returns an unwritten status and the copier skips the segment. Whether that
>       can ride the existing read path as-is for aio-backed children is one of the
>       things I would need to establish.
> 
> 
> If you have a preference between those, that is worth knowing before I spend the time on it.

hmm potentially doing this locally may be more "optimal" as we can avoid sending the data on the wire if it's all zeroes?

> > And you'd have to use similar approach where you have to read chunks of the
> > largest-size allocation-unit right? Otherwise again if destination
> > allocation-unit is larger then it wouldn't discard?
> 
> Right, and this is a genuine gap in what I have today: nothing in my current code looks at `discard_granularity`, the capability probe reads only `discard_max_bytes` and `write_zeroes_max_bytes`. If the destination's allocation unit is larger than the chunk being tested, the discard can free nothing at all, silently. So detection has to work in units of the destination's granularity rather than the rebuild segment size or the source's geometry, which means the destination's capability has to be known before the source decides what to probe.

Also the clusters should be aligned, though this may already be guaranteed as long we use same offsets for all replicas.

> That is another reason I want the shape agreed before writing it.

I would suggest creating an OEP for mayastor on top of zfs/lvm and we can discuss this there
We have our [community](https://github.com/openebs/openebs#community) [monthly meeting](https://zoom-lfx.platform.linuxfoundation.org/meeting/99475226462?password=f57455ef-fed1-49b1-a38b-110dab2b42cc) tomorrow - it'd be awesome if you could join and talk through your use-case and all

niladrih (approved):

Issue #4245, opened 2026-06-22
[OEP 4245]: Drain support for Mayastor DiskPool


abhilashshetty04:
**Enhancement Description**

- Clear and concise enhancement description (one-liner): Drain support on Mayastor DiskPool
- OpenEBS Enhancement Proposal:
- Discussion Link:
- Primary contact (assignee): @abhilashshetty04 
- Responsible OWNERs: @abhilashshetty04 
- Targeted milestone:
  - OpenEBS vx.y
- [x] Targets
  - [x] OEP update PR(s):
    - [https://github.com/openebs/openebs/pull/4248]
  - [ ] Code update PR(s):
  - [ ] E2E Test PR(s):
  - [ ] Docs update PR(s):

Please keep this description up to date. This will help track the evolution of the enhancement efficiently.

yugchaudhari:
Raising a small thing from openebs/mayastor-control-plane#1200 so it is not lost once drain implementation starts. Discussed with @abhilashshetty04 there.

The offline-rebuild reconciler has a grace period before it rebuilds a degraded unpublished volume. openebs/mayastor-control-plane#1161 added a bypass for the case where the volume is degraded because a replica was added that the last publish never saw, since there is nothing transient to wait out there. The bypass needs two things:

- a replica in the topology that is absent from the persisted `NexusInfo`
- `num_replicas > NexusInfo.children.len()`

A drain move satisfies the first but not the second. This design keeps `num_replicas` as the user configured it and introduces `effective_replica = num_replicas + N` for the count reconcilers, so the extra copy is not pruned as excess. Since that gate reads `num_replicas`, an over-replicated volume still looks like "3 wanted, 3 last published", and a drained pool's replicas would sit out the full grace period before moving.

The design already has this reconciler in scope, in the note about bypassing the `offline_rebuild_enabled()` flag for marker-driven moves. This gate is probably one more item on that list: either teach `has_scale_up_rebuild_target` about `effective_replica`, or let the `replica_move` marker bypass the grace period directly, alongside the `self_heal` and feature-flag bypasses already described.

Not a problem today, since nothing sets the marker yet. Flagging it now because the symptom would be a slow drain rather than a failure, which is the kind of thing that is easy to miss in testing.

PR #4248, opened 2026-06-23
oep(4245) for mayastor diskpool drain

abhilashshetty04:
Related issue:

https://github.com/openebs/openebs/issues/4245

### DIFF
--- a/designs/replicated-pv/mayastor/pool-drain.md
+++ b/designs/replicated-pv/mayastor/pool-drain.md
@@ -1,0 +1,995 @@
+---
+oep-number: OEP 4245
+title: mayastor-pool-drain
+authors:
+  - "@abhilashshetty04"
+owners:
+  - "@abhilashshetty04"
+editor: "@abhilashshetty04"
+creation-date: 2026-06-23
+last-updated: 2026-07-20
+status: provisional
+---
+
+# Online Pool Drain
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+- [Drawbacks](#drawbacks)
+- [Future Improvements](#future-improvements)
+- [Testing](#testing)
+
+## Summary
+
+Pool Drain allows administrators to move all storage resources off a DiskPool before decommissioning
+a pool or node, while the pool (and its node) stay **online** and continue serving I/O throughout.
+When an user drains a pool, every volume replica residing on the pool is migrated to eligible
+pools while respecting topology constraints; once the pool's replica allocation reaches zero (or all
+targeted resources are evacuated) the pool transitions to a `Drained` state.
+
+The feature is modeled end-to-end on the existing **node drain** and reuses the existing volume
+scale-up / scale-down and replica rebuild mechanisms. Migration preserves full volume redundancy by
+default by over-replicating a volume by one for the duration of each rebuild rather than the
+faster-but-degrading approach of removing a replica and letting a hot-spare rebuild from a reduced
+copy set. The operation is exposed through a new Pool Drain API across the stack — REST, gRPC, and
+the `kubectl mayastor` plugin — and is driven asynchronously by a dedicated reconciler and a
+persisted drain state machine.
+
+Users interact with it via `kubectl mayastor drain pool <id>` (and `PUT /pools/{id}/drain`). The
+feature includes abort support, progress visibility, a dry-run pre-flight analysis, a cluster-wide cap
+on concurrent pool drain operations, an optional forced-eviction policy for replicas that have no
+valid destination, and an explicit policy for handling snapshots left behind on a drained pool.
+
+## Motivation
+
+Today OpenEBS Mayastor supports volume scale-up, volume scale-down, and replica rebuilds, and the
+control plane can **cordon** a pool (block new replicas) and **drain a node** (move volume targets
+off a node). However, there is no mechanism to evacuate an entire pool — to move all existing
+replicas off one pool to other pools so the pool can be retired or serviced. This feature builds on
+the existing scale-up/scale-down and rebuild capabilities to provide that.
+
+This is required for administrators wishing to:
+
+- Decommission storage
+- Replace hardware
+- Remove a node
+- Rebalance capacity
+
+Pool Drain automates this process while ensuring volume availability is maintained, existing rebuilds
+are not disrupted, topology constraints are preserved, and the pool reaches zero replica allocation.
+
+### Goals
+
+- Drain a pool while it stays **online** by default, with **no data unavailability** for the volumes whose
+  replicas are migrated.
+- Support evacuation of every volume scenario: single-replica published, multi-replica, unpublished,
+  and degraded volumes.
+- Provide an explicit, user-chosen **snapshot policy** for snapshots that remain on a drained
+  pool.
+- Provide **abort** support to cancel a queued or in-progress drain.
+- Provide **progress visibility** for a drain in flight.
+- Provide a **dry-run** pre-flight analysis that reports what a drain would do without mutating state.
+- Rate-limit concurrent pool drains **cluster-wide** to avoid oversubscribing rebuild bandwidth.
+- Optionally **force-evict** replicas that have no valid destination after user specified Duration elapses
+  so a drain cannot stall indefinitely — never evicting a volume's only online copy.
+
+### Non-Goals
+
+- **Snapshot evacuation** (moving snapshots with their replicas to another pool) is out of scope for
+  Phase 1. Phase 1 only moves replicas; snapshots are either left behind or destroyed per the chosen
+  policy. True snapshot evacuation is deferred to Phase 2 which will commence after Phase 1 is code complete.
+- A **per-node** rebuild concurrency cap is out of scope (left as an open question); only
+  cluster-wide and per-pool concurrency limits are introduced.
+
+## Proposal
+
+A drain request lands the pool in a persisted `Queued` state and immediately self-cordons the pool
+(blocking new replicas, snapshots, and restores from landing on it). A dedicated reconciler on pool
+module will promote queued pools to `Draining` in FIFO order, up to a configurable `--pool-drain-limit`.
+A dedicated pool-drain reconciler then migrates every replica off each draining pool, one
+rebuild at a time. It sets ReplicaMoveConfig in VolumeSpec, Thus triggering a Full rebuild.
+When the rebuild completes, associated volume replica from the `Draining` pool is removed. When
+the pool's replica allocation reaches zero, the pool transitions to `Drained` (or `PartiallyDrained`
+when snapshots are intentionally left behind).
+
+The drain is driven by a `DrainPhase` state machine, with its data split across the pool spec and the
+pool state. The drain request and user policy are recorded in a **`DrainSpec`** carried on its own new
+`PoolSpec.drain_spec` field (separate from the existing `cordon_drain` user-cordon field); the
+current `DrainPhase` and progress metrics live in a **`DrainProgress`** on the pool state
+(`PoolState.drain_runtime.drain_progress`). The `DrainSpec` holds the self-cordon applied at admission plus a
+**`DrainArgs`** with the user policy, which holds:
+
+- `request_timestamp` — when the drain was applied; used for FIFO promotion out of the queue.
+- `accept_snapshot_loss` / `drain_replicas_only` — the user-chosen snapshot policy flags.
+- `unsafe_rebuild_otherwise_evict` — the forced-eviction policy: an optional grace period after which
+  a replica that has no placement candidate is force-evicted. `None` disables forced eviction;
+  `Some(d)` force-evicts after waiting `d`. The value must be **strictly positive** — `Some(0)` is
+  **rejected at admission with an error**; immediate eviction is a separate, explicit option
+  (`unsafe_evict`) rather than a zero grace on this field. To disable forced eviction the user simply
+  omits the flag.
+- `unsafe_evict` — a boolean that skips the safe over-replicate flow and evicts replicas directly,
+  accepting a transient degraded rebuild; never evicts a volume's only `Online` copy. Exposed on
+  REST/gRPC and, in the plugin, as a **hidden** flag (registered but omitted from `--help` and the
+  generated docs) so it is reachable by power users who know it but not discoverable by casual CLI use.
+
+`DrainProgress` carries the current `DrainPhase` plus a set of baseline metrics captured once at queue
+time and persisted to etcd (keyed by pool id):
+
+- `initial_replica_count` — number of replicas on the pool when the drain was admitted.
+- `initial_snapshot_count` — number of replica snapshots on the pool at admission.
+- `initial_allocation` — capacity allocated on the pool at admission (bytes).
+- `initial_committed` — capacity committed (sum of all replica size) on the pool at admission.
+
+These baseline values are immutable for the life of the drain and must be persisted (the live counts
+alone cannot tell how much has drained); current replica/snapshot/allocation counts are read live
+from the data plane on each progress query and compared against the baseline to report how far the
+drain has progressed. The `DrainPhase` state machine has the states:
+
+- `Queued` — admitted, self-cordoned, awaiting a concurrency slot. The pool with the longest time
+  elapsed in the queue is picked first.
+- `Draining` — actively migrating replicas.
+- `Drained` — replica allocation reached zero (terminal).
+- `PartiallyDrained` — replicas evacuated, snapshots intentionally left behind (reported when
+  allocation cannot reach zero because snapshots remain).
+- `Aborted` — a transient cleanup phase entered when a drain is cancelled. `drain_spec` is **retained**
+  (with `phase = Aborted`) throughout cleanup — it is *not* cleared up front — so the pool stays
+  discoverable and self-cordoned while a dedicated cleanup reconciler undoes the drain's changes on
+  affected volumes (if any). Only once cleanup completes is `PoolSpec.drain_spec` cleared (which
+  removes the drain's self-cordon with it) along with the `DrainProgress` record, while any user
+  cordon in `cordon_drain` is left untouched. Unlike `Drained`, this is not a retained terminal
+  state — once cleanup completes the drain record is gone.
+
+**Pre-requisites:** the pool must be `Online` and must not already be in a `Queued`, `Draining`, or
+`Aborted` state. Blocking on `Aborted` too means a fresh drain cannot start until the previous
+drain's cleanup has fully torn down (`drain_spec` deleted), which serializes the two and prevents a
+new drain's `replica_move` markers from being confused with the aborting drain's leftovers.
+
+### User Stories
+
+#### Story 1 — Decommission a node's storage
+
+An administrator needs to remove a node from the cluster. This feature will be tied with node drain
+feature but not in the scope of Phase 1. Administrator can drain each pool on the node with
+`kubectl mayastor drain pool <id>`. Each pool's replicas are migrated to other pools elsewhere in the
+cluster with full redundancy maintained throughout. They poll `kubectl mayastor get-drain-progress pool <id>`
+and watch the replica count fall to zero, at which point the pool reports `Drained` and the node can be safely removed.
+
+#### Story 2 — Drain a pool holding snapshots
+
+An administrator drains a pool that holds replica snapshots. Because snapshots cannot be migrated in
+Phase 1, the request is **rejected** unless they pick a policy: `--drain-replicas-only` (move
+replicas, leave snapshots in place; pool ends `PartiallyDrained` and stays online serving them) or
+`--accept-snapshot-loss` (move replicas, then destroy the remaining snapshots; pool allocation
+reaches zero).
+
+#### Story 3 — Abort a drain in progress
+
+An administrator starts a drain, then changes their mind. They run `kubectl mayastor uncordon
+pool <id> --drain` (there is no separate abort command — cancelling a drain is just uncordoning it
+with the drain scope). Already-evacuated replicas stay put, in-flight unrelated
+rebuilds are untouched, and the drain's own half-rebuilt spare replicas are unwound. Once that cleanup
+finishes the drain record (`DrainSpec`/`DrainProgress`) is cleared — which also lifts the drain's
+self-cordon — while any cordon the user applied themselves stays in place, so the pool returns to
+exactly the cordon state it had before the drain.
+
+### Implementation Details/Notes/Constraints
+
+#### Evacuation strategy: over-replicate vs fast-remove
+
+For a multi-replica volume we could simply remove the replica from the draining pool, which would
+trigger a full rebuild elsewhere instantly — pool allocation would drop quickly and the pool would
+reach `Drained` fast. The drawback is that we lose a data copy and the volume runs **Degraded** for
+the duration of the rebuild, which is an availability/durability regression.
+
+Instead, drain uses a **scale-up / scale-down** technique: add a new replica on an eligible pool,
+wait for its rebuild to complete, then scale down by removing the replica on the draining pool. The
+volume keeps its full redundancy the whole time, at the cost of a slower drain (each move waits for a
+rebuild) and one extra replica's worth of transient space. This *slower-but-never-degraded* trade-off
+is the core design choice; the fast-remove path is explicitly rejected for published volumes.
+
+Because the scale-up adds a replica *above* the volume's configured replica count, the drain must be
+able to tell that the extra replica is one it created. This is tracked as a managed `replica_move`
+marker on the volume's runtime metadata (`VolumeSpec.metadata.runtime`), persisted across restarts
+via the pool's `PoolDrainRecord` (see below), initialized when a drain-move begins to start
+Full rebuild (over-replicate) and once rebuild completes and spare child is Online, `replica_move`
+is cleared to kick start scale-down procedure. Existing replica-count-specific reconcilers may need changes
+to cooperate with this marker.
+
+#### Drain workflow and state machine
+
+Pool drain is an asynchronous operation carried out by reconcilers, like most async operations in the
+core agent. When the user issues a drain, the pool's `DrainPhase` is set to `Queued` and the pool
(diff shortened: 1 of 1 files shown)


designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,624 @@
...
+  no placement candidate is force-evicted. `None` disables forced eviction; `Some(d)` force-evicts
+  after waiting `d`. If user passed it as Some(0),  Should we not perform over-replication way of drain and
+  evict without attempting scale-up?
+
+`DrainProgress` carries the current `DrainPhase` plus a set of baseline metrics captured once at queue
+time and persisted to etcd (keyed by pool id):
+
+- `initial_replica_count` — number of replicas on the pool when the drain was admitted.
+- `initial_snapshot_count` — number of replica snapshots on the pool at admission.
+- `initial_allocation` — capacity allocated on the pool at admission (bytes).
```
tiagolobocastro, line 127: what is the purpose of this?

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,624 @@
...
+pool state. The drain request and user policy are recorded in a **`DrainSpec`** carried on the
+pool spec as a cordon/drain variant (`PoolSpec.cordon_drain` → `CordonDrainState::PoolDrain(DrainSpec)`);
+the current `DrainPhase` and progress metrics live in a **`DrainProgress`** on the pool state
+(`PoolState.drain_progress`). The `DrainSpec` holds:
+
+- `request_timestamp` — when the drain was applied; used for FIFO promotion out of the queue.
+- `allow_snapshot_loss` / `only_drain_replicas` — the user-chosen snapshot policy flags.
+- `evict_after` — the forced-eviction policy: an optional grace period after which a replica that has
+  no placement candidate is force-evicted. `None` disables forced eviction; `Some(d)` force-evicts
+  after waiting `d`. If user passed it as Some(0),  Should we not perform over-replication way of drain and
```
tiagolobocastro: we could validate that 0 is not a valid parameter?

Though I'm wondering here how this parameter should be calculated; I was initially thinking we'd just compare a particular replica drain with the time elapsed from the start, though what happens if there's a temporary blip after?

Also something worth considering, is prefer to drain replicas which can be rebuilt in a safe way, leaving the others for the end (in case another rebuild target turns up) ?

designs/replicated-pv/mayastor/pool-drain.md
tiagolobocastro: from the text below is to `tell how much has drained`
In that case we should not record allocation, we should record commitment! because allocation may change when existing volumes write more data, commitment will track replicas being removed from the pool
Fine to track both if you still want allocation

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,624 @@
+---
+oep-number: OEP 4245
+title: mayastor-pool-drain
+authors:
+  - "@abhilashshetty04"
+owners:
+  - "@abhilashshetty04"
+editor: "@abhilashshetty04"
+creation-date: 2026-06-23
+last-updated: 2026-06-23
```
tiagolobocastro: missing provisional state

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,624 @@
...
+from the data plane on each progress query and compared against the baseline to report how far the
+drain has progressed. The `DrainPhase` state machine has the states:
+
+- `Queued` — admitted, self-cordoned, awaiting a concurrency slot. The pool with the longest time
+  elapsed in the queue is picked first.
+- `Draining` — actively migrating replicas.
+- `Drained` — replica allocation reached zero (terminal).
+- `PartiallyDrained` — replicas evacuated, snapshots intentionally left behind (reported when
+  allocation cannot reach zero because snapshots remain).
+- `Aborted` — a transient cleanup phase entered when a drain is cancelled. It undoes the drain's
```
tiagolobocastro: I think we have to be clear what we mean here, it doesn't undo a drain in a volume, it only cancels the rebuilds

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,624 @@
...
+**Pre-requisites:** the pool must be `Online` and must not already be in a `Queued`/`Draining` state.
+
+### User Stories
+
+#### Story 1 — Decommission a node's storage
+
+An administrator needs to remove a node from the cluster. This feature will be tied with node drain
+feature but not in the scope of Phase 1. Administrator can drain each pool on the node with
+`kubectl mayastor drain pool <id>`. Each pool's replicas are migrated to other pools elsewhere in the
+cluster with full redundancy maintained throughout. They poll `kubectl mayastor drain-progress pool <id>`
```
tiagolobocastro: `drain-progress` is not a verb

what about we use the same command `kubectl mayastor drain pool <id>` ?

in a related matter, could be interesting to add a `--wait` on drain to wait there, like K8s

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,624 @@
...
+
+An administrator drains a pool that holds replica snapshots. Because snapshots cannot be migrated in
+Phase 1, the request is **rejected** unless they pick a policy: `--only-drain-replicas` (move
+replicas, leave snapshots in place; pool ends `PartiallyDrained` and stays online serving them) or
+`--accept-snapshot-loss` (move replicas, then destroy the remaining snapshots; pool allocation
+reaches zero).
+
+#### Story 3 — Abort a drain in progress
+
+An administrator starts a drain, then changes their mind. They run `kubectl mayastor abort-drain
```
tiagolobocastro: `abort-drain` feels a little odd, though not sure if I can suggest anything better

`kubectl-mayastor abort drain`
?

@Abhinandan-Purkait @niladrih any suggestions

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,624 @@
...
+that has waited longest. The reconciler's goal is to bring the pool's current replica count and
+current capacity allocation to zero; when that happens the pool is marked `Drained`. If snapshots
+remain (and the policy left them in place), the pool is marked `PartiallyDrained` since allocation
+will not reach zero. If user issues abort drain command on a Pool which is in `Queued` or `Draining`
+state, then the `DrainPhase` is set to `Aborted`.
+
+#### Drain lifecycle and policy flags
+
+The per-replica evacuation flow below answers "how is *one* replica moved". At the pool level, the
+three policy flags (`--only-drain-replicas`, `--accept-snapshot-loss`, `--evict-after`) operate at two
```
tiagolobocastro: `--only-drain-replicas`

use verb first, ie `--drain-replicas-only` or perhaps switch around?
`--skip-snapshots` ?

designs/replicated-pv/mayastor/pool-drain.md
tiagolobocastro: Ah for nodes I think we also have `kubectl-mayastor get drain nodes` ?

designs/replicated-pv/mayastor/pool-drain.md
tiagolobocastro: For the rebuild/eviction behavior, I wonder if a strategy would fit, similar to K8s

Like `--strategy={--evict,--rebuild-then-evict,--try-rebuild-otherwise-evict}` ?

Or perhaps even more like K8s: `--strategy={recreate,rebuild,rebuild-recreate}`

CC @Abhinandan-Purkait @niladrih

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,625 @@
...
+pool spec as a cordon/drain variant (`PoolSpec.cordon_drain` → `CordonDrainState::PoolDrain(DrainSpec)`);
+the current `DrainPhase` and progress metrics live in a **`DrainProgress`** on the pool state
+(`PoolState.drain_progress`). The `DrainSpec` holds:
+
+- `request_timestamp` — when the drain was applied; used for FIFO promotion out of the queue.
+- `allow_snapshot_loss` / `only_drain_replicas` — the user-chosen snapshot policy flags.
+- `evict_after` — the forced-eviction policy: an optional grace period after which a replica that has
+  no placement candidate is force-evicted. `None` disables forced eviction; `Some(d)` force-evicts
+  after waiting `d`. If user passed it as Some(0),  Should we not perform over-replication way of drain and
+  evict without attempting scale-up?
```
abhilashshetty04: I was thinking we can `evict-first` without `over-replicating` if its set to 0. We will not remove child if its last Online replica. wdyt? @tiagolobocastro

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Fair point that allocation may grow post cordon if new write come in and clusters are allocated. It's okay if it grows also. Shall we keep both commitment and allocation ? Gives more information on overall pool usage. We will add `commitment` in current metrics also.

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: We remove `TempReplicaConfig` on VolumeSpec which basically removes `over-replicated` child if its `Degraded`. We undo drain's change on VolumeSpec.  There may be chance that we have already removed Replica from `Draining` pool if `over-replicated` child is rebuilt fully.

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: >> @tiagolobocastro , in a related matter, could be interesting to add a --wait on drain to wait there, like K8s
 Do you mean we don't Queue the Pool until it elapses `--wait `duration?
 
`get-drain-progress `works? Its a verb

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: >> kubectl-mayastor abort drain

We may need a resource in cmd to suggest the resource type. We may support node `abort-drain` too when we integrate this with node drain.

Abhinandan-Purkait: Hmm, isn't abort for drain uncordon? I mean that's the k8s way. Might not be applicable for us. Maybe we can extend the mayastor pool uncordon with a flag?

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Losing some reschedule attempts due to temporary blip seems harmless. Otherwise, we may need some state (also persist it) to track number of failed attempts in finding placement candidate? cc @tiagolobocastro

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,625 @@
...
+
+#### Dry-run
+
+The user can set a dry-run flag on the drain request. The service runs the same candidate
+selection the real drain would use for every replica and returns an analysis — total replicas,
+snapshot count, degraded count, and the number of replicas with no valid destination broken down by
+cause (insufficient free space vs. topology/anti-affinity constraints) — **without** queuing,
+migrating any replica, or involving the reconciler.
+
+To make the analysis meaningful, the dry-run **self-cordons the pool** before evaluating candidates,
```
abhilashshetty04: We will not self-cordon here.

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: --unsafe-rebuild-otherwise-evict
--unsafe-evict (not exposed to UX)

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,625 @@
...
+
+The feature is modeled end-to-end on the existing **node drain** and reuses the existing volume
+scale-up / scale-down and replica rebuild mechanisms. Migration preserves full volume redundancy by
+default by over-replicating a volume by one for the duration of each rebuild rather than the
+faster-but-degrading approach of removing a replica and letting a hot-spare rebuild from a reduced
+copy set. The operation is exposed through a new Pool Drain API across the stack — REST, gRPC, and
+the `kubectl mayastor` plugin — and is driven asynchronously by a dedicated reconciler and a
+persisted drain state machine.
+
+users interact with it via `kubectl mayastor drain pool <id>` (and `PUT /pools/{id}/drain`). The
```
Abhinandan-Purkait:
```suggestion
Users interact with it via `kubectl mayastor drain pool <id>` (and `PUT /pools/{id}/drain`). The
```

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,625 @@
...
+
+- Drain a pool while it stays **online** by default, with **no data unavailability** for the volumes whose
+  replicas are migrated.
+- Support evacuation of every volume scenario: single-replica published, multi-replica, unpublished,
+  and degraded volumes.
+- Provide an explicit, user-chosen **snapshot policy** for snapshots that remain on a drained
+  pool.
+- Provide **abort** support to cancel a queued or in-progress drain.
+- Provide **progress visibility** for a drain in flight.
+- Provide a **dry-run** pre-flight analysis that reports what a drain would do without mutating state.
```
Abhinandan-Purkait, line 78: Maybe while documenting this, add a NOTE saying the dry run only computes the current state and doesn't account for changes between the dry run output and actual drain execution, just to set clear expectations.

Abhinandan-Purkait (approved):

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,701 @@
...
+
+pub struct TempReplicaConfig {
+    /// Extra replicas added on top of the user-defined count (1 for drain).
+    pub replica_count: i8,
+    pub requester: TempReplicaRequester,
+}
+
+pub enum TempReplicaRequester {
+    PoolDrain = 1,
+}
```
tiagolobocastro: How about something like this?

```suggestion
pub enum TempReplicaConfig {
    PoolDrain(DrainingReplica)
}

pub struct DrainingReplica {
    replica: ReplicaId,
    pool: PoolId,
    node: NodeId,
}
```

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,701 @@
...
+blocks it. Concretely, the drain's `self_cordon` blocks new `replicas`, `snapshots`, and `restores`
+(nothing new should land on a pool being emptied) but deliberately leaves `import = false`: a
+draining pool must still be importable so it can be brought back online after a node/engine restart
+and continue serving I/O while it drains. The user cordon retains full say over imports — if the
+user cordoned the pool with `import = true`, `effective_cordon().import` stays `true` and the pool
+remains blocked for imports, since the OR keeps whichever field is more restrictive. So the drain
+never *relaxes* a user-imposed import block; it only refrains from adding one of its own.
+
+On `VolumeSpec`, a managed `temporary_replica` marker records that the volume was over-replicated for
+a drain move. The reconciler reads `effective_replica = num_replicas (user-defined) +
```
tiagolobocastro: I think we need to clarify which component would be using the virtual replica increase? Do we want all reconcilers/components to see replica count of 3 instead of 2 for example?
Or do we only want the volume replica-count reconcilers to see it?

What about snapshots, should they be taken in all replicas?

Also, as soon as replica is out we don't want to see replicas as 3 right? So, should the `TempReplicaConfig` include the ReplicaId as well?

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,701 @@
...
+
+On `PoolState`, `DrainProgress` carries the current phase and the initial-vs-current statistics. The
+baseline (`InitialAllocationStats`) is captured once at queue time and **persisted to etcd** as a
+`PoolDrainRecord` keyed by pool id — persisting both the baseline and the phase is what lets a drain
+resume from where it left off if the core agent restarts, since the live counts alone cannot tell how
+much has already drained. Current values are read live from the data plane on each progress query. Current
+replica and snapshot count is already part of `PoolState`:
+
+```rust
+pub struct PoolState {
```
tiagolobocastro, line 789: PoolState is the data-plane state, this can't contain the drain progress, like you want the `CtrlPoolState` or the `PoolConfig`

```suggestion
pub struct CtrlPoolState {
```

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,701 @@
...
+    pub current_stats: CurrentUsageStats,
+}
+
+/// Captured once at queue time; immutable for the life of the drain.
+pub struct InitialUsageStats {
+    pub initial_replica_count: u64,
+    pub initial_snapshot_count: u64,
+    pub initial_allocation: u64,
+    pub initial_committed: u64
+}
```
tiagolobocastro: Just have a single struct, no need to keep 2?

```suggestion
pub struct UsageStats {
    pub replica_count: u64,
    pub snapshot_count: u64,
    pub allocation: u64,
    pub committed: u64
}
```

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,701 @@
...
+
+The drain is driven by a `DrainPhase` state machine, with its data split across the pool spec and the
+pool state. The drain request and user policy are recorded in a **`DrainSpec`** carried on its own new
+`PoolSpec.drain_spec` field (separate from the existing `cordon_drain` user-cordon field); the
+current `DrainPhase` and progress metrics live in a **`DrainProgress`** on the pool state
+(`PoolState.drain_progress`). The `DrainSpec` holds the self-cordon applied at admission plus a
+**`DrainArgs`** with the user policy, which holds:
+
+- `request_timestamp` — when the drain was applied; used for FIFO promotion out of the queue.
+- `allow_snapshot_loss` / `only_drain_replicas` — the user-chosen snapshot policy flags.
```
tiagolobocastro: `--drain-replicas-only` or `--skip-snapshots`

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,701 @@
...
+**Snapshot replicas.** Snapshots cannot be migrated in Phase 1, and moving a snapshotted replica
+preserves its snapshots on the pool. The user chooses a policy via mutually-exclusive flags that
+govern only the snapshots left behind:
+
+- `--accept-snapshot-loss` — move all replicas, then destroy the remaining snapshots on the pool.
+  Pool allocation reaches zero and the pool reaches `Drained`.
+- `--only-drain-replicas` — move all replicas, leaving any snapshots in place. The pool reaches
+  `PartiallyDrained` with non-zero allocation and stays online serving the snapshots; the user
+  can later create volumes from those snapshots once the pool is uncordoned. Because allocation
+  stays non-zero, a subsequent `DestroyPool` still fails (*"please drain the diskpool before
```
tiagolobocastro, line 320: At this point how to delete the snapshots?
We could allow user to re-issue drain with --accept-snapshot-loss? Then Drain would resume and delete the snapshots?

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,701 @@
...
+since it would conflate "wait then evict" with "evict immediately" on the same field — immediate
+eviction is instead a separate, explicit option (`unsafe_evict`, below). To disable forced eviction
+the user simply omits the flag.
+
+As a hard safety rule, forced eviction is **never** applied when the replica on the draining pool is
+the volume's only `Online` child, since removing it would cause data unavailability. By default
+(when `--unsafe-rebuild-otherwise-evict` is omitted) the design never degrades a volume to make progress, and a pool with
+an unplaceable replica stays `Draining` until the user either frees a destination or aborts.
+
+**Direct unsafe eviction (`unsafe_evict`, REST/gRPC only).** For power users who explicitly accept
```
tiagolobocastro: Should be on the plugin as well, but with hidden docs and not documented on the website perhaps

tiagolobocastro (commented):
What happens when a user tries to snapshot a volume containing a replica in a draining pool? Pool is cordoned so same behavior we have today but I can't remember, is the create snapshot failed?

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Yes, that is the way.

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: So, don't need count of replica, Pool drain requested will do scale up (+1) anyway?

tiagolobocastro: We could have a list of `TempReplicaConfig`, though one at a time might be a good start?

CC @Abhinandan-Purkait

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Shall we put this in 

#[derive(Serialize, Deserialize, Debug, PartialEq, Default, Clone)]
pub struct PoolState {
    /// Pool information returned by the Io-Engine.
    pub pool: transport::PoolState,
    pub ctrl_pool_state: CtrlPoolState,
}

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Do you mean what if snapshot request comes in for volume who is having TempReplicaConfig?

abhilashshetty04: I was thinking scale down should always try to Remove child from the `Draining` pool.  So we do not need ReplicaId in the TempReplicaConfig particularly.

abhilashshetty04: `volume_replica_count_reconciler` and `nexus_replica_count_reconciler` will definately need to see effective_num_replica. Im thinking health may not report Volume as degraded when we are `over-replicating`. On usual scale up we mark volume as `Degraded`. Here shall we just check if `num_replica` is Online in Volume and report `Healthy`?

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: I updated the design to ensure `TempReplicaConfig` targeted to move a specific replica from drain pool to help serialize multiple replica drains from same volume. So a concurrent drain of a pools housing a child-sibling will be deferred until `TempReplicaConfig` is cleared post Full rebuild of spare. 

We will have `need_spare: bool` which will act as over-replication specifier and will decide effective_replica count. We do not evict the child if it elapses timeout while its waiting on siblings drain completion. Now we record the time when TempReplicaConfig was applied to drain a particular replica and that time is used to check elapsed Duration. So each drain replica gets fair duration to get scheduled elsewhere.
@tiagolobocastro , @Abhinandan-Purkait

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Do we support snapshot creation on nodes less then num_replicas? If not, then we should not allow. On the draining node we anyway cordon for snapshot.

tiagolobocastro: > Im thinking health may not report Volume as degraded when we are over-replicating. On usual scale up we mark volume as Degraded. Here shall we just check if num_replica is Online in Volume and report Healthy?

Not sure; when we have a nexus we usually just mirror whatever state it has; let's say in future there's another reason to mark nexus as degraded even if all replicas all online, we shouldn't override that
Should be fine to show volume as Degraded but healthy

tiagolobocastro: > Do we support snapshot creation on nodes less then num_replicas? If not, then we should not allow. On the draining node we anyway cordon for snapshot.

I'm not sure, please check on this

tiagolobocastro (approved):

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -0,0 +1,894 @@
...
+    /// Replica on the draining pool; removed once the move completes.
+    draining_replica: ReplicaId,
+    /// Pool/node being drained (where `draining_replica` lives).
+    pool: PoolId,
+    node: NodeId,
+    /// true  → safe over-replicate: create a spare, wait for it, then scale down
+    ///          `draining_replica`. Contributes +1 to `effective_replica`.
+    /// false → unsafe direct eviction: remove `draining_replica` and let the
+    ///          self-heal loop recreate it degraded. Contributes +0.
+    need_spare: bool,
```
tiagolobocastro: Why is this needed? If we don't need spare then we don't even need to set `TempReplicaConfig`

abhilashshetty04: This is to serialise the evict if we end up in a scenario where 2 replica of the same volume is in Draining Pool.

abhilashshetty04: Shall we call this something other then TempReplicaConfig?

tiagolobocastro: > This is to serialise the evict if we end up in a scenario where 2 replica of the same volume is in Draining Pool.

Then what you might want instead is a new Operation to be added to the Volume, which will serialize via the update interface

Also this should not go in VolumeSpec directly, but the metadata fields and to consider if we want these fields to be persisted to etcd or part of runtime only.

> Shall we call this something other then TempReplicaConfig?

One easy one, drop the e: `TmpReplicaConfig`

@Abhinandan-Purkait @niladrih  are suggestions?

Or maybe a more generic `ReplicationConfig` ?

niladrih (approved):

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Its part of `VolumeMetadataRuntime` field now. persisted in etcd via `PoolDrainRecord`. updates of this data will be done through `ResourceMove` trait.

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Snapshot is not created if Volume is in `Degraded` state or any one of the replica pool is cordoned for snapshot. I feel we can just keep the existing behavior. As soon as any replica hosting Pool is in `Queued` state its self-cordon it for snapshot so that means we block snapshot request until replica gets moved to non `Draining` pool.

abhilashshetty04:
@mergify queue

PR #4253, opened 2026-06-24
chore: add sync workflow for dependabot config

pchandra19:


.github/sync.yml
```diff
@@ -0,0 +1,15 @@
...
+  repos:
+    - openebs/mayastor
+    - openebs/mayastor-control-plane
+    - openebs/mayastor-extensions
+    - openebs/mayastor-dependencies
+    - openebs/spdk-rs
+    - openebs/lvm-localpv
+    - openebs/zfs-localpv
+    - openebs/linux-utils
+    - openebs/dynamic-localpv-provisioner
```
tiagolobocastro: These are go rather than rust, so we need to change the file to include go

.github/workflows/sync-dependabot.yml
```diff
@@ -0,0 +1,24 @@
+name: Sync Dependabot Configuration
+
+on:
+  workflow_dispatch:
+
+  push:
+    branches:
+      - main
```
tiagolobocastro:
```suggestion
      - develop
```

tiagolobocastro (approved):

.github/sync.yml
pchandra19: Yes, handled this scenario using templating.

.github/workflows/sync-dependabot.yml
pchandra19: Removed trigger on push.

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

PR #4255, opened 2026-06-26
docs: document recursive submodule setup for local development

immanuwell:
## Description
Document recursive submodule setup for local Rust dev in `README.md` and `CONTRIBUTING.md`.

## Motivation and Context
Fresh clones cant run `cargo test -p openebs-upgrade --lib` out of the box. It fails before compile because `mayastor/constants/Cargo.toml` is missing. Kinda rough for first time contributors.

This points folks to `git clone --recurse-submodules`, `git submodule update --init --recursive`, and the existing helper script.

## Testing
Repro on a fresh clone:
`git clone https://github.com/openebs/openebs.git`
`cd openebs`
`cargo test -p openebs-upgrade --lib`

It fails with:
`failed to read .../mayastor/constants/Cargo.toml`

After the documented fix:
`git submodule update --init --recursive`
`cargo test -p openebs-upgrade --lib`

That one passes and runs the `openebs-upgrade` unit test, yep.

## Backward Compatibility
Docs only. No runtime change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,26 @@ These features make OpenEBS a robust and flexible solution for managing persiste
 - [Release Process](./RELEASE.md)
 - [Roadmap Tracker](https://github.com/orgs/openebs/projects/78)
 
+### Local Development
+
+The Rust workspace depends on the `mayastor` submodule and its nested submodules. If you want to build or test the workspace locally, clone the repository with submodules:
+
+```bash
+git clone --recurse-submodules https://github.com/openebs/openebs.git
+```
+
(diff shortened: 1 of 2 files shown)


README.md
tiagolobocastro, line 1: I think we should not add this to README, instead we should point to a docs/ similar what we have on other repos, ie:
https://github.com/openebs/mayastor/blob/develop/doc/contributor.md
https://github.com/openebs/zfs-localpv/blob/develop/docs/developer-setup.md

PR #4274, opened 2026-07-28
docs(5198): disk health (SMART) support in io-engine

susobhandey:
Add the OpenEBS Enhancement Proposal for reporting disk health from
    the io-engine data plane.

    io-engine currently exposes no health information for the devices
    backing a Mayastor pool, so a degrading disk is only noticed once it
    starts failing I/O. This proposal adds a single device_health() method
    on the BlockDevice trait, backed by two readers chosen by attachment
    type: smartctl for kernel-attached disks (aio/uring, covering SATA,
    SAS and non-VFIO NVMe), and an NVMe GET LOG PAGE admin command for the
    SMART log page (0x02) on VFIO-attached NVMe, which has no /dev node.
    Both normalise into one DeviceHealth struct, surfaced through a new
    GetPoolHealth RPC on the pool service. Devices that cannot report
    health return supported = false rather than failing the call.

    Requires no SPDK source changes; only enables an opcode constant
    already present in the spdk-rs bindings.

    Signed-off-by: Susobhan Dey <susobhan.dey@gmail.com>

### DIFF
--- a/designs/replicated-pv/mayastor/smart-health-dataplane.md
+++ b/designs/replicated-pv/mayastor/smart-health-dataplane.md
@@ -1,0 +1,620 @@
+---
+oep-number: OEP 5198
+title: Disk Health (SMART) Support in io-engine
+authors:
+  - "@susobhandey"
+owners:
+  - "@susobhandey"
+editor: TBD
+creation-date: 2026-07-02
+last-updated: 2026-08-17
+status: implementable
+---
+
+# Disk Health (SMART) Support in io-engine
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Implementation Details](#implementation-details)
+    - [Data Model](#data-model)
+    - [Device Information](#device-information)
+    - [SMART Attributes and Thresholds](#smart-attributes-and-thresholds)
+    - [The Common Method](#the-common-method)
+    - [Branch Selection](#branch-selection)
+    - [Common Start and End](#common-start-and-end)
+    - [Path A — smartctl (kernel-attached disks)](#path-a--smartctl-kernel-attached-disks)
+    - [Path B — NVMe SMART log page (VFIO-attached NVMe)](#path-b--nvme-smart-log-page-vfio-attached-nvme)
+    - [Combined Flow](#combined-flow)
+    - [gRPC / Protobuf Interface](#grpc--protobuf-interface)
+    - [Error Handling](#error-handling)
+    - [Concurrency and Locking](#concurrency-and-locking)
+    - [File-Level Changes](#file-level-changes)
+    - [Runtime Requirements](#runtime-requirements)
+    - [Assumptions and Dependencies](#assumptions-and-dependencies)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed](#infrastructure-needed)
+- [Testing](#testing)
+  - [Unit tests](#unit-tests)
+  - [Hardware and integration tests](#hardware-and-integration-tests)
+  - [BDD tests](#bdd-tests)
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


designs/replicated-pv/mayastor/smart-health-dataplane.md
```diff
@@ -0,0 +1,495 @@
+| oep-number    | OEP 5198                                            |
```
tiagolobocastro: hmm the format is not like this, please check existing OEP, like [this](https://github.com/openebs/openebs/blob/develop/designs/replicated-pv/mayastor/offline-volume-rebuild.md) or check the template [here](https://github.com/openebs/openebs/blob/develop/designs/oep-template.md).

designs/replicated-pv/mayastor/smart-health-dataplane.md
```diff
@@ -0,0 +1,495 @@
...
+| `temperature_celsius` | Composite temperature in Celsius |
+| `available_spare_percent` | Remaining spare capacity |
+| `available_spare_threshold_percent` | Spare capacity warning level |
+| `percentage_used` | Estimated life consumed |
+| `data_units_read` / `data_units_written` | Volume read / written |
+| `power_cycles` | Number of power cycles |
+| `power_on_hours` | Hours powered on |
+| `unsafe_shutdowns` | Number of unsafe shutdowns |
+| `media_errors` | Media / data integrity errors |
+| `num_error_log_entries` | Number of error log entries |
```
tiagolobocastro, line 168: missing the `SMART Attributes with Thresholds` ?

tiagolobocastro (commented):
btw would we want to show the device information as well? Like this:

```
=== START OF INFORMATION SECTION ===
Model Family:     Crucial/Micron Client SSDs
Device Model:     CT250MX500SSD1
Serial Number:    1809E1319870
LU WWN Device Id: 5 00a075 1e1319870
Firmware Version: M3CR023
User Capacity:    250,059,350,016 bytes [250 GB]
Sector Sizes:     512 bytes logical, 4096 bytes physical
Rotation Rate:    Solid State Device
Form Factor:      2.5 inches
TRIM Command:     Available
Device is:        In smartctl database 7.5/6061
ATA Version is:   ACS-3 T13/2161-D revision 5
SATA Version is:  SATA 3.3, 6.0 Gb/s (current: 6.0 Gb/s)
Local Time is:    Tue Jul 28 18:03:49 2026 BST
SMART support is: Available - device has SMART capability.
SMART support is: Enabled
AAM feature is:   Unavailable
APM level is:     254 (maximum performance)
Rd look-ahead is: Enabled
Write cache is:   Enabled
DSN feature is:   Unavailable
ATA Security is:  Disabled, frozen [SEC2]
```

Would things like the model,firmware and link speed be useful?

susobhandey:
@tiagolobocastro However I think all the controller information should be prefetched not during the runtime. I need to change that

tiagolobocastro:
@susobhandey another thing worth noting is adding the smart data to the support bundle generated by the plugin
Though that should come in scope only when we have the control-plane api to fetch from the data plane

designs/replicated-pv/mayastor/smart-health-dataplane.md
```diff
@@ -0,0 +1,620 @@
...
+oep-number: OEP 5198
+title: Disk Health (SMART) Support in io-engine
+authors:
+  - "@susobhandey"
+owners:
+  - "@susobhandey"
+editor: TBD
+creation-date: 2026-07-02
+last-updated: 2026-07-29
+status: provisional
```
tiagolobocastro: Let's mark this down since you already have the code in PR
@openebs/org-maintainers pls review

```suggestion
status: implementable
```

tiagolobocastro (approved):
You have a typo making tests fail, pls fix:
> [2](https://github.com/openebs/openebs/actions/runs/31642852742/job/94456016944?pr=4274#step:5:133)
designs/replicated-pv/mayastor/smart-health-dataplane.md:319: unparseable ==> unparsable

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

abhilashshetty04:
spellcheck:
designs/replicated-pv/mayastor/smart-health-dataplane.md:319: unparseable ==> unparsable

Also DCO has warning
@susobhandey

niladrih (approved):

designs/replicated-pv/mayastor/smart-health-dataplane.md
susobhan87: done

tiagolobocastro:
@mergify queue

PR #4275, opened 2026-07-28
make pool drain oep as implementable

abhilashshetty04:


designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -139,14 +99,12 @@ drain has progressed. The `DrainPhase` state machine has the states:
...
-  cordon in `cordon_drain` is left untouched. Unlike `Drained`, this is not a retained terminal
-  state — once cleanup completes the drain record is gone.
-
-**Pre-requisites:** the pool must be `Online` and must not already be in a `Queued`, `Draining`, or
+  (with `phase = Aborted`) throughout cleanup — not cleared up front — so the pool stays discoverable
+  and self-cordoned while the drain's changes are unwound; only once cleanup completes are
+  `drain_spec` and the `PoolDrainRecord` cleared. Unlike `Drained` this is not a retained terminal
+  state (see *Abort support*).
+
+**Pre-requisites:** the pool must not already be in a `Queued`, `Draining`, or
```
tiagolobocastro, line 112: I thought we were going to take drain command again for accepting the snapshot loss if the user had not first done so?

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -257,178 +243,81 @@ flowchart TD
...
+- **Never published:** there is no data to rebuild, so — the only case that differs — remove the
+  draining-pool child and add a replacement on an eligible pool.
+
+**Snapshot replicas.** Snapshots cannot be migrated in Phase 1 and a moved replica leaves its snapshots
+behind on the pool, so the user must choose a policy — a drain of a pool holding snapshots with
+**neither** flag set is **rejected**:
+
+- `--accept-snapshot-loss` — move all replicas, then destroy the snapshots left on the pool; allocation
+  reaches zero and the pool reaches `Drained`.
+- `--drain-replicas-only` — move all replicas and leave the snapshots; the pool reaches
```
tiagolobocastro: Something like this seems clearer to me but I'll let others weigh in

```suggestion
- `--ignore-snapshots` — move all replicas and leave the snapshots; the pool reaches
```

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -257,178 +243,81 @@ flowchart TD
...
+  `PartiallyDrained`, stays online serving them, and can be uncordoned and restored from later. Because
+  allocation stays non-zero, `DestroyPool` still fails (*"please drain the diskpool before attempting
+  destroy"*) until the user re-drains with `--accept-snapshot-loss`.
+
+(A volume with multiple replicas keeps its snapshot on the other replicas. The platform does not
+currently support having fewer snapshots than the replica count, so true snapshot evacuation waits for
```
tiagolobocastro, line 299: > The platform does not
currently support having fewer snapshots than the replica count

Is this really true? After taking 1 snapshot can't we then scale up the volume?

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -257,178 +243,81 @@ flowchart TD
...
+`Some(SpareReplica(None))`, the replica is treated as unplaceable and is force-evicted.
 
 Because `applied_at` is stamped when the move *begins* (i.e. when this replica becomes the volume's
 active move, not while it sits deferred behind a sibling), each replica gets its **own full grace
-window** regardless of how long it waited its turn. The grace period must be **strictly positive**: a
-zero duration is **rejected at admission with an error**, since it would conflate "wait then evict"
+window** regardless of how long it waited its turn. It is re-stamped in exactly one other place — a
+`Respare` unwind, where a second drain displaces this move's healthy spare (see *The spare's own pool
+enters a drain*).
+A zero duration is **rejected at admission with an error**, since it would conflate "wait then evict"
```
tiagolobocastro: > A zero duration is **rejected at admission with an error**, since it would conflate "wait then evict"

a zero duration here could mean we check if we can over-replicate, and if we can we do it, otherwise we evict

But `unsafe-evict` never over-replicates, always evicts?

tiagolobocastro (commented):
Half-way through only

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: They can do that after the current request completes with `PartiallyDrained`? This stops them from modifying ongoing drain request. We can allow also, snapshot cleanup is triggered after we move replicas.

We should block when its in `Aborted` (transient state, cleared when we undo drain changes).

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: unsafe-evict is always evict without trying. Shall we support zero duration (trying once)?

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Does work if its BestEffort policy.

designs/replicated-pv/mayastor/pool-drain.md
tiagolobocastro: > They can do that after the current request completes with PartiallyDrained?

Can't we just accept before even getting there? Make it more declarative

> We should block when its in Aborted (transient state, cleared when we undo drain changes).

Yeah this makes sense, if aborted then return appropriate error

designs/replicated-pv/mayastor/pool-drain.md
tiagolobocastro: And if we had scaled up right?

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -238,261 +223,150 @@ flowchart TD
...
-"not placed yet."
-
-Volume health is unaffected by this recovery: it simply follows the nexus status (see *Volume health
-during a drain move*), reporting `Degraded` while any spare — the dropped one or its fresh replacement
-— is rebuilding, and returning to `Online` once the move completes.
+detects `spare_replica: Some(SpareReplica { replica_id: Some(id) })` while that child is no longer present/`Online`
+in the nexus, and resets it — along with `placement_started_at`, since placement had succeeded here and
+the fresh round is entitled to its own single attempt (see *Unplaceable replicas and forced eviction*) —
+so the volume reconciler loop re-enters the placement branch and schedules
+a **fresh spare — possibly on a different eligible pool** — starting a new full rebuild. No new phase is needed;
```
tiagolobocastro, line 364: What if the rebuild keeps failing for some reason?
I think this starts to get into the realm of the "ha scheduling improvements", though I wonder if for drain we should add some sort of backoff?

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -238,261 +223,150 @@ flowchart TD
...
+#### The spare's own pool enters a drain
+
+A spare is placed on an eligible, uncordoned pool — but nothing stops *that* pool from being drained
+afterwards. (The reverse ordering cannot happen: the self-cordon is applied at admission, i.e. already
+in `Queued`, so placement never picks a pool whose drain has been requested.) Pool B's drain reconciler
+then enumerates its replicas and finds one that is not a plain replica at all but pool A's in-flight
+spare — recognised by the owning volume's marker carrying `spare_replica == that id`. We will attempt
+to remove the `spare_replica` here and set `spare_replica` to None so that we attempt to schedule it again.
+If we wait on it, it will get evicted once it becomes online from the Pool B's drain anyway and its a wasted
+Full rebuild operation.
```
tiagolobocastro: If we first cordon the pool (either by user or by the drain command itself), then can't we make the drain be in queued phase until there's no spare-replica set to that pool?

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -822,174 +624,267 @@ pub struct PoolUsageStats {
...
+        - selfCordon
+        - policy
+
+    PoolDrainPolicy:
+      description: The user's drain policy - the knobs chosen at request time
+      type: object
+      properties:
+        acceptSnapshotLoss:
+          type: boolean
+        drainReplicasOnly:
```
tiagolobocastro:
```suggestion
        ignoreSnapshots:
```

designs/replicated-pv/mayastor/pool-drain.md
```diff
@@ -737,83 +537,85 @@ pub enum ReplicaMoveRequester {
...
-    /// `need_spare = true`. `None` until the spare has been scheduled/created.
-    pub spare_replica: Option<ReplicaId>,
+    /// `None` if the user explicitly chose the unsafe_evict option — no spare is ever
+    /// wanted for this move. `Some` for the safe over-replicate flow, whose inner
+    /// `replica_id` gets populated once the spare replica is created by the scheduler.
+    pub spare_replica: Option<SpareReplica>,
+    /// Set while this move's `spare_replica` is being torn down, naming *why* so the
+    /// observer knows what to do once it is gone. `None` in steady state. Both variants
+    /// drive the same removal in the count reconciler.
+    pub unwind: Option<UnwindSpare>,
```
tiagolobocastro, line 570: IIRC we discussed some of these updates may not always be persisted to etcd, I think we should highlight this

tiagolobocastro (approved):
There's a lot of impl details here which hasn't been exactly the practice in the OEP
We may deviate somewhat during actual impl, that's fine, but major changes should then require update of the OEP to make sure it's accurate at least for the initial

niladrih (approved):

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Here i meant Restoring a snapshot into a new volume normally requires as many surviving replica-snapshots as the new volume's `num_replicas`; a restore from fewer succeeds only when the volume's snapshot restore policy is `besteffort`. We scale up once restored volume is created with fewer replicas.

User will need this if they choose --allow-snapshot-loss.

abhilashshetty04: >> And if we had scaled up right?

Applies here as well. User will have to use BestEffort

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Yes, we could have a separate rebuild limit for drain created rebuilds. keeping it simple for the Phase1. I will add this in future enhancement.

We could also record number of rebuild failure while evacuating a draining_replica in ReplicaMove and do something from there perhaps.

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: Still its a wasted full rebuild operation right if we are letting over-replication rebuild complete. We still don't touch the replica if its Degraded and rebuilding due to usual Degradation(not a spare_replica).

designs/replicated-pv/mayastor/pool-drain.md
abhilashshetty04: For unwind we wanted to test scale down on Degraded volume. We do remove Degraded child on scale down. That changes design on spare unwind a little. I will check if we can make UnwindSpare in memory only.

We had decided on not writing draining_replica being set None to etcd when drain completes. We will write to in memory only. I will mention about it.

Issue #4279, opened 2026-07-30
[OEP 4279]: complete the mayastor lvm pool backend


patrickdk77:
### Enhancement Description

- Clear and concise enhancement description (one-liner): Bring the Mayastor LVM pool backend to functional parity via dm-thin (thin provisioning, snapshots/clones), plus stats, grow, properties, and encryption
- OpenEBS Enhancement Proposal: designs/replicated-pv/mayastor/lvm-backend-completion.md (OEP PR to follow; will be linked here)
- Discussion Link: raised alongside the ZFS pool backend OEP (tracking issue to be cross-linked once filed)
- Primary contact (assignee): @patrickdk77
- Responsible OWNERs:
  - @patrickdk77
- Targeted milestone:
  - OpenEBS vx.y
- [ ] Targets
  - [ ] OEP update PR(s):
  - [ ] Code update PR(s):
    - [ ] mayastor: io-engine lvm backend phase 1 (dm-thin: thin pool in the VG, thin replicas, snapshots/clones, honest capability reporting)
    - [ ] mayastor: io-engine lvm backend phase 2 (stats, grow, properties allowlist)
    - [ ] mayastor: io-engine lvm backend phase 3 (dm-crypt encryption, pool- and volume-scoped)
  - [ ] E2E Test PR(s):
    - [ ] BDD: thin pool/replica/snapshot/clone flows on loop-device VGs; fstrim reclaims thin-pool space; capability matrix (thick-on-discardless-PV advertises no unmap, thin advertises chunk-size granularity)
  - [ ] Docs update PR(s):

Notes:
- Shares a common plumbing layer with the ZFS pool backend OEP (per-replica `properties` map; control-plane `PoolBackend` first-class end to end including CRD `poolType`, scheduler filter, CSI params). Either OEP can carry that layer; which one does is a review-time decision (carrying it here keeps the ZFS merge compact).
- Related: OEP 4074 (trim/unmap; dm-thin is what turns LVM trim passthrough into real space reclamation), OEP 3843 (at-rest encryption), the volume encryption OEP (phase 3 implements to its standard).

_Please keep this description up to date. This will help track the evolution of the enhancement efficiently._

Issue #4278, opened 2026-07-30
[OEP 4278]: mayastor zfs pool backend


patrickdk77:
### Enhancement Description

- Clear and concise enhancement description (one-liner): Add a ZFS pool backend to Mayastor (replicas as zvols, native snapshots/clones, thin provisioning, ZFS data-integrity guarantees)
- OpenEBS Enhancement Proposal: designs/replicated-pv/mayastor/zfs-pool-backend.md (OEP PR to follow; will be linked here)
- Discussion Link: https://github.com/openebs/openebs/issues/4010
- Primary contact (assignee): @patrickdk77
- Responsible OWNERs:
  - @patrickdk77
- Targeted milestone:
  - OpenEBS vx.y
- [ ] Targets
  - [ ] OEP update PR(s):
  - [ ] Code update PR(s):
    - [ ] mayastor-dependencies: `PoolType::Zfs` + `CreateReplicaRequest.properties` (io-engine API protos)
    - [ ] mayastor: io-engine `zfs` backend module (pool/replica/snapshot/clone lifecycle, options, ENABLE_ZFS gate)
    - [ ] mayastor-control-plane: pool backend type first-class end to end (transport/store, gRPC, REST/OpenAPI, translation, DiskPool CRD `poolType`, scheduler filter, CSI params)
  - [ ] E2E Test PR(s):
    - [ ] Python BDD suites: pool/replica/snapshot (gated on zfs kmod availability on runners)
  - [ ] Docs update PR(s):
    - [ ] Deployment prerequisites (zfs userland in io-engine image, /dev/zvol mount, ENABLE_ZFS)

Related: OEP 4074 (trim/unmap; zvol reclaim rides its backend-general passthrough), OEP 3843 (at-rest encryption; ZFS-native encryption satisfies the same `PoolArgs.enc_key` surface via the container dataset, deferred to the volume encryption OEP standard).

_Please keep this description up to date. This will help track the evolution of the enhancement efficiently._

Issue #4074, opened 2025-10-06
[OEP 4074]: Trim/Unmap should free up diskpool storage


tiagolobocastro:
### Enhancement Description

- Clear and concise enhancement description (one-liner): Trim/Unmap should free up diskpool storage 
- OpenEBS Enhancement Proposal: 
- Discussion Link: 
- Primary contact (assignee):
- Responsible OWNERs:
  - @tiagolobocastro 
- Targeted milestone:
  - OpenEBS vx.y
- [ ] Targets
  - [ ] OEP update PR(s):
  - [ ] Code update PR(s):
  - [ ] E2E Test PR(s):
  - [ ] Docs update PR(s):

_Please keep this description up to date. This will help track the evolution of the enhancement efficiently._

tiagolobocastro:
This was on the old "roadmap" but not transferred as a github issue.
As the next step, I'll go ahead a create the initial proposal

biclighter81:
We are currently running OpenEBS Mayastor with thin-provisioned volumes and replication factor 2 in production, and the lack of diskpool reclaim after TRIM/UNMAP is a significant operational problem for us.

In our cluster we cleaned up a large amount of data inside several PVCs (~50 GB). The files were deleted at the filesystem level, and we verified that discard is working correctly:

* `lsblk -D` shows discard support on the NVMe devices exposed by Mayastor.
* Running `fstrim` on the host mounts under `/var/lib/kubelet/plugins/kubernetes.io/csi/io.openebs.csi-mayastor/.../globalmount` successfully reports multiple gigabytes trimmed per volume.

Example:

```
fstrim -av
.../globalmount: 15.4 GiB trimmed on /dev/nvme15n1
.../globalmount: 3.6 GiB trimmed on /dev/nvme1n1
.../globalmount: 2.9 GiB trimmed on /dev/nvme0n1
```

However, the diskpool `USED` capacity does not decrease, and the `ALLOCATED` values of the volume replicas remain unchanged.

This means that even though the filesystem and the block device support discard, the freed space is not returned to the diskpool allocator. The only way for us to reclaim space currently is to delete and recreate PVCs and migrate data manually.

In our case this affects workloads like ClickHouse and other stateful services where data churn is normal.

Proper reclaim of trimmed blocks back to the diskpool would make thin provisioning behave as expected and significantly reduce operational overhead for clusters using Mayastor.

[...]

PR #4281, opened 2026-07-30
oep(4279): add oep for completing the mayastor lvm pool backend

patrickdk77:
Adds the OEP for bringing the Lvm pool backend to functional parity, phased on
dm-thin: thin provisioning and snapshots/clones (phase 1), pool/replica stats,
grow, and replica properties (phase 2), and dm-crypt encryption to the volume
encryption OEP standard (phase 3). Includes the layering comparison with the
blobstore path and the relationship to the ZFS backend OEP (either OEP can
carry the shared control-plane plumbing; review-time decision).

Doc-only PR per the OEP process; no other changes.

Part of #4279
Related: #4074, #4278

### DIFF
--- a/designs/replicated-pv/mayastor/lvm-backend-completion.md
+++ b/designs/replicated-pv/mayastor/lvm-backend-completion.md
@@ -1,0 +1,227 @@
+---
+oep-number: OEP 4279
+title: Completing the Mayastor LVM pool backend
+authors:
+  - "@patrickdk77"
+owners:
+  - "@patrickdk77"
+editor: TBD
+creation-date: 30/07/2026
+last-updated: 30/07/2026
+status: provisional
+see-also:
+  - OEP 4074
+  - OEP 3843
+  - OEP 4278
+---
+
+# OEP: Completing the Mayastor LVM pool backend
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Relationship to the ZFS backend OEP](#relationship-to-the-zfs-backend-oep)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Layering, compared with the blobstore path](#layering-compared-with-the-blobstore-path)
+  - [Phase 1: dm-thin (thin provisioning + snapshots together)](#phase-1-dm-thin-thin-provisioning--snapshots-together)
+  - [Phase 2: stats, grow, properties](#phase-2-stats-grow-properties)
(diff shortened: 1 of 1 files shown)


designs/replicated-pv/mayastor/lvm-backend-completion.md
```diff
@@ -0,0 +1,227 @@
...
+### Phase 3: encryption
+
+We defer to the volume encryption OEP as the standard to follow for the
+implementation (key sourcing, key lifecycle, and the user-facing surface), with
+dm-crypt as the mechanism satisfying it for LVM:
+
+- Pool-scoped (parity with OEP 3843): dm-crypt under the PV (or LUKS on the PV
+  device), managed at pool create/import; the SPDK crypto-vbdev path used for LVS
+  does not apply because the pool is not an SPDK bdev.
+- Volume-scoped: dm-crypt per LV. LVM has no native per-LV encryption (unlike ZFS
```
tiagolobocastro, line 181: Potentially we could use the `crypto-vbdev` here, depending on how we end up implementing it for the blobstore;
ie if we place it on top of the nexus, then we could leverage the same here

designs/replicated-pv/mayastor/lvm-backend-completion.md
```diff
@@ -0,0 +1,227 @@
...
+- BDD suites mirroring the existing LVM pool/replica suites, extended with thin
+  pool/replica/snapshot/clone flows on loop-device VGs.
+- Trim: fstrim on a thin LVM replica reclaims thin-pool space (extends the OEP 4074
+  e2e test to LVM thin).
+- Capability matrix tests: thick-on-discardless-PV must not advertise unmap; thin
+  must advertise at chunk-size granularity.
+
+## Open items to settle during review
+
+1. Thin-pool sizing/growth policy (fixed at create vs auto-extend via
```
tiagolobocastro, line 218: We could have this as a user option?
Though it has to also be enabled on `/etc/lvm/lvm.conf` IIRC

tiagolobocastro (approved):

designs/replicated-pv/mayastor/lvm-backend-completion.md
patrickdk77: Yes, just thoughts, without actual design, this is just how it might work. As the diagram shows, nothing at or above nexus is touched, so if encryption is handled at or above nexus, nothing for lvm/zfs to worry about.

niladrih (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
@mergify queue

PR #4282, opened 2026-07-30
oep(4278): add oep for mayastor zfs pool backend

patrickdk77:
Adds the OEP for a third Mayastor pool backend, Zfs: replicas as zvols with
native snapshots/clones, thin provisioning, per-pool/per-replica options, and
ZFS data-integrity guarantees. Covers the object model, ownership via
io.mayastor: user properties, snapshot/clone/destroy semantics (including the
HasLiveSnapshots divergence needing maintainer sign-off), encryption mapping
(deferred, see-also OEP 3843), layering versus the blobstore path, and
implementation phases.

Doc-only PR per the OEP process; no other changes.

Part of #4278
Related: #4074, #4279

### DIFF
--- a/designs/replicated-pv/mayastor/zfs-pool-backend.md
+++ b/designs/replicated-pv/mayastor/zfs-pool-backend.md
@@ -1,0 +1,284 @@
+---
+oep-number: OEP 4278
+title: Mayastor ZFS pool backend
+authors:
+  - "@patrickdk77"
+owners:
+  - "@patrickdk77"
+editor: TBD
+creation-date: 30/07/2026
+last-updated: 30/07/2026
+status: provisional
+see-also:
+  - OEP 4074
+  - OEP 3843
+  - OEP 4279
+---
+
+# OEP: Mayastor ZFS pool backend
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Layering, compared with the blobstore path](#layering-compared-with-the-blobstore-path)
+  - [Object model](#object-model)
+  - [Ownership via ZFS user properties](#ownership-via-zfs-user-properties)
+  - [Snapshot / clone / destroy semantics](#snapshot--clone--destroy-semantics)
+  - [Options](#options)
+  - [Encryption](#encryption)
+- [User-facing changes](#user-facing-changes)
+- [Implementation surface (3 repos, coordinated)](#implementation-surface-3-repos-coordinated)
+- [Graduation Criteria](#graduation-criteria)
+- [Testing](#testing)
+- [Alternatives](#alternatives)
+- [Decided positions](#decided-positions)
+- [Implementation phases](#implementation-phases)
+- [Implementation History](#implementation-history)
+
+## Summary
+
+Add a third Mayastor pool backend, `Zfs`, alongside `Lvs` (SPDK blobstore) and `Lvm`
+(volume group). A ZFS pool's `disks` entry is a ZFS dataset path; replicas are zvols
+exposed to SPDK through an aio bdev; snapshots and clones map to native `zfs snapshot` /
+`zfs clone`. This gives Mayastor native, first-class snapshots/clones on a widely deployed
+filesystem, modeled on how openebs/zfs-localpv drives ZFS.
+
+## Motivation
+
+- Data integrity guarantees against corruption: ZFS checksums every block end to end,
+  self-heals from redundant vdevs on checksum mismatch, and scrubs proactively.
+  Neither the blobstore nor LVM detects silent corruption (bit rot, misdirected
+  writes, phantom writes) at all; with ZFS backing, each replica independently
+  guarantees the integrity of the data it returns, which composes with (rather than
+  relies on) Mayastor's replica redundancy.
+- The `Lvm` backend cannot do snapshots or clones at all; `Lvs` snapshots are blobstore
+  internal. ZFS provides mature, native snapshot/clone/send primitives.
+- Operators already running ZFS (via zfs-localpv or otherwise) can use the same storage
+  model under Mayastor's replication/nexus layer (prior demand: openebs/openebs#4010,
+  "Unable to use zvol in a disk pool").
+- A shared ZFS pool can serve more than Mayastor volumes: the same zpool (and the
+  same operational model) can host zfs-localpv volumes and filesystem datasets
+  backing an S3 object store alongside Mayastor's container dataset. Mayastor
+  claiming only its own container dataset (ownership via local `io.mayastor:`
+  properties, capacity bounded by `?quota=`) is what makes that coexistence safe.
+- The backend is CLI-driven (`zfs(8)` arg-vector exec, no libzfs), so it carries no new
+  linked dependency, exactly like zfs-localpv.
+
+### Goals
+
+- A `Zfs` pool backend selectable per DiskPool, gated by `ENABLE_ZFS`.
+- Replicas as zvols; pool/replica options (volblocksize, compression, dedup, logbias,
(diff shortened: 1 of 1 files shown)


designs/replicated-pv/mayastor/zfs-pool-backend.md
```diff
@@ -0,0 +1,284 @@
...
+  relies on) Mayastor's replica redundancy.
+- The `Lvm` backend cannot do snapshots or clones at all; `Lvs` snapshots are blobstore
+  internal. ZFS provides mature, native snapshot/clone/send primitives.
+- Operators already running ZFS (via zfs-localpv or otherwise) can use the same storage
+  model under Mayastor's replication/nexus layer (prior demand: openebs/openebs#4010,
+  "Unable to use zvol in a disk pool").
+- A shared ZFS pool can serve more than Mayastor volumes: the same zpool (and the
+  same operational model) can host zfs-localpv volumes and filesystem datasets
+  backing an S3 object store alongside Mayastor's container dataset. Mayastor
+  claiming only its own container dataset (ownership via local `io.mayastor:`
```
tiagolobocastro, line 66: This is good, and we document that properties that start with this prefix must not be edited directly via zfs

designs/replicated-pv/mayastor/zfs-pool-backend.md
```diff
@@ -0,0 +1,284 @@
...
+  properties, capacity bounded by `?quota=`) is what makes that coexistence safe.
+- The backend is CLI-driven (`zfs(8)` arg-vector exec, no libzfs), so it carries no new
```
tiagolobocastro, line 68: I'm not even sure if we could use libzfs as CNCF restricts the licenses we may depend on

designs/replicated-pv/mayastor/zfs-pool-backend.md
```diff
@@ -0,0 +1,284 @@
...
+### Snapshot / clone / destroy semantics
+
+- Deferred destroy: destroying a snapshot that still has clones sets
+  `io.mayastor:discarded=true` and `zfs destroy -d`; it lingers until the last clone goes.
+- Replica destroy with snapshots: no snapshots -> destroy; only discarded clone-less
+  snapshots -> `destroy -r`; snapshot with clones -> `zfs promote` then re-evaluate; a
+  remaining **live** clone-less snapshot -> refuse with `HasLiveSnapshots`. This diverges
+  from `Lvs` (where snapshots outlive the replica) because a ZFS snapshot is physically
+  bound to its dataset; the control-plane delete-but-keep-snapshots flow must delete
+  snapshots first. **This divergence is the main item needing maintainer sign-off.**
```
tiagolobocastro, line 156: so there's 2 options:
1. we can delete the snapshot control-plane object, keeping the data-plane snapshot alive
2. mark control-plane object as non-needed, or something like that, to be deleted once/if the clone is deleted

I think I prefer 1, as it may not make sense to keep snapshot object in control-plane, since the clone may continue living "forever"

designs/replicated-pv/mayastor/zfs-pool-backend.md
```diff
@@ -0,0 +1,284 @@
...
+
+### Options
+
+- Pool defaults: query params on the disks string
+  (`tank/data?compression=zstd&volblocksize=16k&quota=200GiB`); ZFS inheritance applies
+  them to child zvols.
+- Per-replica overrides: `map<string,string> properties` on `CreateReplicaRequest`
```
tiagolobocastro, line 163: I had a comment on the other PR, I'll ask it here as well for visibility

1. should we have keys for each pool type
2. should each option be specified in the protobuf itself rather than a generic

CC @Abhinandan-Purkait @abhilashshetty04

designs/replicated-pv/mayastor/zfs-pool-backend.md
```diff
@@ -0,0 +1,284 @@
...
+1. mayastor-dependencies: `PoolType::Zfs`, `CreateReplicaRequest.properties` (proto).
+2. mayastor / io-engine: the `zfs` backend module + backend/gRPC/env registration +
+   `bdev_as_replica` URI-based resolution (nexus snapshot path) + BDD tests.
+3. mayastor-control-plane: `PoolBackend` as a first-class concept end-to-end (transport,
+   store with serde defaults for etcd back-compat, gRPC, REST/OpenAPI, translation
+   un-hardcode, DiskPool CRD version, scheduler pool-type filter, CSI param plumbing).
+
+## Graduation Criteria
+
+- Pool/replica/snapshot/clone BDD suites green on zfs-kmod-capable runners.
```
tiagolobocastro, line 214: We may want to split the CI's into parallel workflows testing all backends
Hopefully some of the existing tests can be ran for all backend types... however some are, unfortunately, with hardcoded values for allocation etc that would probably require changes to support other backends
And ofc some tests will not happy to other backends, and vice versa

designs/replicated-pv/mayastor/zfs-pool-backend.md
```diff
@@ -0,0 +1,284 @@
...
+- fstrim on a published volume reclaims zpool space through a sparse zvol replica.
+- Pool export/import (including node reboot) preserves replicas, shares, and snapshots.
+- Control-plane places replicas only on Zfs pools when the StorageClass requests it.
+
+## Testing
+
+- Rust unit tests (no zfs binary): options/argv building, `-Hp` parse fixtures
+  (local/inherited/none sources), property round-trips, dataset-name validation.
+- Python BDD mirroring LVM (pool/replica/snapshot), gated to skip when `/sys/module/zfs`
+  is absent (needs the zfs kmod on the runner).
```
tiagolobocastro, line 225: I think we could enforce zfs to be present on CI at least, to ensure zfs are not silently skipped?

designs/replicated-pv/mayastor/zfs-pool-backend.md
```diff
@@ -0,0 +1,284 @@
...
+   rename remains a documented option if control-plane compatibility later demands
+   it, but is not part of this OEP.)
+2. Nexus-driven snapshots are part of the initial implementation, not a follow-up.
+   The nexus resolves a child replica via `IReplicaFactory::bdev_as_replica(bdev)`
+   (both the local path and the remote NVMe admin passthru path); for `Zfs` the
+   replica bdev is an aio bdev, so the implementation derives the replica identity
+   from the bdev URI (`aio:///dev/zvol/<dataset>?uuid=<uuid>`), which carries both the
+   dataset and the uuid. Subsequent operations are async and may query `zfs` as
+   needed.
+3. Deferred-destroyed snapshots vanish when their last clone goes; the control-plane
```
tiagolobocastro, line 248: As long as we can distinguish from when the zfspool is not found vs the snapshot not found all good
I think we have already this kind of logic for lvs/lvm with different error code (maybe failed precondition or something like that)

tiagolobocastro (approved):
Looks good so far, I'm away for work so haven't had much time to check things
We can also merge and iterate, with status transition to implementable as the final approval

niladrih (approved):

designs/replicated-pv/mayastor/zfs-pool-backend.md
Abhinandan-Purkait: Yep, option 1 sounds good.

designs/replicated-pv/mayastor/zfs-pool-backend.md
Abhinandan-Purkait: I also feel maybe typed and an enum might be better, I understand the intent here is to keep it absolutely generic, but the conversion on the receiver will not be trivial probably.

Abhinandan-Purkait (approved):

designs/replicated-pv/mayastor/zfs-pool-backend.md
tiagolobocastro: @patrickdk77 are you ok with this?

PR #4285, opened 2026-08-07
oep(4268): add the SpaceWeighted scheduler to zfs pool pattern

krishnaGajabi:
VolumeWeighted and CapacityWeighted both weigh a node by what has already been provisioned into its pool, and neither says anything about what is left in it, so a node with a small untouched pool outranks a node with a large, moderately used one.

Specify a third algorithm, SpaceWeighted, which orders the nodes by the free capacity of their roomiest matching pool, the same pool resolvePool would place the volume in. It is opt-in; CapacityWeighted stays the default, since this OEP already changes that algorithm's metric on upgrade and re-ordering existing storageclasses a second time, by a different axis, is one change too many for a single release.

Two properties are called out as load-bearing, both consequences of lib-csi's scheduler semantics: the weight has to be inverted, since the scheduler prefers the least weighted node, and a node whose matching pool is full has to keep its entry rather than be dropped as lvm-localpv does, since an omitted node is front loaded rather than excluded. Fit stays the suitability intersection's job.

### DIFF
--- a/designs/local-pv/zfs/poolpattern.md
+++ b/designs/local-pv/zfs/poolpattern.md
@@ -7,7 +7,7 @@ owners:
   - "@krishnaGajabi"
 editor: "@krishnaGajabi"
 creation-date: 2026-07-14
-last-updated: 2026-07-29
+last-updated: 2026-08-07
 status: implementable
 ---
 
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

designs/local-pv/zfs/poolpattern.md
```diff
@@ -137,6 +142,38 @@ output, not by omission from the weight map.
...
+with a small, untouched pool therefore outranks a node with a large, moderately used one,
+even though the latter has far more room for the volume. `SpaceWeighted` orders by the free
+capacity of the node's roomiest matching pool instead — the same pool `resolvePool` would
+place the volume in, so the metric matches where the volume actually lands. It is the ZFS
+analog of lvm-localpv's [`SpaceWeighted`](https://github.com/openebs/lvm-localpv/blob/develop/pkg/driver/schd_helper.go),
+and it applies to `poolname` and `poolpattern` alike.
+
+Two properties of it are load-bearing:
+
+* **The weight is inverted.** `lib-csi`'s `Scheduler` sorts *ascending* and prefers the
```
abhilashshetty04, line 156: Are you planning to make changes in lib-csi or is this handled in localpv-zfs?

niladrih (approved):

designs/local-pv/zfs/poolpattern.md
krishnaGajabi: This is handled in localpv-zfs

krishnaGajabi:
@Abhinandan-Purkait , Can you please help me merge this ?

